PR #12, opened 2023-11-29
Docs: Add proper playground

### LINKED ISSUE #732: [docs] Improve the StyleX playground
The StyleX website has an empty page for a playground. This can be implemented using "WebContainers".

A basic playground should ship React, StyleX and a simple bundler. It should have four panes:

1. A source editor pane that support tabs for multiple files
2. A "transformed JS" pane that shows what the current file looks like after compilation.
3. A "CSS" pane that shows the overall generated CSS
4. A "preview" pane that shows a preview of the website.

nmn:
https://stylex-docusaurus-nxsbvzf80-nmn.vercel.app

For now, this is a proof of concept. I have used a webcontainer to successfully compile and show the output of a file.

The next step is to make improvements until there is a minimal viable experience.

Fixes #732


nmn:
> once unpacked by the browser the minified size matters

We can pack this into smaller string expressions, but it's hard to do that while maintaining className order. We used to do that in the past.

PR #44, opened 2023-12-06
Perf: bail out fast if input doesn't contain `stylex`

SukkaW:
The PR applies performance improvements similar to https://github.com/johanholmerin/style9/pull/74.

Currently, when the stylex babel plugin collects no `stylex` import, the babel plugin will bail out the tranformation.

IMHO we can push that even further. The entire `babel.transformAsync` can be avoided for better performance if the input code doesn't contain the string `stylex`.


necolas:
FWIW we're planning to support configuration to process files using imports with names other than `stylex`, so at some point this probably wouldn't be a hardcoded string.

The last idea was had was to extend the [`importSources` API](https://stylexjs.com/docs/api/configuration/babel-plugin/#importsources) (used to alias package name) to support defining custom specifiers too. For example

```js
{
  importSources: [
    // name of package
    // `import * as styles from 'stylex'
    'stylex',
    // name of package and specifier
    // `import * as css from 'other-package'
    [ 'other-package', { specifier: 'css' } ]
  ]
}
```

packages/nextjs-plugin/src/custom-webpack-loader.js
```diff
@@ -9,8 +9,14 @@
 
 const PLUGIN_NAME = 'stylex';
 
-module.exports = function stylexLoader(inputCode) {
+module.exports = function stylexLoader(inputCode, inputSourceMap) {
   const callback = this.async();
+
+  // bail out early if the input doesn't contain "stylex"
+  if (!inputCode.includes('stylex')) {
```
nmn, line 16: If an alias is used, this will break. Instead, you can read the `inputSources` option (defaulting to `@stylexjs/stylex`) and use that for checking instead.

nmn (changes requested):
This is a good idea, but needs to account for `inputSources`

packages/nextjs-plugin/src/custom-webpack-loader.js
SukkaW: Sure, let me fix that.

SukkaW:
@nmn @necolas Well, I found out that things are a little bit complicated:

- The stylex's webpack plugin has already implemented some fast bail-out
  - However, the bail-out  was based on an option `stylexImports`, which is an inconsistent name as the Babel plugin's option is `importSources`
  - The PR passes the webpack plugin's `stylexImports` to the Babel plugin's `importSources`
- The stylex's Rollup plugin hasn't implemented the fast bail-out
  - The PR adds that.

I have rebased the PR.

packages/nextjs-plugin/src/custom-webpack-plugin.js
```diff
@@ -84,6 +84,7 @@ class StylexPlugin {
           type: 'commonJS',
           rootDir,
         },
+        importSources: stylexImports,
```
nmn, line 87: 👍

packages/rollup-plugin/src/index.js
```diff
@@ -23,6 +23,7 @@ module.exports = function stylexPlugin({
   unstable_moduleResolution = { type: 'commonJS', rootDir: process.cwd() },
   fileName = 'stylex.css',
   babelConfig: { plugins = [], presets = [] } = {},
+  stylexImports = ['stylex', '@stylexjs/stylex'],
```
nmn, line 26: We can probably change this to just `@stylexjs/stylex`.

packages/rollup-plugin/src/index.js
```diff
@@ -47,6 +48,11 @@ module.exports = function stylexPlugin({
       return false;
     },
     async transform(inputCode, id) {
+      if (!stylexImports.some((importName) => inputCode.includes(importName))) {
+        // In rollup, returning null from any plugin phase means "no changes made".
+        return null;
```
nmn, line 53: 👍

nmn:
Could you please run prettier so we can merge? Thanks!

packages/rollup-plugin/src/index.js
SukkaW: IMHO we could keep `['stylex', '@stylexjs/stylex']` to prevent any potential breaking changes.

PR #48, opened 2023-12-06
fix: rollup plugin add necessary deps

nonzzz:
# Background

`@stylexjs/rollup-plugin` loose necessary dependencies, So this pull request add them.


nmn:
You also need to run `npm i` in the root dir to update the `package-lock.json` file.

nonzzz:
@nmn thx

PR #60, opened 2023-12-06
Feat: Try to add types to the Rollup Plugin

nmn:
Adds Flow and TS types to the Rollup plugin.


apps/rollup-example/rollup.config.mjs
```diff
@@ -15,7 +15,7 @@ const config = {
     file: './.build/bundle.js',
     format: 'es',
   },
-  plugins: [stylexPlugin({ fileName: 'stylex.css' })],
+  plugins: [stylexPlugin.default({ fileName: 'stylex.css' })],
```
nmn, line 18: It feels wrong to be forced to do this. Something needs to change.

Issue #46, opened 2023-12-06
Add types for the @stylexjs/rollup-plugin package


HorusGoul:
**Is your feature request related to a problem? Please describe.**
The current implementation of the Rollup plugin doesn't use Flow as other packages like `@stylexjs/babel-plugin`, so it doesn't include TypeScript or Flow type definitions.

**Describe a solution you'd like**
Type the package.

**Describe alternatives you've considered**
An alternative would be to publish the typings to DefinitelyTyped, but that doesn't feel right since this project ships type definitions for other packages.

**Additional context**
\-

---

I'm open to creating a PR about this, but I've never used Flow before so I need to take a look at the docs first lol.

nmn:
I'm already working on this!

kevintyj:
It seems that it already generates a types file, adding the types to package.json and initially testing it seems to function as indented.

nmn:
Merged.

PR #64, opened 2023-12-06
chore: Add type references to package.json

kevintyj:
Adds "types" to all of the packages' package.json (if the build output includes a type declaration)

## Linked PR/Issues
#46 

## Additional Context
[TS Docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)

## Pre-flight Checklist
- [x] I have read the [contributing guidelines](https://github.com/facebook/stylex/blob/main/.github/CONTRIBUTING.md)
- [x] I have signed the contributing [agreement](https://code.facebook.com/cla)
- [ ] I have written unit tests where necessary (If it is a feature commit)
- [x] Made sure PR title follows the conventional commit specification

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1884,6 +1884,8 @@
     },
     "node_modules/@babel/plugin-proposal-class-properties": {
       "version": "7.18.6",
+      "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-class-properties/-/plugin-proposal-class-properties-7.18.6.tgz",
+      "integrity": "sha512-cumfXOF0+nzZrrN8Rf0t7M+tF6sZc7vhQwYQck9q1/5w2OExlD+b4v4RpMJFaV1Z7WcDRgO6FqvxqxGlwo+RHQ==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


nmn:
We use a different pattern to publish types where the type definition files shadow the source files. These files are regular Typescript code instead of declaration files.

AFAIK, this has been working correctly without the `"types"` field in package.json

I want to verify that this PR won't cause a regression before we can merge. Also, if a Typescript expert can chime in, that would be helpful.

joshuajaco:
As long as the declaration file has the same name and path, as it does in this case, it will work.

See [package.json "main" and "types"](https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-main-and-types)
> If "types" and "typings" are not present or cannot be resolved, TypeScript will read the "main" field and perform [extension substitution](https://www.typescriptlang.org/docs/handbook/modules/reference.html#file-extension-substitution) to find a declaration file.

nmn:
Thanks for the link @joshuajaco 

> When publishing a typed package to npm, it’s recommended to include a "types" field even if [extension substitution](https://www.typescriptlang.org/docs/handbook/modules/reference.html#file-extension-substitution) or [package.json "exports"](https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports) make it unnecessary, because npm shows a TS icon on the package registry listing only if the package.json contains a "types" field.

This confirms that adding the `types` field to package.json is an improvement. Happy to merge once you rebase.

nmn:
Thank you for your PR @kevintyj !

This change and more has been incorporated in a different PR that also adds ESM support, so I'm now closing this PR.

PR #72, opened 2023-12-06
Remove `React.MixedElement` from the typescript example

none23:
AFAIK, `React.MixedElement` is not a valid type in Typescript


nmn (approved):
Good catch!

PR #73, opened 2023-12-06
fix: StyleX typo

maks-n:
Fixes typo from (Stylex) to (StyleX)


necolas:
Please can you combine this with your other PR? No need for 2 PRs that each change one character. Thanks

PR #75, opened 2023-12-06
Update 02-thinking-in-stylex.mdx

eliranlevi:


necolas:
This is fixed on `main`. Thanks

Issue #76, opened 2023-12-06
Background do not have a sufficient contrast ratio for "Search"


abhishekjani08:
**Is your feature request related to a problem? Please describe.**
The background does not have a sufficient contrast ratio for the "Search" feature, making it difficult for users with impaired vision to read the text.

**Describe a solution you'd like**
I would like to request an improvement in the color contrast of the background for the "Search" feature. This could involve adjusting the background color or the text color to ensure that there is a proper and accessible contrast ratio.

**Describe alternatives you've considered**

1. Implementing a customizable color scheme: Allow users to choose a color scheme that suits their preferences and improves readability.

2. Adding an accessibility mode: Introduce a feature that enhances contrast for users who specifically need it without affecting the default design.

**Additional context**
![image](https://github.com/facebook/stylex/assets/88500027/d2766a3c-4c9f-4c65-8f67-09da31475c94)
As we can see in the Lighthouse, the color contrast is not proper. For some users with not proper vision it is very difficult to read.
![image](https://github.com/facebook/stylex/assets/88500027/4f5bfdbc-cffa-401b-a6b5-acca097db7d9)

nmn:
Changing the placeholder color to increase contrast is probably the quickest fix. PRs would be appreciated!

Abhay5855:
Can you assign me this issue please, I can work on it

nmn:
The issue has been resolved already.

PR #81, opened 2023-12-07
docs: increase contrast in home CTA and search

nedjulius:
Addresses the issue #76 

Changed contrast color for both CTA and search placeholder texts.

| Before | After |
| ------ | ----- |
| image | image |

Passes the lighthouse test:
image

### DIFF
--- a/apps/docs/components/CtaButton.js
+++ b/apps/docs/components/CtaButton.js
@@ -61,8 +61,8 @@ const styles = stylex.create({
   },
   pink: {
     color: {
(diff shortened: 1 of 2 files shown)


nmn:
I think this PR goes a bit too far.

1. To improve the contrast of the CTA, I'll tweak the shade of pink being used instead.
2. The search placeholder now no longer looks like a Placeholder.

Could you revert the CTA change, and choose a medium color for the placeholder?

Issue #93, opened 2023-12-07
Lifecycle script `build` failed


abrahamvphilip:
hey folks, need some help here with this error -

<img width="580" alt="image" src="https://github.com/facebook/stylex/assets/36453614/98429feb-9446-43b9-8fdf-899d2e93ddc6">

necolas:
Hi, please fill out the issue template properly next time. For example, we can't debug errors without knowing what command was run first.

PR #94, opened 2023-12-07
refactor: nextjs-example page.tsx code refactored for better readability

abrahamvphilip:
Refactored the Card Component Rendering in page.tsx

Issue: The Card component was instantiated multiple times to render multiple content.

Solution: Implemented a more favorable approach by creating a 'cards' array of objects and utilizing the map function to render each Card component. This not only reduces code redundancy but also improves the maintainability and readability.

---

I tried earlier but had some issues and fixed that, but now there's one issue that persists and am not sure if this will work. Am sorry, didn't mean to do it twice. I've mentioned the issue in #93

### DIFF
--- a/apps/nextjs-example/app/page.tsx
+++ b/apps/nextjs-example/app/page.tsx
@@ -136,6 +136,29 @@ const s = stylex.create({
 const HOMEPAGE = 'https://stylexjs.com';
 
 export default function Home() {
+  const cards = [
+    {
+      body: 'Learn how to use StyleX to build UIs',
+      href: `${HOMEPAGE}/docs/learn/`,
+      title: 'Docs',
+    },
+    {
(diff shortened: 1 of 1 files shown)


necolas (changes requested):
I don't think the indirection is an improvement. Nothing wrong with the existing code. I'll leave this open for a bit in case you want to submit just the typo fixes though

abrahamvphilip:
> I don't think the indirection is an improvement. Nothing wrong with the existing code. I'll leave this open for a bit in case you want to submit just the typo fixes though

Uhh okay, no probs. i fixed the typo on my new code. do you want me to revert it back to original or this works?

necolas:
Yes revert the changes please

nmn (approved):
Thanks for the grammar fixes.

PR #95, opened 2023-12-07
Update setup instructions

necolas:
* Removes misleading information from the docs workspace README.
* Ensures the packages are automatically built after `npm install`

### DIFF
--- a/apps/docs/README.md
+++ b/apps/docs/README.md
@@ -1,36 +1,11 @@
-# Website
+# StyleX website
 
 This website is built using [Docusaurus 2](https://docusaurus.io/), a modern static website generator.
 
-### build stylex packages
+## Local Development
 
-- stylex actual package, and then the @stylexjs/babel-plugin plugin in node_modules
-
-```bash
-$ npm run build -w @stylexjs/stylex -w @stylexjs/shared -w @stylexjs/eslint-plugin -w @stylexjs/babel-plugin
-```
-
-### Installation
-
-```bash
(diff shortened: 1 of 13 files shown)


apps/docs/README.md
```diff
@@ -1,36 +1,11 @@
-# Website
+# StyleX website
```
necolas, line 1: The deployment instructions are also probably wrong, but there are no other instructions on how deploys work.

apps/docs/package.json
```diff
@@ -13,10 +13,7 @@
     "serve": "docusaurus serve",
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
-    "ci": "npm run lint && npm run format:diff",
-    "lint": "eslint --cache \"**/*.js\" && stylelint \"**/*.css\"",
-    "format": "prettier --config .prettierrc --write \"**/*.{js,jsx,ts,tsx,md,mdx}\"",
-    "format:diff": "prettier --config .prettierrc --list-different \"**/*.{js,jsx,ts,tsx,md,mdx}\""
+    "lint": "eslint --cache \"**/*.js\" && stylelint \"**/*.css\""
```
necolas, line 16: We never run this, but maybe we should? Hook up any relevant tasks to this pacakge's `test` script, so it gets run when `npm test` is run from the monorepo root inc in CI.

nmn:
Merging. Will delete the deployment section myself.

PR #102, opened 2023-12-08
perf: babel-plugin exports static methods type

nonzzz:
Make `@stylexjs/babel-plugin` types happy.The pr exports method `styleXTransform` static methods types.


packages/babel-plugin/src/index.js
```diff
@@ -302,3 +302,12 @@ function addSpecificityLevel(selector: string, index: number): string {
 
   return `${beforeCurly}${pseudo}${afterCurly}`;
 }
+
+export type StyleXTransformObj = {
+  (): PluginObj<>,
+  withOptions: typeof styleXTransform.withOptions,
+  processStylexRules: typeof processStylexRules,
+  ...
+};
```
nmn, line 311:
```suggestion
export type StyleXTransformObj = $ReadOnly<{
  (): PluginObj<>,
  withOptions: typeof styleXTransform.withOptions,
  processStylexRules: typeof processStylexRules,
  ...
}>;
```

nmn: Would prefer the type to be read-only if possible.

nmn (approved):
This is a nice improvement. Thanks!

PR #104, opened 2023-12-08
refactor(babel-plugin): remove unnecessary check

jason89521:
In `validateStyleXCreate`, we have checked whether the first argument is `ObjectExpression` or not. It seems that we don't need to check it again. Furthermore, the error message should be `messages.NON_OBJECT_FOR_STYLEX_CALL`


nmn:
Good catch. Thanks!

Issue #83, opened 2023-12-07
esbuild plugin


mohamedmansour:
**Is your feature request related to a problem? Please describe.**
I use esbuild to build my web apps, and need stylex to generate the css for me.

**Describe a solution you'd like**
esbuild plugin

**Describe alternatives you've considered**
Reverting to webpack again which I don't want.

mohamedmansour:
> The work to create a Vite plugin is underway. We can tackle this after that is done.

Thank you! Will that by any chance include Module support instead of CommonJS?

nmn:
We have a POC for a Vite plugin that works with Qwik. Qwik relies on SSR in a way that is similar to Next App dir, so Making Qwik would imply that it should work with most Vite apps.

https://github.com/nmn/qwik-stylex

I will be porting this to a few additional Vite frameworks to verify that it works as expected.

eryue0220:
I have written the plugin through unplugin to support: rspack, esbuild and vite. The repo is [here](https://github.com/eryue0220/unplugin-stylex).

But there is in a primitive stage, and there are still many things to complete. 

Any questions DM me, or commit an issue in the repo.

Thanks~

nmn:
An Esbuild plugin is available as `0.5.0-alpha.2` please see if it works for your use-cases.

nmn:
Fixed in #319

PR #113, opened 2023-12-09
feat: add `esbuild` plugin

nedjulius:
Addresses issue #83 

Adds a _StyleX_ plugin for _esbuild_ bundler. There are some caveats with how the _esbuild_ plugin API works, and it seems that the consumer will have to provide a full path to the build directory directly to the plugin, because there's no other way to access the `build` path in the `onEnd` callback. I am using the `onEnd` callback in this case (which runs at the end of the build, implied by the name), as we first have to transform each `ts[x]/js[x]` before we can process the `StyleX` rules. Let me know if you have simpler ideas on how to solve the bundling of CSS.

An example of how consuming the plugin would look like:
```
await esbuild.build({
  entryPoints: [path.resolve(__dirname, 'src/index.js')],
  outdir: 'dist',
  outfile: 'bundled.js',
  plugins: [stylexPlugin({ absoluteFilePath: path.resolve(__dirname, 'dist/stylex.css') })],
});
```

Some remaining things to do:
- [x] Improve tests, specifically test for injections in dev and for the output in non-dev build
- [x] Add CSS writes to disk


packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,105 @@
...
+          ],
+        });
+
+        if (!dev && metadata.stylex !== null && metadata.stylex.length > 0) {
+          stylexRules[args.path] = metadata.stylex;
+        }
+
+        return {
+          contents: code,
+          loader: currFileName.endsWith('.js') ? 'js' : 'tsx',
```
nedjulius, line 98: `tsx` loader can handle `ts`, `tsx` and `jsx` files

packages/esbuild-plugin/__tests__/__fixtures__/.babelrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "ie": 11
```
nmn, line 7: Please use the same config as the other plugins.

packages/esbuild-plugin/__tests__/__fixtures__/index.js
```diff
@@ -0,0 +1,39 @@
...
+    height: 700,
+    backgroundColor: 'red',
+    ':hover': {
+      backgroundColor: 'pink',
+    },
+  },
+});
+
+export default function App() {
+  return stylex(fooStyles.foo, styles.bar);
```
nmn, line 38: Use `stylex.props` instead.

packages/esbuild-plugin/__tests__/__fixtures__/index.js
```diff
@@ -0,0 +1,39 @@
...
+const styles = stylex.create({
+  bar: {
+    animationName: fadeInAnimation,
+    display: 'flex',
+    marginLeft: 10,
+    height: 700,
+    backgroundColor: 'red',
+    ':hover': {
+      backgroundColor: 'pink',
+    },
```
nmn, line 33: This is a legacy pattern. Use the officially supported API. 

```suggestion
    backgroundColor: {
      default: 'red',
      ':hover': 'pink',
    },
```

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,105 @@
...
+const PACKAGE_NAME = 'esbuild-plugin-stylex';
+
+const IS_DEV_ENV =
+  process.env.NODE_ENV === 'development' ||
+  process.env.BABEL_ENV === 'development';
+
+function stylexPlugin({
+  dev = IS_DEV_ENV,
+  unstable_moduleResolution = { type: 'commonJS', rootDir: process.cwd() },
+  stylexImports = ['stylex', '@stylexjs/stylex'],
```
nmn, line 24:
```suggestion
  stylexImports = ['@stylexjs/stylex'],
```

nmn:
Looks promising so far!

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+        const { code, metadata } = await babel.transformAsync(inputCode, {
+          babelrc: false,
+          filename: currFileName,
+          presets,
+          plugins: [
+            ...plugins,
+            /\.jsx?/.test(path.extname(currFileName))
+              ? flowSyntaxPlugin
+              : typescriptSyntaxPlugin,
+            jsxSyntaxPlugin,
```
nmn, line 85: I've run into probably because of this sometimes. So let's add it.

```suggestion
              : [typescriptSyntaxPlugin, {isTSX: true}],
            jsxSyntaxPlugin,
```

packages/esbuild-plugin/__tests__/__fixtures__/.babelrc.json
nedjulius: removed this fixture `babelrc` altogether, since its not used

packages/esbuild-plugin/src/index.js
nedjulius: update: turns out this wasn't entirely correct. encountered some import resolution problems when using `tsx` loader for `JSX` files, even though the `esbuild` documentation said that either loader is fine. for the sake of simplicity, I just match extension to loader.

apps/esbuild-example/package.json
```diff
@@ -0,0 +1,19 @@
+{
```
nedjulius, line 1: added an example package that provides a simple build script example for a React application. it is also easier to develop the plugin this way.

nmn:
@nedjulius Do you consider this PR ready for review?

nedjulius:
@nmn yeah, I think we can give it a go!

apps/esbuild-example/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "build": "node scripts/build.js"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@stylexjs/stylex": "^0.3.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  },
+  "devDependencies": {
+    "@stylexjs/esbuild-plugin": "^0.0.0",
```
nmn, line 16: Can you enable the ESLing plugin for this example as well? Like the NextJS and Docs apps.

apps/esbuild-example/public/index.html
```diff
@@ -0,0 +1,19 @@
...
+<html>
+  <head>
+    <title>@stylexjs/esbuild-plugin</title>
+    <meta charset="utf-8" />
+    <style>
+      body {
+        box-sizing: border-box;
+        padding: 0;
+        margin: 0;
+      }
```
nmn, line 11: Since you have `useCSSLayers: true`, the reset needs to be in a layer as well.

```suggestion
      @layer reset {
        html {
          box-sizing: border-box;
          padding: 0;
          margin: 0;
        }
      }
```

apps/esbuild-example/scripts/build.js
```diff
@@ -0,0 +1,39 @@
...
+    bundle: true,
+    outfile: OUTFILE,
+    minify: true,
+    plugins: [
+      stylexPlugin({
+        useCSSLayers: true,
+        absoluteFilePath: STYLEX_BUNDLE_PATH,
+        unstable_moduleResolution: {
+          type: 'commonJS',
+          rootDir: path.resolve(__dirname, '..'),
```
nmn, line 34: Change this to the stylex repo root dir?

apps/esbuild-example/src/App.jsx
```diff
@@ -0,0 +1,47 @@
...
+ *
+ *
+ */
+
+'use strict';
+
+import * as React from 'react';
+import ReactDOM from 'react-dom';
+import * as stylex from '@stylexjs/stylex';
+import { spacing, color } from './tokens.stylex.js';
```
nmn, line 15: ...And then you can try using `open-props` here as well.

packages/esbuild-plugin/README.md
```diff
@@ -0,0 +1,3 @@
+# @stylexjs/esbuild-plugin
+
+...
```
nmn, line 3: You can probably add some basic documentation now.

packages/esbuild-plugin/__tests__/index-test.js
```diff
@@ -0,0 +1,202 @@
...
+  });
+
+  return { js: outputFiles[0], css: outputFiles[1] };
+}
+
+describe('esbuild-plugin-stylex', () => {
+  it('extracts and bundles CSS without inject calls, bundles JS', async () => {
+    const { js, css } = await build();
+
+    expect(js.text).toMatchInlineSnapshot(`
```
nmn, line 33: Can we use file snapshots here? They're pretty long.

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,118 @@
+/**
```
nmn, line 1: We're starting to add types to the packager plugins. See if you're up to the task of adding Flow types. If not, I'll do it after the fact.

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,118 @@
...
+const IS_DEV_ENV =
+  process.env.NODE_ENV === 'development' ||
+  process.env.BABEL_ENV === 'development';
+
+function stylexPlugin({
+  dev = IS_DEV_ENV,
+  unstable_moduleResolution = { type: 'commonJS', rootDir: process.cwd() },
+  stylexImports = ['@stylexjs/stylex'],
+  // absolute path to bundled CSS file
+  absoluteFilePath = path.resolve(__dirname, 'stylex.css'),
```
nmn, line 29: Can we rename this option to something more descriptive?

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,118 @@
...
+  babelConfig: { plugins = [], presets = [] } = {},
+  useCSSLayers = false,
```
nmn, line 31: Again, let's leave this without a default value.

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,118 @@
...
+        outputFiles.push({
+          path: '<stdout>',
+          contents: new TextEncoder().encode(collectedCSS),
+          get text() {
+            return collectedCSS;
+          },
+        });
+      });
+
+      onLoad({ filter: /\.[jt]sx?$/ }, async (args) => {
```
nmn, line 69:
```suggestion
      onLoad({ filter: /\.[mc]?[jt]sx?$/ }, async (args) => {
```

To account for `.mjs` and `cjs` etc.

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,118 @@
...
+        const currFilePath = args.path;
+        const inputCode = await fs.readFile(currFilePath, 'utf8');
+
+        if (
+          !stylexImports.some((importName) => inputCode.includes(importName))
```
nmn, line 74: Oh I see why you have a default value for `stylexImports` now.

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,118 @@
...
+        const currFileExtension = path.extname(currFilePath);
+
+        const { code, metadata } = await babel.transformAsync(inputCode, {
+          babelrc: false,
+          filename: currFilePath,
+          presets,
+          plugins: [
+            ...plugins,
+            /\.jsx?/.test(currFileExtension)
+              ? flowSyntaxPlugin
```
nmn, line 90: In the non-TS, i.e. Flow case, we need to also use the `hermes-parser` plugin. This is something that needs to be added to each of the bundler plugins.

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,118 @@
...
+          return;
+        }
+
+        outputFiles.push({
+          path: '<stdout>',
+          contents: new TextEncoder().encode(collectedCSS),
+          get text() {
+            return collectedCSS;
+          },
+        });
```
nmn, line 66: What is going on here? Are you just printing the result to `stdout`?

nmn (commented):
I don't see anything particularly wrong. Left a bunch nit-picky comments.

Will pull these changes locally to test how this works.

packages/esbuild-plugin/src/index.js
nedjulius: yeah. if the consumer specifies `write: false` in the `esbuild` configuration, then the bundled files are written to `outputFiles` array that contains objects of the structure above (they also have a `hash` property, which is generated by `esbuild` internally). just wanted to replicate the behavior for consistency, like if some plugin consumer decided to save `esbuild` output in memory first and then write everything to disk manually.

packages/esbuild-plugin/src/index.js
nedjulius: changed it to `generatedCSSFileName`. I don't think it should have the `absolute` in it's name as it can also consume relative file paths, though using absolute path is a recommendation.

nedjulius:
@nmn I addressed all of your comments. the only issue left now is that `flow` complains about not being able to resolve some modules. not sure why though, but I assume it has something to do with those modules not having `flow` types (?). I saw that you created `flow_modules` file with type definitions for `rollup` and some `babel` plugins in the `rollup-plugin` -- is there a way to generate that somehow?

nmn:
> is there a way to generate that somehow?

There are some scripts online for it as well, but I manually translate the Typescript types for them.

nedjulius:
@nmn got you, thanks!

I added translations for _esbuild_ and also ported babel types from the _rollup-plugin_. perhaps `@babel/core` definition should be placed in `flow-typed` as it is now used in two multiple packages?

besides that, I also added the `node-libraries` to `flow-typed` as I was getting type resolution issues with filesystem module.

apps/docs/docs/learn/03-installation.mdx
```diff
@@ -196,6 +196,48 @@ provides plugins for Webpack, Rollup, and Next.js.
 
 </details>
 
+<details>
```
nedjulius, line 199: Added _esbuild_ plugin example to the installations page in docs.

nmn: Please be patient as it may take me a few days to review this properly.

nmn:
Lots of failing tests, so please work on that as I make my way through the PR.

nedjulius:
@nmn yup, looking into it. probably related to package lock merge conflicts that I tried to resolve yesterday.

flow-typed/node-libraries.js
```diff
@@ -0,0 +1,15 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+/* eslint-disable ft-flow/no-types-missing-file-annotation */
+/* eslint-disable no-unused-vars */
+
+declare module 'fs/promises' {
```
nmn, line 13: Don't add this. The types for `fs` should already exist.

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,176 @@
...
+  ...
+}>;
+
+export default function stylexPlugin({
+  dev = IS_DEV_ENV,
+  unstable_moduleResolution = { type: 'commonJS', rootDir: process.cwd() },
+  stylexImports = ['@stylexjs/stylex'],
+  generatedCSSFileName = path.resolve(__dirname, 'stylex.css'),
+  babelConfig: { plugins = [], presets = [] } = {},
+  useCSSLayers,
```
nmn, line 48: Default this to `false`

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,176 @@
...
+            contents: code,
+            loader,
+          };
+        },
+      );
+    },
+  };
+}
+
+function getEsbuildLoader(fileName: string) {
```
nmn, line 154: Also, `.mjs` and `.cjs` and `.mts` and `.cts`

packages/esbuild-plugin/src/index.js
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  if (fileName.endsWith('ts')) {
+    return 'ts';
+  }
+
+  return 'js';
+}
+
+function getFlowOrTypeScriptBabelSyntaxPlugins(fileName: string) {
```
nmn, line 170: We also need to add check to verify that we only attempt to transform JS files. This means ignoring `.json` files.

This regex might help.
```js
/[mc]?[jt]sx?$/
```

nmn (commented):
It's almost there! I've left a few comments. Please address those and I'll test it merge it within a few days.

packages/esbuild-plugin/src/index.js
nedjulius: the context in which this function is used guarantees that we only transform JS files:

https://github.com/facebook/stylex/blob/cd1f8b94a1856bad591d34dcfdd4e325416fa016/packages/esbuild-plugin/src/index.js#L93

I assume we don't want to allow processing invalid filetypes like `.mjsx` or `.cjsx`, hence I added an explicit filter above:

https://github.com/facebook/stylex/blob/cd1f8b94a1856bad591d34dcfdd4e325416fa016/packages/esbuild-plugin/src/index.js#L28

packages/esbuild-plugin/src/index.js
nedjulius: `ts` loader handles `.mts` and `.cts` files, while `js` loader handles `.cjs` and `.mjs` -- [see here](https://esbuild.github.io/content-types/) for the reference

nedjulius:
hey @nmn are you still considering adding this?

nmn:
@nedjulius Yes. Still planning to merge this for the next release.

PR #140, opened 2023-12-12
docs: enhance sentence flow and readability

neerajkumarc:
improvement in docs for enhancing clarity and readability


necolas:
I think the existing copy is fine, and things like the comma were added in a previous pass at improving readability. Thanks though

PR #141, opened 2023-12-12
fix: strict report for stylex.create argument as object expression

kotarella1110:

I have fixed to report when a value other than an `ObjectExpression` is specified as an argument for `stylex.create`.

## Additional Context

スクリーンショット 2023-12-13 0 58 04

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/src/stylex-valid-styles.js
+++ b/packages/eslint-plugin/src/stylex-valid-styles.js
@@ -2347,8 +2347,7 @@ const stylexValidStyles = {
         node &&
         node.type === 'CallExpression' &&
         isStylexCallee(node.callee) &&
-        node.arguments.length === 1 &&
-        node.arguments[0].type === 'ObjectExpression'
+        node.arguments.length === 1
       );
     }
 
--- a/packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -360,6 +360,18 @@ eslintTester.run('stylex-valid-styles', rule.default, {
   ],
   invalid: [
     {
+      code: `
+        import stylex from 'stylex';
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -2347,8 +2347,7 @@ const stylexValidStyles = {
         node &&
         node.type === 'CallExpression' &&
         isStylexCallee(node.callee) &&
-        node.arguments.length === 1 &&
-        node.arguments[0].type === 'ObjectExpression'
+        node.arguments.length === 1
```
nmn, line 2350: The rest of the code still depends on it being an object expression.

nmn (commented):
I don't think this change can be made without a bigger refactor, but I'll let the tests run in case this is accounted for.

packages/eslint-plugin/src/stylex-valid-styles.js
nmn: Interesting! The tests are still working.

nmn:
I'll merge for now since the tests are still passing. The bigger refactor can wait.

kotarella1110:
@nmn 

> I don't think this change can be made without a bigger refactor, but I'll let the tests run in case this is accounted for.

What do you mean?

The only fix is that the following code is executed when a value other than `ObjectExpression` is specified as the argument to `stylex.create`. 

https://github.com/facebook/stylex/blob/ca1dc8bfda5cccdb722fefdaaa0a48b2caa32b75/packages/eslint-plugin/src/stylex-valid-styles.js#L2745-L2755

Prior to this fix, the following code would not report a problem, but now it does.

```
stylex.create("test")
stylex.create(1)
const foo = {}
stylex.create(foo)
```

nmn:
@kotarella1110 That's my misunderstanding. Thanks!

PR #146, opened 2023-12-12
fix: cta section responsive issue

mehedihsajib:
FIX: CTA section's responsive issue fixed

### Issue was:
When screen size is around 370-380px the CTA section touches the edge which is not looks good (Please check the attached screenshot)

### Changes I made:
- There was several options to fix it. I just reduced the padding-x of those button when screen size is less then 390px

### Before:
cta-before

### After: 
cta-after

### Notes for Reviewers:
Hope you are going great. 
Please check my changes and if you have suggestions or need any modification please let me know. 

Thank you for your time and consideration!

## Pre-flight checklist

### DIFF
--- a/apps/docs/components/CtaButton.js
+++ b/apps/docs/components/CtaButton.js
@@ -25,6 +25,8 @@ export default function CtaButton({children, color, to}) {
   );
 }
 
+const CTA_BREAK = '@media (max-width: 390px)';
+
 const styles = stylex.create({
   base: {
     display: 'flex',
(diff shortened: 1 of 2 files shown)


nmn (changes requested):
The CTA button is also used for the button the `/playground` page. We should add padding around the buttons instead.

mehedihsajib:
> The CTA button is also used for the button the `/playground` page. We should add padding around the buttons instead.

Hi,
Thanks for your feedback. 
The parent of those button is 'flex' without 'flex-wrap' that's why when screen gets narrower the button crossed the screen and margin/padding not get reflected. 

So instead of using padding I just change the media query 360px to 385px and now it seems okay. 
If you still any better suggestion please do let me know.

Thanks.

nmn (approved):

nmn:
Please fix the prettier failure.

mehedihsajib:
> Please fix the prettier failure.

Hi, the prettier failure is fixed now.

PR #158, opened 2023-12-13
fix: `stylex.create` types

nonzzz:

Can't get right types using Typescript or vanilla.js in current version. This pr is a fix.

## Linked PR/Issues

None.

## Additional Context

image

## Pre-flight checklist

### DIFF
--- a/packages/stylex/src/StyleXTypes.d.ts
+++ b/packages/stylex/src/StyleXTypes.d.ts
@@ -75,8 +75,8 @@ export type NestedCSSPropTypes = Partial<
 
 type UserAuthoredStyles = {
   [P in keyof CSSPropertiesWithExtras]?:
-    | CSSPropertiesWithExtras[P]
-    | { [key: string]: unknown };
+    | (CSSPropertiesWithExtras[P] | (string & {}))
+    | { [key: string]: CSSProperties[P] | (string & {}) };
 };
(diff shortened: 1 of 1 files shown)


packages/stylex/src/StyleXTypes.d.ts
```diff
@@ -73,7 +73,11 @@ export type NestedCSSPropTypes = Partial<
   }>
 >;
 
-type UserAuthoredStyles = CSSPropertiesWithExtras | { [key: string]: unknown };
+type UserAuthoredStyles = {
+  [P in keyof CSSPropertiesWithExtras]?:
+    | CSSPropertiesWithExtras[P]
+    | { [key: string]: unknown };
+};
```
nmn: This is incorrect.

nmn (changes requested):

packages/stylex/src/StyleXTypes.d.ts
nonzzz: I don't understand why is wrong :( .Before this the type is `CSSPropertiesWithExtras`. I guess stylex want user using `CSSPropertiesWithExtras` as attr .

nmn:
This issue has already been fixed but hasn't been released. You'll see if you play around in the NextJS example (which is the only part of the codebase that uses Typescript and not Flow)

PR #163, opened 2023-12-13
Perf: Remove runtime injection from runtime

nmn:

The `inject` function has been removed from the main export of `@stylex/stylex`. Now the inject function can only be used by manually importing `@stylex/stylex/lib/stylex-inject`.

The Babel plugin has similarly been updated to import the `inject` function from the new location when `runtimeInjection` is turned on.

Also, `runtimeInjection` now accepts a string or `{from: string, as: string}` object to be able to override the `inject` function module location.

## Additional Context

image

Tests have been updated to reflect the changes. The updated snapshots will show what has changed.

NOTE: The Webpack plugin is currently inlining the `stylex-inject` module in the generated code in the snapshot test. Trying to configure Webpack to not do that is causing strange errors.

## Pre-flight checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31133,6 +31133,7 @@
       "version": "0.3.0",
       "license": "MIT",
       "dependencies": {
+        "@babel/helper-module-imports": "^7.22.15",
         "@stylexjs/shared": "0.3.0"
       },
       "devDependencies": {
@@ -31295,6 +31296,7 @@
         "@stylexjs/babel-plugin": "0.3.0"
       },
       "devDependencies": {
+        "@babel/cli": "^7.23.0",
         "@babel/preset-env": "^7.16.8",
         "@rollup/plugin-babel": "^6.0.0",
         "@rollup/plugin-commonjs": "^25.0.3",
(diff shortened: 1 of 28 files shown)


necolas:
We should probably be checking the size of the bundled runtime in the CI task. It hasn't been showing the contribution of inject or styleq, so the size savings aren't showing up in the results

nmn:
> We should probably be checking the size of the bundled runtime in the CI task

I am looking into that. The numbers above are by using `browserify`, `terser` and `gzip`. I expect using rollup might squeeze this down a bit more.

necolas:
oh actually it looks like styleq *is* inlined in `lib/stylex.js`, so we're good. the CI workflow already minifies and gzips.

nmn:
> 2.29 kB

This is in the ballpark of the size I got in local testing, but the comparison is still wrong since this PR cut the size in half.

Will figure this out later.

necolas:
It's just because we don't inline `./stylex/lib/StyleXSheet.js` into the main export of the package - the size of that file is listed separately in the "unchanged" files table. If we had size-diffed a production rollup bundle, we'd have seen the 3K reduction from inject not being imported anymore.

Issue #135, opened 2023-12-12
[Bug] Properties not recognized by eslint-plugin


olivierpascal:
**The problem**

The `outlineOffset` property is not recognized by the eslint-plugin.

**How to reproduce**

With the following code:

```ts
import * as stylex from '@stylexjs/stylex';

export const styles = stylex.create({
  container: {
    outlineOffset: '2px', // eslint(@stylexjs/valid-styles) error here
  },
});
```

I have the following eslint error from `@stylex/valid-styles`:

```
outlineOffset value must be one of:
'outlineOffset' is not supported yet. Please use 'outline' instead
```

https://github.com/facebook/stylex/blob/f968076548895dd1adc9736f1bfc371057beb94e/packages/eslint-plugin/src/stylex-valid-styles.js#L1948-L1950

But `outline` syntax does not allow an offset:

```
outline = 
  <'outline-color'>  ||
  <'outline-style'>  ||
  <'outline-width'>
```
Cf. https://developer.mozilla.org/en-US/docs/Web/CSS/outline

olivierpascal:
Same for `transition`:

```ts
export const styles = stylex.create({
  container: {
    transition: 'box-shadow 150ms cubic-bezier(0.4, 0, 0.2, 1)', // eslint(@stylexjs/valid-styles) error here
  },
```

```
This is not a key that is allowed by stylex eslint(@stylexjs/valid-styles)
```

nmn:
NOTE: A few properties are intentionally not allowed by the ESLint plugin even though they will work correctly in the Babel plugin:
- `background`
- `transition`
- `animation`
- `border`, `border[Direction]`.
- `flex`

This decision was made for a few reasons:
1. To encourage better styling practices. We've noticed that a lot of developers don't understand what the various parts of some of these properties are.
2. To enable smaller CSS. `transitions`, `animations`, and `background` are all properties where many combinations of the same values may be used.
3. To align a bit with React Native for our future RN implementation.

We are going to be discussing and reconsidering this decision.

---

Any other missing property is an unknown bug so please report!

olivierpascal:
Well noted, thanks. Maybe the eslint error message should be more explicit about it, encouraging using individual properties like transitionProperty, transitionDuration and transitionTimingFunction?

purohitdheeraj:
I can work on this , looks interesting

IbraheemHaseeb7:
[...]

PR #174, opened 2023-12-13
Altered eslint error message for outlineOffset

tusharkc:
while outlineOffset is not supported, still eslint error was promoting to use 'outline' instead, and 'outlineOffset' is not one of the shorthand property of outline, in css.

## Linked PR/Issues

Fixes # #135 

## Additional Context

altered the eslint error message, so that it does not mis guides the user to use outline. 

## Pre-flight checklist

- [* ] I have read the contributing guidelines
- [ *] Performed a self-review of my code


nmn:
There is a different PR that actually enables these. This PR is a regression.

PR #175, opened 2023-12-13
Fix:outlineOffset is not recognized by eslint-plugin, but cannot be set otherwise.

### LINKED ISSUE #135: [Bug] Properties not recognized by eslint-plugin

xonx4l:
issue-:#135

Added outlineOffset, outlineStyle and outlineWidth properties and set them to valid style types (width and outlineStyle).

Removed the showError calls for those properties and allowed them to be used.

Left outline as-is to still support shorthand outline property.

## Linked PR/Issues

Fixes #135

### DIFF
--- a/packages/eslint-plugin/src/stylex-valid-styles.js
+++ b/packages/eslint-plugin/src/stylex-valid-styles.js
@@ -1942,18 +1942,11 @@ const CSSProperties = {
   order: order,
   orphans: orphans,
   outline: outline,
-  outlineColor: showError(
-    "'outlineColor' is not supported yet. Please use 'outline' instead",
-  ),
-  outlineOffset: showError(
-    "'outlineOffset' is not supported yet. Please use 'outline' instead",
-  ),
-  outlineStyle: showError(
-    "'outlineStyle' is not supported yet. Please use 'outline' instead",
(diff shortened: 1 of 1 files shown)


nmn (changes requested):
Please fix the errors.

nmn:
Closing due to inactivity. Also, there's only a few lines of code so no work is lost.

PR #177, opened 2023-12-13
docs: add treeshakeCompensation option to Next.js installation guide

mathieumetral:
## Why ?
Considering the significance of variables in StyleX, I suggest adding the `treeshakeCompensation` property to the installation steps for Next.js. Without this property, variables are removed during compilation as they are considered dead code. This can lead to confusion and difficulty for developers in setting up a project effortlessly.

### DIFF
--- a/apps/docs/docs/learn/03-installation.mdx
+++ b/apps/docs/docs/learn/03-installation.mdx
@@ -176,6 +176,7 @@ provides plugins for Webpack, Rollup, and Next.js.
           dev: process.env.NODE_ENV === 'development',
           runtimeInjection: false,
           genConditionalClasses: true,
+          treeshakeCompensation: true,
           unstable_moduleResolution: {
             type: 'commonJS',
             rootDir: __dirname,


nmn:
This change is not needed. `treeshakeCompensation` is always `on` in the nextjs plugin. `treeshakeCompensation` is something you need for some bundlers and is harmless in most others. There are very small edge-cases where you might need to turn it off. It only needs to be exposed and documented on the Babel level.

mathieumetral:
Thank you for the clarification! However, I feel there might be a subtle difference whether we explicitly set `treeshakeCompensation` or not, even though it's specified in the Next.js plugin. For instance, if I take the [example app](https://github.com/facebook/stylex/tree/main/apps/nextjs-example) and remove `treeshakeCompensation` from the `babelrc.js`, I notice that variables like [colors.blue3](https://github.com/facebook/stylex/blob/main/apps/nextjs-example/app/Card.tsx#L77) are not defined.

**With `treeshakeCompensation`**
Screenshot 2023-12-13 at 11 27 18 PM

**Without `treeshakeCompensation`**
Screenshot 2023-12-13 at 11 26 48 PM

nmn:
Yes. We want to remove the need for the `.babelrc` file entirely and this issue is already being tracked.

nmn:
Thank you for the PR though. If you want to submit a PR specifically documenting the `.babelrc` requirement, i will accept that.

mathieumetral:
I'm having a little trouble understanding what is expected.

If I understand correctly, today, we need to have both a Babel config file and the Next.js plugin working together. However, going forward, the requirement for the Babel config file is anticipated to become unnecessary.

So, in the meantime, until the need for the Babel config file is phased out, should it be aligned with the configuration defined in the Next.js plugin ([here](https://github.com/facebook/stylex/blob/main/packages/nextjs-plugin/src/custom-webpack-plugin.js#L74-L88))? And the change proposed in this PR aligns with that understanding.

Perhaps you would like me to add a comment explaining this temporary requirement?

nmn:
@mathieumetral This is my mistake.

nmn (approved):

PR #189, opened 2023-12-14
Support replacement tokens in the webpack filename

### LINKED ISSUE #105: [Webpack Plugin] Support add [hashcontent] to file name.
**Is your feature request related to a problem? Please describe.**
Currently with webpack config like below, a file with the name `styles.css` is created after the build.
image

**Describe a solution you'd like**
I'm expecting a hashcontent will be added to the file name according to the webpack `output.filename` config to resolve the caching issue.

**Describe alternatives you've considered**
Or can `@stylexjs/webpack-plugin` support add `[hashcontent]` to the `filename` config? like this:

```
new StylexPlugin({
      filename: 'styles.[hashcontent].css',
}),
```

blowery:

Webpack only supports token replacement for chunks and modules, not plain assets. Instead, teach the plugin to do token replacement the same way that webpack does it for supported types.

## Linked PR/Issues

Fixes #105 

## Pre-flight checklist

### DIFF
--- a/packages/webpack-plugin/src/index.js
+++ b/packages/webpack-plugin/src/index.js
@@ -146,17 +146,66 @@ class StylexPlugin {
           },
         );
       } else {
-        // Consume collected rules and emit the stylex CSS asset
-        compilation.hooks.additionalAssets.tap(PLUGIN_NAME, () => {
-          try {
-            const collectedCSS = getStyleXRules();
-            if (collectedCSS) {
-              compilation.emitAsset(this.filename, new RawSource(collectedCSS));
-            }
-          } catch (e) {
-            compilation.errors.push(e);
+        // We'll emit an asset ourselves. This comes with some complications in from Webpack.
+        // If the filename contains replacement tokens, like [contenthash], we need to
+        // process those tokens ourselves. Webpack does provide a way to reuse the configured
+        // hashing functions. We'll take advantage of that to process tokens.
+        const getContentHash = (source) => {
+          const { outputOptions } = compilation;
+          const { hashDigest, hashDigestLength, hashFunction, hashSalt } =
+            outputOptions;
+          const hash = compiler.webpack.util.createHash(hashFunction);
+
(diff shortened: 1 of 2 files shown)


packages/webpack-plugin/src/index.js
```diff
@@ -146,17 +146,66 @@ class StylexPlugin {
...
+
+          return fullContentHash.toString().slice(0, hashDigestLength);
+        };
         // Consume collected rules and emit the stylex CSS asset
-        compilation.hooks.additionalAssets.tap(PLUGIN_NAME, () => {
-          try {
-            const collectedCSS = getStyleXRules();
-            if (collectedCSS) {
-              compilation.emitAsset(this.filename, new RawSource(collectedCSS));
+        compilation.hooks.processAssets.tap(
```
blowery, line 170: `additionalAssets` was deprecated in favor of `processAssets` with the correct stage. They perform the same though.

blowery:
Filled out the CLA.

blowery:
Not part of this PR, but the approach the plugin is taking is probably ... wrong? I'd expect it to build a file per entry point, not one for the entire build, though perhaps that's counter to "thinking in stylex." For things like a standalone entry with minimal CSS requirements (a login page for example) the savings in CSS size could be substantial. 

Another alternative would be to find a way to get the plugin to emit a true CSS asset (maybe a https://github.com/webpack/webpack/blob/main/lib/CssModule.js) which would then be processed in the normal webpack-y ways. How that would tie into the chunk & module graph would take some consideration.

blowery:
I probably need to be covered under the Automattic corp CLA. Working on some back-channels to get myself added to the list.

nmn:
> I'd expect it to build a file per entry point, not one for the entire build

No it's very much intentional to generate a single CSS file. Always. The CSS size will not be an issue.

----

Outside of that, I'm happy for any PRs to make this plugin better. I've always struggled to make sense of the Webpack API.

PR #202, opened 2023-12-17
Fixing the issue with aliased imports in Nextjs

### LINKED ISSUE #40: Importing a .stylex.ts does not work with tsconfig paths settings
**The problem**
When importing a variable file like is listed in [the docs](https://stylexjs.com/docs/learn/theming/defining-variables/), the build is only succeeding if the import path is a relative path (i.e. ../src/lib/theme/theme.stylex), but fails if the import path uses the tsconfig "paths" setting (i.e. @/lib/theme/theme.stylex when tsconfig paths is { '@/*' : '.src/*'} ). 

With non-relative paths, it gives the `Only static values are allowed inside of a stylex.create() call` error message.

**How to reproduce**

Steps to reproduce:
1. Create a next.js app with the default of `paths: { '@/*' : '.src/*'}` in the tsconfig.json file.
2. Create a file called theme.stylex.ts in `/src/lib/` that defines a color using `export const colors = stylex.defineVars({test: '#000000'})`
3. Try to import the colors using `import { colors } from @/lib/theme.stylex` and use `colors.test` in a `stylex.create` call.

**Expected behavior**
I would expect this to allow `colors` to be used in `stylex.create` as specified in [the docs](https://stylexjs.com/docs/learn/theming/defining-variables/).

**Environment (include versions). Did this work in previous versions?**
- next 14.0.2 using an app router
- @stylexjs/stylex - 0.3.0
- @stylexjs/nextjs-plugin - 0.3.0

rayan1810:

Added changes in StateManager class in babel plugin, introduced an extra step of resolution if aliases are provided in babel and nextjs config. This fixes some part of this issue https://github.com/facebook/stylex/issues/40

## Linked PR/Issues

[Fixes #40 ]

## Pre-flight checklist


apps/nextjs-example/.babelrc.js
```diff
@@ -13,6 +12,10 @@ module.exports = {
         unstable_moduleResolution: {
           type: 'commonJS',
           rootDir: path.join(__dirname, '../..'),
+          aliases: {
+            '@/app/*': './app/*',
+            '@/components/*': './components/*',
+          },
```
nmn, line 18: Would this work too?

```suggestion
            '@/*': './*',
          },
```

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -26,15 +26,18 @@ type ModuleResolution =
       type: 'commonJS',
       rootDir: string,
       themeFileExtension?: string,
+      aliases?: any,
```
nmn, line 29: The type are obviously wrong. I'm reading the rest of the code and I can help with the types as needed.

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -46,6 +49,7 @@ export type StyleXOptions = $ReadOnly<{
   treeshakeCompensation?: boolean,
   genConditionalClasses: boolean,
   unstable_moduleResolution: void | ModuleResolution,
+  aliases: any,
```
nmn, line 52: This is the only place where `aliases` should be added. Remove the other ones from `ModuleResolution` type.

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -212,16 +217,25 @@ export default class StateManager {
...
       return false;
     }
+
     switch (this.options.unstable_moduleResolution?.type) {
       case 'commonJS': {
         const rootDir = this.options.unstable_moduleResolution.rootDir;
+        const aliases =
+          this.options.unstable_moduleResolution.aliases ??
+          this.options.aliases ??
+          undefined;
```
nmn, line 227:
```suggestion
        const aliases = this.options.aliases;
```

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -260,12 +279,40 @@ export default class StateManager {
     this.styleVarsToKeep.add(memberExpression);
   }
 }
-
+// a function generate Regex to match the path of files with aliases
+function generateAliasRegex(alias: string) {
```
nmn, line 283:
```suggestion
function generateAliasRegex(alias: string): RegExp {
```

Return types are required in Flow.

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -260,12 +279,40 @@ export default class StateManager {
...
+  const regex = new RegExp('^' + alias + '/(.*)');
+  return regex;
+}
+// a function that resolves the path of files with aliases
+function aliasPathResolver(importPath: string = '', aliases: any) {
```
nmn, line 288: This function can be simpler IMO.

packages/babel-plugin/src/utils/state-manager.js
nmn: Here's the type that you can use for `Aliases`.

```ts
type Aliases = $ReadOnly<{[string]: string}>;
```

Please replace all the `any` types.

nmn (commented):
Thanks for taking this on! This is 90% right already.

Some things I want to consider:
1. I want to look at how Typescript accepts aliases to ensure we support all the same patterns
2. There will need to be some kind of validation of the `aliases` object. And we should probably store the regular expressions in the `options` object up-front.
3. The function to actually resolve the paths can be simpler
4. The output path currently starts with `./`, we will need to `path.join` that with the `rootDir` path.

apps/nextjs-example/.babelrc.js
rayan1810: Yes this works.

packages/babel-plugin/src/utils/state-manager.js
nmn: - You don't need a regEx at all.
- There are two possible cases:
  1. The `alias` ends with `*`
     - If so, check if the importedPath starts with the alias (minus the `*`)
     - If it does, replace the `alias` with the replacement at the beginning of the string 
  3. The `alias` does not end with `*`
     - Check if the imported path is *equal* to the alias
     - If so, replace it with the alias replacement
- If the imported path was replaced by an alias, `path.join` it with the `rootDir` option to get an absolute file path.

rayan1810: @nmn so I looked into this a little more and see that it doesn't make any sense if alias does not end with `*`. A case that is possible is something like this: `"@util/*": ["./src/util/*/src/index.ts"]` making this `@util/someDir` correspond to `./src/util/someDir/src/index.ts`. The current regex logic also doesn't match this. I'm fixing with this in mind now. Please let me know if you have any use case other than this.

nmn: I've been studying how the aliases work in `tsconfig` and `package.json` and here's what I found.

I'll use the term `key` and `value` to refer to the import path configurations.

## Common Patterns

1. Aliases can contain 0 or 1 `*` characters. Never More.
2. The `*` character is *NOT* a glob pattern, but rather equivalent to the regex `/.*/` and will match *any* string.
3. If the `key` has a `*` then the should also contain a single `*`. This patterns means that the string in the position of the `*` in the `key` should be inserted into the `*` in the value to get the final path.
4. The `*` can exist anywhere other than the beginning of the `key`.

## Specific TS patters

1. When multiple `key`s match, Typescript use the one with the longest string *before* the `*`.
2. The `value` can be an array of strings. These strings are tried in order as they may or may not exist.

## Specific `package.json` patterns

1. All `key`s must start with the `#` character, but not `#/`. E.g. `#src/*` is OK.
2. The `value`s must be single strings. No arrays.
3. The first matching `key` is used, not the longest.

## TS is adopting `package.json` from TS 5.4

In the next version, Typescript will automatically pick up the "import path" configuration from `package.json`. The option to manually configure `tsconfig` will also remain.

## What a simple implementation can look like.

Since there is at most a single `*` character in the paths, there is no need for RegEx.

Assuming there are no arrays, the code could looks something like this:

```ts
function getActualPath(pathConfig, importPath) {
  for (const [key, value] of Object.entries(pathConfig)) {
    if (key.includes('*')) {
      const [before, after] = key.split('*');
      if (importPath.startsWith(before) && importPath.endsWith(after)) {
        const wildcard = importPath.slice(
          before.length,
          after.length > 0 ? -after.length : undefined
        );
        return value.split('*').join(wildcard);
      }
    } else if (key === importPath) {
      return value;
    }
  }
}
```

This code can be extended a bit for the case where value is an Array to return the first path that actually exists, but I'm happy to skip that for now.

rayan1810: Hey @nmn, this was really helpful. Also addressed all the changes requested [previously](https://github.com/facebook/stylex/pull/202#pullrequestreview-1785432665).

nmn: Let me know when you consider the PR ready again.

rayan1810: @nmn you can review this, I've also updated the docs api reference as well.

nmn: I've taken a look and it looks good. I'll come back to this in a few days. As I'm gonna be taking a break for New Years!

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -75,6 +77,63 @@ type StyleXStateOptions = $ReadOnly<{
   ...
 }>;
 
+function validateAliases(aliases: any): StyleXOptions['aliases'] {
```
rayan1810, line 80: Adding any since options has been defined as object of string: mixed, causing issues with the aliases which should be type object but are getting assigned mixed.

rayan1810: This should not break any typing since this is just a validation function and returns the currect types for the aliases object.

nmn: Use `$FlowFixMe` instead of `any`. Other than that, I'll add validation for this. I'm also looking into using `zod` or something similar instead of the hand-rolled solution I have right now.

I'm looking for something with good error messages.

rayan1810: Sure @nmn, I've updated the typing for now. I'll also check into zod and try to add validation using that while you are looking into it.

nmn:
@rayan1810 Sorry for the churn, but could you rebase this again? I'm getting ready to merge this.

rayan1810:
@nmn do you want me to remove the validation function from the aliases for now?

nmn:
@rayan1810 Yeah, I can add validation after merging using the same framework as I'm using otherwise.

Remember to rebase onto the `release/v0.5.0` branch.

rayan1810:
@nmn I've resolved the conflicts and removed the validation function. Had to add `// $FlowFixMe` because of a mismatch in options typings and aliases types. If you have any suggestions for that right now I'm happy to implement.

nmn:
@rayan1810 Thanks! I'll take from here.

Issue #132, opened 2023-12-12
[Webpack plugin] CSS file is not getting attached to html template.


omkar-nath:
**The problem**
I am not using create-react-app and trying with webpack configuration in a dummy react project.
I am trying StyleX with my webpack configured dummy react application.  My webpack config looks like below.

```js
const StylexPlugin = require("@stylexjs/webpack-plugin");
const HtmlWebpackPlugin = require("html-webpack-plugin");
const MiniCssExtractPlugin = require("mini-css-extract-plugin");
const path = require("path");
function config(env, argv) {
  return {
    entry: {
      main: "./src/index.js",
    },
    output: {
      path: path.resolve(__dirname, "build"),
      filename: "[name].js",
    },
    module: {
      rules: [
        {
          test: /\.js$/,
          exclude: /node_modules/,
          use: {
            loader: "babel-loader",
            options: {
              presets: ["@babel/preset-env", "@babel/preset-react"],
            },
          },
        },

        {
          test: /\.css$/i,
          use: [MiniCssExtractPlugin.loader, "css-loader"],
        },
      ],
    },

    plugins: [
      // Ensure that the stylex plugin is used before Babel
      new StylexPlugin({
        filename: "omkar.css",
        // get webpack mode and set value for dev
        dev: argv.mode === "development",

        // Use statically generated CSS files and not runtime injected CSS.
        // Even in development.
        // optional. default: 'x'
        classNamePrefix: "x-om",
        // Required for CSS variable support
        unstable_moduleResolution: {
          // type: 'commonJS' | 'haste'
          // default: 'commonJS'
          type: "commonJS",
          // The absolute path to the root directory of your project
          rootDir: path.join("__dirname"),
        },
      }),
      new MiniCssExtractPlugin(),
      new HtmlWebpackPlugin({
        inject: "head",
        template: path.join(__dirname, "public", "index.html"),
      }),
    ],
  };
}
module.exports = config;

```

When I am doing production build. I can see seperate CSS file generated but it is not attached to the index file . Am I missing something?

**Expected behavior**

The html file under build should be having css file attached to it.

**Steps to reproduce**
Clone: https://github.com/omkar-nath/styleX-react/blob/master/webpack.config.js
Run: npm run build

**Environment (include versions). Did this work in previous versions?**

nmn:
> When I am doing production build. I can see seperate CSS file generated but it is not attached to the index file . Am I missing something?

[...]

PR #203, opened 2023-12-17
Extract webpack plugin

### LINKED ISSUE #197: StyleX Webpack Plugin
**The problem**

I tried using StylexPlugin combined with HtmlWebpackPlugin to showcase my page.

The generated Css file is not associated with the html. I'm not sure if I still need to use StylexPlugin after using @stylexjs/babel-plugin.

I just want to have css link injection of stylex output in the output HTML as well, But there is no such thing now.

I looked up @stylexjs/webpack-plugin's code, Maybe you can try to associate HTMLwebpackPlugin's beforeEmit hook here?

image

To be honest, I have not carefully studied the internal mechanism of @stylexjs/webpack-plugin, and I am not sure whether the compiled .tsx file will retain the introduction of css, At least as a result, the import of css is not retained in .tsx, so that output html cannot introduce stylex'css file into the template.

**How to reproduce**

This is a simple example:

[project.zip](https://github.com/facebook/stylex/files/13692411/project.zip)

Steps to reproduce:
1. run `pnpm install`.
2. run `npm run dev` or `npx webpack`.
3. dist/index.html  - No style file injection.

**Expected behavior**

I  hope to be able to inject style files.

minhtranite:

This PR to add a new webpack plugin that propose a new approach to extract StyleX styles and create css file for each webpack entry. The new plugin can help resolve some limitions of current plugin:
- It allows collect and create css file for each webpack entry instead of create a single css file for all entries like current webpack plugin.
- It support use [name], [contenthash] in filename.
- By use this new plugin with `html-webpack-plugin, the css files auto inject to HTML.

I created a new plugin instead of modify the current plugin because I see the approach are difference. But I'm happy to bring and merge the changes to current plugin if needed.

## Linked PR/Issues

Fixes #197, #132, and #105

## Pre-flight checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7767,6 +7767,10 @@
       "resolved": "packages/eslint-plugin",
       "link": true
     },
+    "node_modules/@stylexjs/extract-webpack-plugin": {
+      "resolved": "packages/extract-webpack-plugin",
+      "link": true
+    },
     "node_modules/@stylexjs/nextjs-plugin": {
       "resolved": "packages/nextjs-plugin",
       "link": true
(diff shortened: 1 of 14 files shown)


nmn:
> It allows collect and create css file for each webpack entry instead of create a single css file for all entries like current webpack plugin.

We don't want this. We always want a single CSS file. Please see https://stylexjs.com/docs/learn/thinking-in-stylex/#one-small-file-over-many-smaller-files

The other changes are in the right direction, but we will need to ensure that the common case still works correctly.

minhtranite:
> We don't want this. We always want a single CSS file. Please see https://stylexjs.com/docs/learn/thinking-in-stylex/#one-small-file-over-many-smaller-files

Thank for your feedback! It seams that that the approach used in this PR are not fit with StyleX principles.

PR #210, opened 2023-12-18
Bug/overrides for token

develop084:

The `OverridesForTokenType` type was updated to allow more flexibility when defining overrides in the `stylex.createTheme` function. The property names are no longer restricted to a specific type, providing greater flexibility for direct use of string literals.

## Linked PR/Issues

Fixes # (replace with the relevant issue number, if applicable)

## Additional Context

No additional context is necessary for this change.

## Pre-flight checklist

### DIFF
--- a/packages/stylex/src/StyleXTypes.js
+++ b/packages/stylex/src/StyleXTypes.js
@@ -177,11 +177,12 @@ export type Theme<
   [string]: StyleXClassNameFor<string, IDFromVarGroup<T>>,
 }>;
 
-export type OverridesForTokenType<Config: { +[string]: mixed }> = {
+export type OverridesForTokenType<Config extends { [key: string]: unknown }> = {
   [Key in keyof Config]:
     | Config[Key]
-    | { +default: Config[Key], +[string]: Config[Key] },
+    | { default: Config[Key]; [property: string]: Config[Key] };
 };
+
 
 export type StyleX$CreateTheme = <
   BaseTokens: VarGroup<{ +[string]: mixed }>,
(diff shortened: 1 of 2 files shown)


packages/stylex/src/StyleXTypes.js
```diff
@@ -176,12 +177,13 @@ export type Theme<
   [string]: StyleXClassNameFor<string, IDFromVarGroup<T>>,
 }>;
 
-export type OverridesForTokenType<Config: { +[string]: mixed }> = {
+export type OverridesForTokenType<Config extends { [key: string]: unknown }> = {
   [Key in keyof Config]:
     | Config[Key]
-    | { +default: Config[Key], +[string]: Config[Key] },
+    | { default: Config[Key]; [property: string]: Config[Key] };
```
nmn, line 183: Revert this line.
`+` means `readonly` in Flow and is needed.
`[string]` is valid in Flow, and there is no need for the `property:` bit.
Please bring back the `+` here.

nmn (changes requested):
I'm not sure this fixes the initial problem, and I think it causes some regressions.

packages/stylex/src/StyleXTypes.js
develop084: Okay I will correct this

nmn: What is the purpose of your PR? Every change this PR breaks Flow by trying to change it to Typescript syntax.

develop084: @nmn  I thought the problem occurred , can be fixed by changing the type, it was my first try to contribute to open source I think I am not at the ryt place then.
Thanks

develop084: I will try to fix what's bad in my PR, I am sure that I'll learn something :)

nmn: @develop084 From what I understand, you're trying to change the types of `createTheme` to allow changing *some* of the variables instead of all of them.

I've already made this fix and it'll be out in the next release in a few days.

----

I don't want to discourage you from learning and contributing. I would suggest reading the Contribution.md document, and learn more about Flow at flow.org

Then, make sure you have the Flow plugin for VSCode installed and functional before you make your next attempt.

PR #220, opened 2023-12-20
feat: support default values in dynamic style functions

### LINKED ISSUE #198: Better errors to say that Dynamic Style functions don't support default values
Is there a limitation to the kind of arguments that can be passed as arguments to dynamic styles?

I keep getting the following error in the code below:

Error message:
`
common/components/ui/Flex/index.js: Only static values are allowed inside of a stylex.create() call.`

Here is my Flex component:

```javascript

import React from 'react';

    row: (moreProps = {}) => ({
        display: 'flex',
        flexDirection: 'row',
        ...moreProps
    }),
    column: (moreProps = {}) => ({
        display: 'flex',
        flexDirection: 'column',
        ...moreProps
    })
});

const Flex = ({
    direction = 'row',
    justifyContent = 'flex-start',
    alignItems = 'stretch',
    children,
    style
}) => {

    return (
        <div {...stylex.props(styles[direction]({ justifyContent, alignItems }), style)}>
            {children}
        </div>
    );
};

export default Flex;

```

What could be wrong and what's a better to fixing this?

nedjulius:

This PR adds support for dynamic function parameter default values.

```
  foo: (props = {}) => ({
    fontSize: props.size
  })
});
```

- [ ] Update the `eslint-plugin` accordingly
- [x] Add some tests

## Linked PR/Issues

Fixes #198 

## Additional Context

Currently, the default value assignment throws `NON_STATIC_VALUE` error and does not support assigning default value to the named parameter.

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
+++ b/packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
@@ -73,8 +73,11 @@ export function evaluateStyleXCreateArg(
       NodePath<t.Identifier | t.SpreadElement | t.Pattern>,
     > = fnPath.get('params');
     const params: Array<t.Identifier | t.AssignmentPattern> = allParams
-      .filter((param): param is NodePath<t.Identifier | t.AssignmentPattern> =>
-        pathUtils.isIdentifier(param) || pathUtils.isAssignmentPattern(param),
+      .filter(
+        (param): param is NodePath<t.Identifier | t.AssignmentPattern> =>
+          pathUtils.isIdentifier(param) ||
+          (pathUtils.isAssignmentPattern(param) &&
+            t.isIdentifier(param.node.left)),
       )
       .map((param) => param.node);
     if (params.length !== allParams.length) {
--- a/packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
+++ b/packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
@@ -31,7 +31,7 @@ export function evaluateStyleXCreateArg(
   deopt?: null | NodePath<>,
   fns?: {
     [string]: [
-      Array<t.Identifier>,
+      Array<t.Identifier | t.AssignmentPattern>,
       { +[string]: t.Expression | t.PatternLike },
     ],
   },
@@ -43,7 +43,7 @@ export function evaluateStyleXCreateArg(
   const value: { [string]: mixed } = {};
   const fns: {
     [string]: [
-      Array<t.Identifier>,
+      Array<t.Identifier | t.AssignmentPattern>,
       $ReadOnly<{ [string]: t.Expression | t.PatternLike }>,
     ],
   } = {};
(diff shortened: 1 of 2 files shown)


packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
```diff
@@ -72,9 +72,9 @@ export function evaluateStyleXCreateArg(
     const allParams: Array<
       NodePath<t.Identifier | t.SpreadElement | t.Pattern>,
     > = fnPath.get('params');
-    const params: Array<t.Identifier> = allParams
-      .filter((param): param is NodePath<t.Identifier> =>
-        pathUtils.isIdentifier(param),
+    const params: Array<t.Identifier | t.AssignmentPattern> = allParams
+      .filter((param): param is NodePath<t.Identifier | t.AssignmentPattern> =>
+        pathUtils.isIdentifier(param) || pathUtils.isAssignmentPattern(param),
```
nmn, line 77: We need an additional check that the lhs of the assignment pattern is an identifier.

nmn (approved):
Good start!

necolas:
Can't the default value be set as a non-dynamic style? Or otherwise handled in JS before passing the dynamic value to stylex (e.g., default prop value or `?? defaultValue`)

nmn:
@necolas makes a good point. @nedjulius The task to fix this issue is going to be more complex than simply adding support for default values. This is going to be a challenging project and *not* a "good first issue".

----

Let me try to explain the problem:

Say you define dynamic styles with a single argument:

```ts
  d: (a) => ({color: a}),
});
```

This gets compiled to something that looks like this:
```ts
// This would be in the CSS file in prod:
inject('.jhsgfkj{color: var(--color)}');

  d: (a) => [
    {color: 'jhsgfkj', $$css: true},
    {'--color': a}
  ],
});
```

The first object defines a className, and the second object is a dynamic inline style.

If we supported default values in the obvious way, we'd be generating inline styles wastefully:

```ts
// This would be in the CSS file in prod:
inject('.jhsgfkj{color: var(--color)}');

  d: (a = 'red') => [
    {color: 'jhsgfkj', $$css: true},
    {'--color': a}
  ],
});
```

Where we should be generating something that looks like this:

```ts
// This would be in the CSS file in prod:
inject('.hfgdhkf{color: red;}');
inject('.jhsgfkj{color: var(--color)}');

  d: (a) => [
    {color: 'hfgdhkf', $$css: true},
    a && {color: 'jhsgfkj', $$css: true},
    a && {'--color': a}
  ],
});
```

Things only get more complicated with multiple arguments.

---

Therefore, I suggest we explicitly disallow default values within dynamic style functions and not support them at all. @nedjulius are you interested in updating your PR to provide better error messages instead?

nedjulius:
@nmn sounds good. I can work on better error messages. should I include an eslint rule for this?

necolas:
@nmn if the generated code looks like this, how are we preventing new `$$css` objects being created every function call (and causing cache misses)?

```js
  d: (a) => [
    {color: 'jhsgfkj', $$css: true},
    {'--color': a}
  ],
});
```

nmn:
@necolas That's a good catch! I'll work on hoisting those objects. @nedjulius Are you OK if I close this PR?

nedjulius:
@nmn yeah, I will close it. will work in a new branch

PR #222, opened 2023-12-20
fix : Handle absolute paths for custom inject source

### LINKED ISSUE #211: Handle absolute paths for alternate inject source
## Background

By default, when `runtimeInjection` is `true`, the function inject styles at runtime is imported as the default import from `'@stylexjs/stylex/lib/stylex-inject'`.

A recent change has added support for string and objects to be used as the value of `runtimeInjection` as well. 
- A string is used as the alternate import source for the inject function. 
- An object of the form `{from: string, as: string}` can be used for a named import from a custom source.

## Current behaviour

Today, any custom import source for the inject function is used as a literal string. This doesn't work well when relative file paths are needed which will be different in each file in a project.

## Expected Change

The `StateManager` class should be updated to detect relative or absolute paths and automatically generate the relative paths necessary from the file being transformed.

Here are the steps involved:

1. If the file is a relative file path. (Starts with `./` or `../`)
   - Convert to an absolute file path by `path.join` with the `rootDir` passed in.
2. Update the `get runtimeInjection()` getter function in `state-manager.js` to automatically convert any absolute file paths to relative file paths from the current file.
   - NOTE: Absolute file paths start with `/` on Unix systems, but may start with `C:\\` on Windows.

purohitdheeraj:

Handle absolute paths for custom inject source
checks for absolute and relative path

## Linked PR/Issues

Fixes #211 

## Additional Context

i will move forward with `importSources` once this gets approved

## Pre-flight checklist


packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -159,9 +159,28 @@ export default class StateManager {
...
+    const runtimeInjection = options.runtimeInjection;
+    const rootDir = this.options.unstable_moduleResolution.rootDir || '';
+    const filename = this.filename || '';
+
+    if (typeof runtimeInjection === 'string' && rootDir && filename) {
+      if (
+        runtimeInjection.startsWith('./') ||
+        runtimeInjection.startsWith('../')
+      ) {
+        const absolutePath = path.join(rootDir, runtimeInjection);
```
nonzzz, line 172: I think should add a normalize function to platform differences.

purohitdheeraj: you mean path.normalize(path.join()) ?

nonzzz: perhaps you can refer https://github.com/vitejs/vite/blob/5684fcd8d27110d098b3e1c19d851f44251588f1/packages/vite/src/node/utils.ts#L231-L233

necolas:
Closing old PRs

Issue #235, opened 2023-12-21
CSS parser failed to execute insertRule while using multiple rules string from defineVars


mjangir:
**The problem**
I was experimenting with stylex in a vite-react-ts application and tried defining vars for both light and dark theme and used them in stylex.create which is not working at all during the development. However, when I build my vite application and serve it, it works fine.

**How to reproduce**

1. Create a vite application with `npm create vite@latest my-app -- --template react-ts`

2. Create `Button.stylex.ts` with the following content:

```tsx
import * as stylex from "@stylexjs/stylex";

const DARK = "@media (prefers-color-scheme: dark)";

export const tokens = stylex.defineVars({
  bg: {
    default: "yellow",
    [DARK]: "green",
  },
});

```

3. In `App.tsx`, simple create a button style and apply it to the button

```tsx
import * as stylex from "@stylexjs/stylex";
import { tokens } from "./Button.stylex";

const buttonStyles = stylex.create({
  main: {
    background: tokens.bg,
  },
});

function App() {
  return (
    <div>
      <button {...stylex.props(buttonStyles.main)}>Submit</button>
    </div>
  );
}
```

4. Simply run `npm run dev` and the styles are not applied at all.

**Root case I found was**
When we use the media query in the `defineVars`, the rule is not properly inserted in the sheet as the CSS parser failed to execute multiple rules at a time. I tried to execute the same code in the browser console and got this:

<img width="1404" alt="image" src="https://github.com/facebook/stylex/assets/5525274/b477f218-4d26-4c2c-8951-911eb10f8a82">

It is going inside in an unhandled catch block here:

<img width="639" alt="image" src="https://github.com/facebook/stylex/assets/5525274/30c2f25d-bcb4-4bd7-93b2-17db78436d71">

**Related To**
https://stackoverflow.com/questions/40187198/is-adding-multiple-rules-to-css-using-insertrule-possible

nmn:
Thank you so much for finding this bug! This explains a lot! I'll fix this ASAP!

mjangir:
> Thank you so much for finding this bug! This explains a lot! I'll fix this ASAP!

I tried fixing this in the closed PR above but found it impacting a lot of code. So just closed it. Perhaps we may need  to inject each css selector separately in the runtimeInjection.

nmn:
This is fixed and will be released within a day.

PR #241, opened 2023-12-22
Fix: Insert individual rules for runtimeInjection

nmn:

When defining variables with `defineVars` and `createTheme`, the generated CSS was generated as a single string. When `@` rules were used this string contained multiple CSS rules.

In development, when `runtimeInjection` is `true`, this would try to insert the whole string with the browser's `insertRule` API. This would error.

This PR, breaks up the generated CSS into individual rules that can be injected separately at runtime (when in development)

**NOTE:** This is a fix for development workflows. Everything already worked for production builds.

## Linked PR/Issues

#235 

## Additional Context

Test snapshots have been updated and I have verified that everything looks right.


mjangir:
Thank you so much @nmn for addressing this issue so early. Great to see it in action.

PR #244, opened 2023-12-22
feat: improve error messaging in `stylex.create`

### LINKED ISSUE #198: Better errors to say that Dynamic Style functions don't support default values

nedjulius:

This PR improves the error messaging for `stylex.create` dynamic values. In particular, it throws a more verbose error when consumers attempt to set default parameters or don't use named parameters. Previously, in both cases it would just throw `NON_STATIC_VALUE` error. I also improved the error message in `eslint-plugin`.

Let me know if the error messages themselves can be improved.

image

## Linked PR/Issues

Fixes #198 

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
+++ b/packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
@@ -73,30 +73,14 @@ export function evaluateStyleXCreateArg(
     const allParams: Array<
       NodePath<t.Identifier | t.SpreadElement | t.Pattern>,
     > = fnPath.get('params');
+
+    validateDynamicStyleParams(allParams);
+
     const params: Array<t.Identifier> = allParams
       .filter((param): param is NodePath<t.Identifier> =>
         pathUtils.isIdentifier(param),
       )
       .map((param) => param.node);
-
-    if (
-      allParams.some(
-        (param) =>
-          pathUtils.isObjectPattern(param) || pathUtils.isRestElement(param),
-      )
-    ) {
-      throw new Error(
-        messages.ONLY_NAMED_PARAMETERS_IN_DYNAMIC_STYLE_FUNCTIONS,
-      );
-    }
-
-    if (allParams.some((param) => pathUtils.isAssignmentPattern(param))) {
-      throw new Error(messages.NO_DYNAMIC_STYLE_DEFAULT_PARAMETERS);
-    }
-
-    if (params.length !== allParams.length) {
-      return { confident: false, deopt: valPath, value: null };
-    }
 
     const fnBody = fnPath.get('body');
     if (!pathUtils.isObjectExpression(fnBody)) {
@@ -262,4 +246,21 @@ function evaluateObjKey(
     confident: true,
     value: String(key),
   };
+}
+
+function validateDynamicStyleParams(
+  params: Array<NodePath<t.Identifier | t.SpreadElement | t.Pattern>>,
+) {
+  if (
+    params.some(
+      (param) =>
+        pathUtils.isObjectPattern(param) || pathUtils.isRestElement(param),
+    )
+  ) {
+    throw new Error(messages.ONLY_NAMED_PARAMETERS_IN_DYNAMIC_STYLE_FUNCTIONS);
+  }
+
+  if (params.some((param) => pathUtils.isAssignmentPattern(param))) {
+    throw new Error(messages.NO_DYNAMIC_STYLE_DEFAULT_PARAMETERS);
+  }
 }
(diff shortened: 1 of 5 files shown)


packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
```diff
@@ -77,9 +78,27 @@ export function evaluateStyleXCreateArg(
...
+      throw new Error(
+        messages.ONLY_NAMED_PARAMETERS_IN_DYNAMIC_STYLE_FUNCTIONS,
+      );
+    }
+
+    if (
+      allParams.some(
+        (param) =>
+          pathUtils.isAssignmentPattern(param) &&
+          t.isIdentifier(param.node.left),
```
nedjulius, line 92: actually, I don't think this identifier check is necessary. any assignment is invalid

packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
```diff
@@ -77,9 +78,21 @@ export function evaluateStyleXCreateArg(
...
+      throw new Error(
+        messages.ONLY_NAMED_PARAMETERS_IN_DYNAMIC_STYLE_FUNCTIONS,
+      );
+    }
+
+    if (allParams.some((param) => pathUtils.isAssignmentPattern(param))) {
+      throw new Error(messages.NO_DYNAMIC_STYLE_DEFAULT_PARAMETERS);
+    }
+
     if (params.length !== allParams.length) {
```
nedjulius, line 92: I think this check is obsolete at this point. can't think of other corner cases besides ones covered above

nmn: You're right. Only simple named arguments are supported.

packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
```diff
@@ -72,14 +73,15 @@ export function evaluateStyleXCreateArg(
     const allParams: Array<
       NodePath<t.Identifier | t.SpreadElement | t.Pattern>,
     > = fnPath.get('params');
+
+    validateDynamicStyleParams(allParams);
```
nedjulius, line 77: hmm. this works fine when I just run transformation, but I tried to get these messages thrown via the webpack plugin and they get swallowed. I wonder if we should just propogate the error message to the `transformStyleXCreate` function and then throw it from there

nmn: The webpack plugin tends to swallow errors, let's focus on making the errors work when Babel is used directly in this PR. Making the bundler plugins handle those errors correctly is a separate task.

packages/babel-plugin/src/visitors/stylex-create/parse-stylex-create-arg.js
```diff
@@ -263,3 +247,20 @@ function evaluateObjKey(
     value: String(key),
   };
 }
+
+function validateDynamicStyleParams(
```
nmn, line 251: Why not simplify this function to validate that all params as `t.Identifier` instead?

For example, this pattern will silently ignore array destructuring. We don't really support *anything* other than simple named arguments, so let's just check for that.

nedjulius: shouldn't we differentiate between different cases and specify the error message with with what's not allowed?

otherwise it would still be an improvement even if the error specified that `only named parameters are allowed` -- i can simplify it to this as well

nmn: I think a single simple error message is more helpful, but perhaps add more details. So a single error which says:
`"Only named parameters are allowed. No destructuring. No default values."`

What do you think?

nedjulius: yeah, I think it makes sense and I agree. describing one allowed pattern is as effective as trying to describe multiple forbidden patterns, and has less boiler plate. I will restructure the PR.

nmn (approved):
This looks good! I will merge this within a day or two.

Thank you!

PR #248, opened 2023-12-23
feat: Remove style rules that are not needed

nmn:

At a high-level, the StyleX compiler works in the following steps:
1. All the function calls other than `stylex()` and `stylex.props()` are compiled down. `stylex.create` calls are replaced by object expressions that map keys to classNames.
2. We try to precompute `stylex()` and `stylex.props()` calls. Replacing them string constants if possible.
3. If all references to the compiled styles from step 1 are removed, we can remove the object entirely.

This PR improves the dead-code elimination from step 3. Now, when there are dynamic references that exist after compilation to style objects, only the style objects that are needed are kept, and other style rules within the same variable declaration can be removed.

TLDR; Extra sub-objects are removed from the compiled code.

## Additional Context

A couple of snapshots have been updated as a result of this change and I've added an additional test to demonstrate the change.

## Next Step

The style objects contain many properties with the value `null` in order to override previous styles. We should be able to remove all of these keys when a style object *only* occurs as the first argument of `stylex()` or `stylex.props()`.

Further, the second argument only needs to `null`-out properties that existed on previous style objects as arguments on the same function call and so on.

The metadata for doing this is already being collected in this PR. A follow-up PR will actually remove the `null` properties.

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-call-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-call-test.js
@@ -314,6 +314,53 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
 
+    test('stylex call keeps only the styles that are needed', () => {
+      expect(
+        transform(`
+          import stylex from 'stylex';
+          const styles = stylex.create({
+            foo: {
+              padding: 5
+            },
+            bar: {
+              paddingBlock: 10,
+            },
+            baz: {
+              paddingTop: 7,
+            }
+          });
+          stylex(styles.foo);
+          stylex(styles.foo, styles.bar);
+          stylex(styles.bar, styles.foo);
+          stylex(styles.foo, styles.bar, styles.baz);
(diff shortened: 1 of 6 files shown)


packages/babel-plugin/__tests__/stylex-transform-call-test.js
```diff
@@ -314,6 +314,53 @@ describe('@stylexjs/babel-plugin', () => {
...
+          stylex(styles.foo, somethingElse);
+        `),
+      ).toMatchInlineSnapshot(`
+        "import _inject from "@stylexjs/stylex/lib/stylex-inject";
+        import stylex from 'stylex';
+        _inject(".x14odnwx{padding:5px}", 1000);
+        _inject(".xp59q4u{padding-block:10px}", 2000);
+        _inject(".xm7lytj{padding-top:7px}", 4000);
+        const styles = {
+          foo: {
```
nmn, line 345: Previously, `bar` and `baz` would've been left behind as well.

PR #283, opened 2023-12-30
fix(@stylex/eslint-plugin): Change message when a property is disallowed

### LINKED ISSUE #282: @stylexjs/eslint-plugin: Confusing error message when a value is disallowed with `null`
**The problem**

When disallowing a property with the ESLint plugin, the error message is:

```
{property} value must be one of:
{reason}
```

While not completely incorrect, it's slightly misleading to say `value must be one of:` when there is no valid value.

**How to reproduce**

**ESLint configuration**

image

**Error message**

image

**Expected behavior**

I expect the error message to clearly indicate that the property is disallowed.

**Environment (include versions). Did this work in previous versions?**
```
@stylexjs/eslint-plugin@0.4.1
```

tjosepo:

- Moved the logic that display the line `{key} value must be one of:\n` further down the codebase to be able to display a different message when a property is disallowed. 
- Add a `Property {key} is not allowed.` message.

This results in error messages more specific to the error its reporting.

This is a very naive attempt from my part to solve the problem with minimal changes and there might be more maintainable solutions.

Let me know if you think this is the right approach.

Also, feel free to suggest text for the error message.

## Linked PR/Issues

Fixes #282

## Additional Context

Tests where modified to reflect the changes:

Before:

```
grid value must be one of:
grid properties disallowed for testing
```

After:
```
Property grid is not allowed. grid properties disallowed for testing
```

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/src/stylex-valid-styles.js
+++ b/packages/eslint-plugin/src/stylex-valid-styles.js
@@ -50,6 +50,7 @@ export type RuleCheck = (
   node: $ReadOnly<Expression | Pattern>,
   variables?: Variables,
   prop?: $ReadOnly<Property>,
+  key?: string,
 ) => RuleResponse;
 export type RuleResponse = void | {
   message: string,
@@ -62,7 +63,7 @@ export type RuleResponse = void | {
 
 const showError =
   (message: string): RuleCheck =>
-  () => ({ message });
+  (_node, _variables, _props, key) => ({ message: key ? `Property ${key} is not allowed. ${message}` : message });
 
 const isStringOrNumber = makeUnionRule(isString, isNumber);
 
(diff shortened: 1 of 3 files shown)


tjosepo:
I just signed the CLA.

packages/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -50,6 +50,7 @@ export type RuleCheck = (
   node: $ReadOnly<Expression | Pattern>,
   variables?: Variables,
   prop?: $ReadOnly<Property>,
+  key?: string,
```
tjosepo, line 53: I made this argument optional because I don't know how `RuleCheck` is being used.

Should I make it required?

nmn: This OK to start. We can make this stricter in the future if it makes sense.

PR #298, opened 2024-01-03
Fix: Respect runtimeInjection: false

nmn:

There is a bug right now where setting `runtimeInjection` to `false` doesn't actually work. This PR fixes that.

### DIFF
--- a/packages/babel-plugin/src/utils/state-manager.js
+++ b/packages/babel-plugin/src/utils/state-manager.js
@@ -97,13 +97,16 @@ export default class StateManager {
       dev: !!(options: $FlowFixMe).dev,
       test: !!(options: $FlowFixMe).test,
       runtimeInjection:
-        options.runtimeInjection === true
-          ? DEFAULT_INJECT_PATH
-          : options.runtimeInjection
-            ? options.runtimeInjection
-            : options.dev
-              ? DEFAULT_INJECT_PATH
-              : undefined,
+        // prettier-ignore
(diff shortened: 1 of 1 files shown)


PatrickJS:
A deeply nested ternary 🤔
LGTM

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -97,13 +97,16 @@ export default class StateManager {
...
       runtimeInjection:
-        options.runtimeInjection === true
-          ? DEFAULT_INJECT_PATH
-          : options.runtimeInjection
-            ? options.runtimeInjection
-            : options.dev
-              ? DEFAULT_INJECT_PATH
-              : undefined,
+        // prettier-ignore
+        options.runtimeInjection === true ?
```
nedjulius, line 101: maybe it would be helpful to make a local helper as such

```
function getRuntimeInjectionOptionValue() {
  if (options.runtimeInjection === true) {
    return DEFAULT_INJECT_PATH;
  }
  
  if (options.runtimeInjection === false) {
    return undefined;
  }
  
  if (options.runtimeInjection !== null) {
    return options.runtimeInjection;
  }
  
  if (options.dev) {
    return DEFAULT_INJECT_PATH;
  }
  
  return undefined;
}
```

nmn: I have another PR up with a bigger refactor of this.

PR #317, opened 2024-01-06
Remove unnecessary 'as const' assertion from constant declaration

odin-nido:

Removed redundant 'as const' assertions from constant declarations.


nmn:
The NextJS example is also used as a sanity check to ensure the types, compiler and lint rules are working as expected. The `as const` and other types are included as "tests" of a kind, so we don't want to remove them.

PR #318, opened 2024-01-07
Fix: The object position can also be in percentages or pixel values.

uttampun44:
## Object postion

We can use pixels and percentages in values as well. I have seen the only string we can use in numbers as well.


nmn (approved):
Object position is almost always described as `x y` value pair separated by a space. This needs to be wrapped in a string.

The *only* case that can be represented as a number is a single `px` value. Such a value only sets the `x` position and the `y` position is skipped.

So although, it's not common it could be valid to use a number. I will verify that numbers for `objectPosition` are correctly converted to px values before I merge.

Thanks for spotting this issue. Next time, please open an issue first.

PR #319, opened 2024-01-07
Release: v0.5.0-beta.1 branch

nmn:
This is a beta branch for now, do not merge into main.

### DIFF
--- a/apps/docs/docs/learn/03-installation.mdx
+++ b/apps/docs/docs/learn/03-installation.mdx
@@ -179,6 +179,10 @@ provides (experimental) plugins for Webpack, Rollup, Next.js and esbuild.
           runtimeInjection: false,
           genConditionalClasses: true,
           treeshakeCompensation: true,
+          // If you have opted for aliased paths in your Next.js App add it here as well
+          // aliases: { 
+          //   '@/*': ['./*'],
+          // },
           unstable_moduleResolution: {
             type: 'commonJS',
             rootDir: __dirname,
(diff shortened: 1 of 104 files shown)


apps/docs/docs/learn/03-installation.mdx
```diff
@@ -67,7 +67,7 @@ StyleX plugin. See the API reference for more details on the
 `@stylexjs/babel-plugin` API.
 
 To make this easier for commonly used packages and meta-frameworks, StyleX
-provides (experimental) plugins for Webpack, Rollup, and Next.js.
+provides (experimental) plugins for Webpack, Rollup, Next.js and esbuild.
```
necolas, line 70: oxford comma

```suggestion
provides (experimental) plugins for Webpack, Rollup, Next.js, and esbuild.
```

apps/docs/package.json
```diff
@@ -20,7 +20,7 @@
     "@docusaurus/preset-classic": "2.4.1",
     "@mdx-js/react": "^1.6.22",
     "@orama/plugin-docusaurus": "^1.2.4",
-    "@stylexjs/stylex": "0.4.1",
+    "@stylexjs/stylex": "0.5.0-beta.1",
```
necolas, line 23: Typically we don't use beta channels for other projects, and might instead publish canaries using SHA (or combination of SHA and date) suffixes if a build is being made available for early testing.

nmn: I'm still trying to figure out how to do a SHA hash. 

I didn't publish a `beta` yet, just a couple of `alpha`s.

necolas: > I'm still trying to figure out how to do a SHA hash.

I can't remember if we automatically do anything for React etc, but in my own projects I've just taken 8-9 chars from the git hash at the HEAD commit being canary-published since no extra commit is added for canary releases.

nmn: Oh! That's such a simple solution. Thanks! I'll do that from next time.

necolas:
FYI I tried the 0.5 alpha and got an error, not sure if it's fixed in this branch

image

`dev: true` & `runtimeInjection: true`

nmn:
@necolas That problem should have been fixed in `alpha.3`, if you're still seeing that problem I might need your help figuring out the root cause.

PR #328, opened 2024-01-10
Feat/bundle runtime

nmn:

This PR will change the build step for `@stylexjs/stylex` to prebundle dependencies. This will allow us to ship an ESM export for the runtime without waiting for `styleq` and `invariant` to updated first. This is somewhat urgent to solve an issue with Vite with `0.4.1` where the import for runtime injection isn't handled correctly.

We can remove this additional build step once the dependencies are updated.

### DIFF
--- a/packages/babel-plugin/rollup.config.mjs
+++ b/packages/babel-plugin/rollup.config.mjs
@@ -3,6 +3,8 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
+ *
+ *
  */
 
 import { nodeResolve } from '@rollup/plugin-node-resolve';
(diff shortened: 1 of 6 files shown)


nmn:
@nedjulius I think these changes should make the ESM package work correctly. Can you please take a look?

himself65:
This looks makes sense, I will try this today!

necolas:
Sorry didn't see the PR for styleq among all the stylex notifications!

FWIW I think there was some talk of moving off invariant in other projects, and React may already have done. Maybe we could replace invariant in stylex with our own simple error function?

nmn:
@necolas Yeah, there is only a single usage of `invariant` in the codebase, I'll get rid of it. Also, bundling seems to have worked out in the mean time to unblock ESM exports. So, no rush on `styleq`. There are a couple of failing signals that I struggled with a bit.

PR #331, opened 2024-01-10
feat: Add stylex.attrs for frameworks that expect 'class' over 'className'

nmn:

For frameworks such as Solid, Qwik, Svelte, Vue, etc. that require a `class` instead of `className`, there is now a first-class API called `stylex.attrs` which can be used in place of `stylex.attrs`.

`stylex.attrs` returns an object with a `class` key instead of a `className` key and pre-converts the value of `style` to a string as not every framework accepts objects for inline styles.

In all other ways, `stylex.attrs` is identical to `stylex.props`. At runtime it uses `stylex.props` internally. At compile time, it benefits from the same optimisations as `stylex.props`.

@PatrickJS


nmn:
Published to NPM as `0.5.0-alpha.1`

PR #338, opened 2024-01-11
📝 Updated stylex.create docs

### LINKED ISSUE #327: Safe typings for TypeScript v4
**Is your feature request related to a problem? Please describe.**
The problem is the one mentioned in this issue https://github.com/facebook/stylex/issues/313, where the TypeScript `const` modifier is used (https://github.com/microsoft/TypeScript/pull/51865) which was introduced in v5 from what I can see.

**Describe a solution you'd like**
Some way to make the types safe for earlier versions of TypeScript

**Describe alternatives you've considered**
If that is not possible, than atleast some kind of warning somewhere in the docs should be present to indicate the problem

**Additional context**

image

amjadorfali:

Documentation changes, added a warning for typescript users using TS < 5

## Linked PR/Issues

Fixes #327 

## Additional Context
Referenced typescript limitation in using the new `const` keyword

## Pre-flight checklist


amjadorfali:
@nmn Here's the PR, I wasn't sure about the references but I added them anyway. Please let me know if I missed anything

apps/docs/docs/api/javascript/create.mdx
```diff
@@ -33,3 +33,13 @@ const styles = stylex.create({
   }),
 });
 ```
+
+:::danger Requires Typescript 5+
+
+Since `stylex.create` uses the new `const` keyword in its type definition, it requires Typescript 5+.
```
nmn, line 39: Just say that StyleX requires Typescript 5 or later. The rest of the context is unnecessary.

nmn (changes requested):
Let's also just wait since I might be able to solve this problem.

apps/docs/docs/api/javascript/create.mdx
necolas: I think requirements like version of TS or Node should be listed in the "installation" page rather than docs for specific APIs

necolas (changes requested):

apps/docs/docs/api/javascript/create.mdx
amjadorfali: I tackled the comments. Should i resolve this conversation or should the one who opened it do it?

apps/docs/docs/learn/03-installation.mdx
```diff
@@ -25,6 +25,9 @@ The recommended way to use StyleX in development and production is with the
 build-time compiler.
 
 ### Development
+:::danger TypeScript Limitation
```
nmn, line 28: The title should be changed. I'll think about what it should be.

apps/docs/docs/learn/03-installation.mdx
```diff
@@ -25,6 +25,9 @@ The recommended way to use StyleX in development and production is with the
...
+StyleX requires Typescript 5 or later
```
hamirmahal, line 29:
```suggestion
StyleX requires TypeScript 5 or later.
```

nmn:
I'm closing this PR. I'll update the docs with the minimum typescript requirement.

PR #355, opened 2024-01-16
fix: Use a variable for inject function

nmn:
This is to solve a conflict that occurs when using runtimeInjection with the `@babel/plugin-transform-modules-commonjs` plugin.

This is a workaround, since the bug is a consequence of how Babel works.


nmn:
This is published as `0.5.0-alpha.4`

This fixes the conflict with `@babel/plugin-transform-modules-commonjs`.

necolas (approved):

PR #365, opened 2024-01-19
fix: rollup watch mode shouldn't clean cache

### LINKED ISSUE #363: [bug]: build stylex on rollup watch mode
### Describe the issue

I defined a set of variables via stylex
image
Then I build via rollup, below is my configuration
image
dev scripts:
image
When in rollup development mode, the style packaging is normal when building for the first time, and the defined variables can be output correctly.
image
But when I modify the file, the variable disappears
image

### Expected behavior

The style output in rollup watch mode contains defined variables

### Steps to reproduce

1. clone https://github.com/shaxutang/sx-theme.git
2. pnpm install
3. pnpm dev
4. check the `dist/esm/index.css` file, the variable exists at this time
5. Modify the style content in the `src/button.ts` file
6. check the `dist/esm/index.css` file again. At this time, the variable has no output.

### Test case

_No response_

### Additional comments

_No response_

nonzzz:

Add disable clean cache logic for rollup plugin in watch mode.

## Linked PR/Issues

Fixes #363 

## Additional Context

In Rollup watch mode, when a file is modified, the Rollup instance is reloaded, and all hooks are rerun. Currently, we clean the CSS record at buildStart so that in watch mode, only the modified file is recorded. However, other files that contain StyleX logic may be missing from the record.

First load

image

After modify

image

## Pre-flight checklist

### DIFF
--- a/packages/rollup-plugin/src/index.js
+++ b/packages/rollup-plugin/src/index.js
@@ -41,10 +41,18 @@ export default function stylexPlugin({
   ...options
 }: PluginOptions = {}): Plugin<> {
   let stylexRules: { [string]: $ReadOnlyArray<Rule> } = {};
+  let isWatchMode = process.argv.some((c) => ['--watch', '-w'].includes(c));
   return {
     name: 'rollup-plugin-stylex',
+    options(option) {
+      if (option.watch && !isWatchMode) {
+        isWatchMode = true;
+      }
+    },
     buildStart() {
-      stylexRules = {};
+      if (!isWatchMode) {
+        stylexRules = {};
+      }
     },
     generateBundle(this: PluginContext) {
       const rules: Array<Rule> = Object.values(stylexRules).flat();
--- a/packages/rollup-plugin/src/index.js
+++ b/packages/rollup-plugin/src/index.js
@@ -31,6 +31,17 @@ export type PluginOptions = $ReadOnly<{
   ...
 }>;
 
+function filterByImportSource(
+  importSources: Options['importSources'],
+  code: string,
+) {
+  return importSources.some((importName) =>
+    typeof importName === 'string'
+      ? code.includes(importName)
+      : code.includes(importName.from),
+  );
+}
+
 export default function stylexPlugin({
   dev = IS_DEV_ENV,
   unstable_moduleResolution = { type: 'commonJS', rootDir: process.cwd() },
@@ -41,18 +52,10 @@ export default function stylexPlugin({
   ...options
 }: PluginOptions = {}): Plugin<> {
   let stylexRules: { [string]: $ReadOnlyArray<Rule> } = {};
-  let isWatchMode = process.argv.some((c) => ['--watch', '-w'].includes(c));
   return {
     name: 'rollup-plugin-stylex',
-    options(option) {
-      if (option.watch && !isWatchMode) {
-        isWatchMode = true;
-      }
-    },
     buildStart() {
-      if (!isWatchMode) {
-        stylexRules = {};
-      }
+      stylexRules = {};
     },
     generateBundle(this: PluginContext) {
       const rules: Array<Rule> = Object.values(stylexRules).flat();
@@ -76,13 +79,7 @@ export default function stylexPlugin({
       return false;
     },
     async transform(inputCode, id): Promise<null | TransformResult> {
-      if (
-        !importSources.some((importName) =>
-          typeof importName === 'string'
-            ? inputCode.includes(importName)
-            : inputCode.includes(importName.from),
-        )
-      ) {
+      if (!filterByImportSource(importSources, inputCode)) {
         // In rollup, returning null from any plugin phase means
         // "no changes made".
         return null;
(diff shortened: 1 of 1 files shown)


packages/rollup-plugin/src/index.js
```diff
@@ -41,10 +41,18 @@ export default function stylexPlugin({
   ...options
 }: PluginOptions = {}): Plugin<> {
   let stylexRules: { [string]: $ReadOnlyArray<Rule> } = {};
+  let isWatchMode = process.argv.some((c) => ['--watch', '-w'].includes(c));
```
nmn, line 44: This is not the right way to detect watch mode. Try using `options`.

packages/rollup-plugin/src/index.js
```diff
@@ -41,10 +41,18 @@ export default function stylexPlugin({
...
     name: 'rollup-plugin-stylex',
+    options(option) {
+      if (option.watch && !isWatchMode) {
+        isWatchMode = true;
+      }
+    },
     buildStart() {
-      stylexRules = {};
+      if (!isWatchMode) {
+        stylexRules = {};
```
nmn, line 54: The challenge with this approach is that it won't detect deleted files anymore.

nmn: `shouldTransformCachedModule` should be generating the CSS again anyway. Need to debug why its not.

nonzzz: Actually,`shouldTranformCacheModule` is a perf hook for rollup in `watch` mode, as far as i can know this hook does not affect the result. when i debug this case and try comment out `shouldTransformCacheModule` logic nothing be changed. In my mind, watch mode alway use in user development mode, so the dead code elimination is unnecessary. If we need it we can using result code import statments to rewrite all `stylexRules` in watch mode.

nonzzz: oh i know. i find another way to resolve this.

nonzzz: I removed the logic for clearing the cache and instead used `this.load` to re-insert the stylesheet. I think this is correct and consistent with the behavior of build.

packages/rollup-plugin/src/index.js
```diff
@@ -107,6 +112,27 @@ export default function stylexPlugin({
...
+            if (filterByImportSource(importSources, stmt.source.value)) {
+              // $FlowExpectedError[object-this-reference]
+              const resolved = await this.resolve(stmt.source.value, id);
+              if (resolved && !resolved.external) {
+                // $FlowExpectedError[object-this-reference]
+                const result = await this.load(resolved);
+                if (result) {
+                  stylexRules[resolved.id] = (result.meta: $FlowFixMe).stylex;
+                }
+              }
```
nmn, line 130: Can you write some comments explaining what is going on here?

AFAIK, the intention is to ensure that the generated styles from `.stylex.js` files are included in the collected styles. I think this is needed because after compilation, this import becomes unneccessary and Rollup strips it out.

A simpler solution might be to add the `treeshakeCompensation` option to the Babel plugin which inserts a side-effect-y import which won't be stripped out.

nmn (changes requested):
Waiting for a response before this can be merged.

packages/rollup-plugin/src/index.js
nonzzz: However,I tested the `treeshakeCompensation` option, But it did not affect the results. After my test, I discovered that when I modify a file, only the modified file triggers Rollup. Consequently, Rollup re-calls the buildStart, transform, or other hooks. However, the current behavior records only the modified file, preventing styleRules from capturing all the necessary information. To address this limitation, I added the following logic: I load the information from the Rollup cache and re-insert it.
> Why  `!resovled.external`
- external means `node_modules` or other need exclude from bundle.

nmn: What is happening on this line: `if (filterByImportSource(importSources, stmt.source.value)) {`

I believe this only works because it happens to match files that end with `.stylex`, however, if you're looking for those files, this is the wrong function to use for it.

----

I'm still trying to make sense of the rest of the logic. But my understanding is this: 
- If a file is part of the bundle, but then removed from the bundle.
- Then a different file imports that file
- Then that dependency is unchanged but should be part of the bundle again.
- Your code here manually goes through imports of the newly added/edited file and ensures they're included

Right?

nonzzz: Let me explain why need'filterByportSource',AFAIK only include '.stylex.ext' can be handle. Oh maybe i ignored the follow case like define Theme. I only consider define variable.
About the issue listed, This logic only work at watch mode, Normally rollup will automatically trace them but sometimes rollup smart algorithms can not handle them.
(Although i manually them but the transform result isn't be change.Only load the cached style and re-insert) So in my option those logic are just to ensure the  missing style rule can be insert.

nonzzz: Here is some comment that after my test. Like `stylex.defineVariable` will generate 
```js
export const xx = {
 "variable":"var(--xx-hash)"
}

```
and other files who import those variables. According [Rollup Document](https://rollupjs.org/plugin-development/#this-load) I think the logic can work well.

nmn: Thanks for the PR. I'll have to spend some time with this to make sense of what is going on.

nmn:
Merged in 20c637be9f3c447

PR #366, opened 2024-01-19
feat: add eslint sort keys plugin

### LINKED ISSUE #287: Plugin for sorting style keys
Tailwind has such a plugin: https://tailwindcss.com/blog/automatic-class-sorting-with-prettier.

Is it a good idea to have something similar for StyleX? 🤔

nedjulius:

Adds sort keys plugin for StyleX.

Sorts inside `create` and `keyframes` first argument object, and only starting from the second level nesting (as we only want to sort the style properties, not the namespaces).

```typescript
import { create, keyframes } from 'stylex';

const someAnimation = keyframes({
  '0%': {
    opacity: 0,
    // The rule will report that `display` should be above `opacity`
    display: 'hidden',
  },
  '100%': {
    display: 'block',
    opacity: 1,
  };
});

const styles = create({
  main: {
    display: 'flex',
    // The rule will report that `borderColor` should be above `flex`
    borderColor: 'red'
  }
});
```

## Linked PR/Issues

Fixes #287 

## Additional Context

## Pre-flight checklist


nedjulius:
alright, this is what I have so far. there's still some edge cases to go over and some types to fix, but you can see my overall take on the issue. currently, this rule doesn't offer auto-fixes but we can add that later as we nail down the basic sort requirements.

the entire rule is just a fork of original sort keys from eslint which only works within StyleX `create` calls and has a desc priority comparison for style keys. I took the pragmatic approach and decided to track the nesting level of object expressions and whether the current object is within a StyleX `CallExpression`. simply, we only evaluate if we are within StyleX `create` call and if we are not in the root object expression.

not sure if we want the line separated groups option or the minimum keys option?

packages/eslint-plugin/src/stylex-sort-keys.js
```diff
@@ -0,0 +1,291 @@
...
+        node.arguments.length === 1
+      );
+    }
+
+    let stack: Stack = null;
+    let isInsideStyleXCreateCall = false;
+    let objectExpressionNestingLevel = -1;
+
+    return {
+      ImportDeclaration(node: ImportDeclaration) {
```
nedjulius, line 128: just realized that we should do the same with `require`

nmn:
> the entire rule is just a fork of original sort keys from eslint

The whole plugin can just be that with a custom sort comparator.

packages/eslint-plugin/src/stylex-sort-keys.js
nmn: It might be OK to skip that. `require` is not fully supported. Theming only works with ESM for example.

packages/eslint-plugin/src/stylex-sort-keys.js
```diff
@@ -0,0 +1,292 @@
...
+
+          if (
+            specifier.type === 'ImportSpecifier' &&
+            specifier.imported.name === 'create'
+          ) {
+            styleXCreateImports.add(specifier.local.name);
+          }
+        });
+      },
+      CallExpression(node: { ...CallExpression, ...Rule.NodeParentExtension }) {
```
nmn, line 157: Make sure to wrap every object type with `$ReadOnly`. This will prevent most of the type errors that tend to trip people up.

packages/eslint-plugin/src/utils/getPropertyPriority.js
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+'use strict';
+
+const PSEUDO_CLASS_PRIORITIES = {
```
nmn, line 12: You can just import these values from the `shared` package directly. Making a copy is dangerous as things may go out of sync over time.

nedjulius: according to the `getPriority` function, all pseudo classes have priority over at rules except for `@supports`. should that reflect in the sort as well?

nmn: Yes, all keys should be sorted by priority order in objects!

packages/shared/src/utils/property-priorities.js
```diff
@@ -589,7 +589,7 @@ longHandLogical.add('math-style');
 // CSS Pointer Events
 longHandLogical.add('touch-action');
 
-const PRIORITIES = {
+export const PSEUDO_CLASS_PRIORITIES: { [key: string]: number } = {
```
nedjulius, line 592: these types seem very hacky, but for some reason flow is unable to parse the types of string literals as object keys. I tried to create separate types of unions of literals that are in each of these priority objects, but then I would have trouble passing `prop` to the object, because `string` is not compatible with the union. do you have any other ideas on how to better approach this?

nmn: Flow requires manual annotations for exports.

So technically, you'd need to copy the object and use it as a type as well. Also, always remember to wrap objects in `$ReadOnly`.

packages/eslint-plugin/src/utils/getPropertyPriorityAndType.js
```diff
@@ -0,0 +1,55 @@
...
+
+  if (key.startsWith('@media')) {
+    return { priority: AT_RULE_PRIORITIES['@media'], type: 'atRule' };
+  }
+
+  if (key.startsWith('@container')) {
+    return { priority: AT_RULE_PRIORITIES['@container'], type: 'atRule' };
+  }
+
+  return { priority: 1, type: 'string' };
```
nedjulius, line 54: if the property doesn't fall into any of the clauses, we assume it should be compared as string

nmn:
This is looking good!

nmn (approved):

PR #382, opened 2024-01-26
Fix: Ensure 'inject' variable is defined before use

nmn:

A previous change ensures that the `inject` function is imported and then aliased as a variable before being used.
This change is a workaround for the way Babel deals with ES6 imports.

Previously, this `inject` alias variable was inserted after the last import statement. However, this breaks in some edge-cases where import statements may exist *after* `inject` has already been used.

This PR changes the behaviour to add the variable alias immediately after the `inject` function is imported.


AbhiPartha:
Thank you so much @nmn I was able to reproduce this issue and just when i was about to raise a bug you had already fixed it. PreactX + StyleX + Storybook combination now works out of the box. ;)

Issue #383, opened 2024-01-26
lineHeight: "42px" not valid


steida:
### Describe the issue

 Types of property 'lineHeight' are incompatible.
      Type 'StyleXClassNameFor<"lineHeight", "42px">' is not assignable to type 'StyleXClassNameFor<"lineHeight", Readonly<number | "initial" | "inherit" | "unset" | null | undefined>>'.

### Expected behavior

It should work, or is there any reason why it should not?

### Steps to reproduce

Define a style with lineHeight: "42px",

yousoumar:
Hi @steida! I created this [PR](https://github.com/facebook/stylex/pull/384) to solve the issue. Same issue as #356.

PR #384, opened 2024-01-26
fix: add support for string values for some properties

yousoumar:

Add support for string values for `lineHight` and `borderSpacing`. Fixes issue #383.


nedjulius:
This change won't be reflected in the eslint plugin. You can update it here:
https://github.com/facebook/stylex/blob/77b768ed90e88b422065eb118cccd8426cf5d15e/packages/eslint-plugin/src/stylex-valid-styles.js#L929
and here:
https://github.com/facebook/stylex/blob/77b768ed90e88b422065eb118cccd8426cf5d15e/packages/eslint-plugin/src/stylex-valid-styles.js#L598

the change for both would just be union rule of number, absolute and relative unit rules:
```typescript
const borderSpacing = makeUnionRule(isNumber, isLength)
```

nmn (approved):

yousoumar:
Hi @nedjulius and @nmn ! I'm a lit bit confused on why it seems like there is a duplication of type definition (one in Stylex package and one in ESLint package). I thought that ESLint would pick up the types in Stylex. Is that the case only with TypeScript types and not Flow types ? And when we need to update both packages for type problem and when not (if there is a case)?

nmn:
@yousoumar The two are completely different. The `stylex` package contains Flow/Typescript types. The ESLint rule contains the actual implementation of an ESLint rule. The ESLint rule has been implemented in a way that seems to resemble types, but it's actually full Javascript functions which can do much more sophisticated checking.

For example, even though today the ESLint rule allow any arbitrary string for `color`, we can do full validation of the value to make sure it's a valid color. I haven't been able to prioritize this work, but I've been working on a [full project](https://github.com/nmn/css-value-parser) for parsing CSS values which can eventually be used in the ESLint rule to do more robust validation.

----

The only real duplication is of `flow_modules` folders in the Babel and ESLint packages. This is due to a limitation of our scripts and we can fix this in the future.

yousoumar:
I see better now. Thanks for the elaborated answer @nmn!

PR #392, opened 2024-01-27
fix: fix eslint sort-keys handling nested properties

### LINKED ISSUE #391: `sort-keys` comparing `default` with parent property
### Describe the issue

The new `sort-keys` eslint rule seems to give some unexpected errors, suggesting it's comparing the inner keys with parent keys.

image

### Expected behavior

It's expected that the test case below is valid, however it errors with the message

> StyleX property key "default" should be above "maxWidth".

### Steps to reproduce

`@stylexjs/eslint-plugin@0.5.1`

- Add new test case in `packages/eslint-plugin/__tests__/stylex-sort-keys-test.js` with the below test case as a valid case
- Run `npm run test -w @stylexjs/eslint-plugin` and observe test failure with message `StyleX property key "default" should be above "maxWidth"`

### Test case

```js
      import stylex from 'stylex';
        nav: {
          maxWidth: {
            default: "1080px",
            ["@media (min-width: 2000px)"]: "calc((1080 / 24) * 1rem)"
          },
          paddingVertical: 0,
        },
      });
```

### Additional comments

_No response_

QingqiShi:

Fix a bug where the `sort-keys` rule is giving errors for nested keys, comparing them to the parent keys.

## Linked PR/Issues

Fixes #391 

## Additional Context

I *think* the issue here is that, after adding the outer object to the stack, subsequent nested objects are never actually pushed onto this stack, resulting in a stack that maintains only a single layer.

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/src/stylex-sort-keys.js
+++ b/packages/eslint-plugin/src/stylex-sort-keys.js
@@ -169,7 +169,7 @@ const stylexSortKeys = {
           objectExpressionNestingLevel++;
         }
 
-        if (objectExpressionNestingLevel === 1) {
+        if (objectExpressionNestingLevel > 0) {
           stack = {
             upper: stack,
             prevNode: null,
--- a/packages/eslint-plugin/__tests__/stylex-sort-keys-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-sort-keys-test.js
@@ -147,6 +147,19 @@ eslintTester.run('stylex-sort-keys', rule.default, {
         });
       `,
     },
+    {
+      code: `
+      import stylex from 'stylex';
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin/src/stylex-sort-keys.js
```diff
@@ -169,7 +169,7 @@ const stylexSortKeys = {
           objectExpressionNestingLevel++;
         }
 
-        if (objectExpressionNestingLevel === 1) {
+        if (objectExpressionNestingLevel > 0) {
```
nedjulius, line 172: nice catch! thanks for fixing

nmn (approved):
nodding-meme

PR #412, opened 2024-02-02
A CLI to compile an entire folder with StyleX

nmn:

## Expected Behaviour

The CLI will then:
- [ ] Transform every single JS file within the `source/` folder with the StyleX Babel plugin and output it at the same relative path in the `src/` folder.
- [ ] While transforming it will collect the generated CSS for each file and output a CSS bundle at `src/stylex-bundle.css`.
- [ ] The transformed JS files should be updated to include an import of the `stylex-bundle.css` file. (Configurable)

- [ ] Have a `watch` mode so that only the files that are edited are recompiled, and the generated CSS file is updated accordingly.
- [ ] Can be configured to also pre-compile a given list of `node_modules` packages and write the compiled files in some locations.
  - [ ] Any processed JS files that import these packages, should be updated to rewrite their imports to point to the compiled location.

- [ ] The basic configuration options should be possible with both CLI args and with a file config.
- [ ] It should be possible to configure the StyleX Babel plugin *without* a `.babelrc` file at the project root. (Next goes into Babel mode if it detect a `.babelrc` file)
- [ ] It should be possible to run additional Babel transforms in addition to the StyleX Babel plugin

## Stretch Goals
- [ ] Avoid recompiling unchanged files between runs (even without watch mode).

- `chokidar`
- `node-watch`


necolas:
> Can be configured to also pre-compile a given list of node_modules packages

Note that this is kind of how the RN ecosystem (Expo) handles compiling RN packages for web, and it's pretty fragile because someone has to know a list of all dependencies and sub-dependencies that might be need processing. Probably a bit of a challenge to reliably do that automatically. Do you know if the other CLI-based projects have any alternatives, or does everything expect the developer to list out the packages to be processes?

nmn:
I know that Next has a `transpilePackages` option to compile packages manually. I'm following their lead on this since this.

I understand that the `node_modules` processing is going to be a challenge. I avoided doing this for a while, but there are simply too many esoteric build setups where such a system will always be needed. It unblocks StyleX usage without a huge ecosystem of bundler and framework plugins.

nmn:
Converting to draft while it's progress. Mark it as ready for review, when it's ... ready for review.

Jta26:
This is ready for an initial review.

I have no idea how to make prettier and eslint ignore my test files that are transfomed by babel. I tried adding to the `ignorePatterns` in `.eslintrc.js`, but it doesn't seem to work. I'm open to ideas.

Jta26:
> This is ready for an initial review.
> 
> I have no idea how to make eslint ignore my test files that are transfomed by babel. I tried adding to the `ignorePatterns` in `.eslintrc.js`, but it doesn't seem to work. I'm open to ideas.

I figured it out

shawngustaw:
@nmn any word on when this will land? If this is close I don't mind using a beta version and testing it out if you push a release!

packages/cli/README.md
```diff
@@ -0,0 +1,257 @@
+# @stylexjs/stylex
```
necolas, line 1: Don't forget to replace this README

packages/cli/__tests__/__mocks__/snapshot/components/button.js
```diff
@@ -0,0 +1,40 @@
...
+
+Object.defineProperty(exports, "__esModule", {
+  value: true
+});
+exports.default = Home;
+require("../stylex_bundle.css");
+var stylex = _interopRequireWildcard(require("@stylexjs/stylex"));
+var _otherStyles = _interopRequireDefault(require("./otherStyles"));
+var _npmStyles = _interopRequireDefault(require("./npmStyles"));
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
```
necolas, line 11: I wonder if for the sake of making these snapshots more readable we can avoid the babel module interop code appearing here

packages/cli/__tests__/__mocks__/snapshot/pages/home.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+Object.defineProperty(exports, "__esModule", {
```
necolas, line 3: Are these files mock snapshots? (File path is `__mocks__/snapshot`)

packages/cli/src/index.js
```diff
@@ -0,0 +1,57 @@
...
+  chalk.green(`\n
+  ______   _             _         __   __
+  / ____| | |           | |        \\ \\ / /
+ | (___   | |_   _   _  | |   ___   \\ V / 
+  \\___ \\  | __| | | | | | |  / _ \\   > <  
+  ____) | | |_  | |_| | | | |  __/  / . \\ 
+ |_____/   \\__|  \\__, | |_|  \\___| /_/ \\_\\
+                  __/ |                   
+                 |___/                
+`),
```
necolas, line 31: This doesn't looks particularly readable

packages/cli/src/options.js
```diff
@@ -0,0 +1,29 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+export default {
+  directory: {
+    alias: 'd',
+    describe: 'The directory to compile with Stylex',
```
necolas, line 13:
```suggestion
    describe: 'The directory to compile with StyleX',
```

packages/cli/src/index.js
Jta26: There's some escaped characters. I'll attach a screenshot of what it looks like in the CLI.

packages/cli/__tests__/__mocks__/snapshot/pages/home.js
Jta26: These are files I created by manually running __mocks__/source through the CLI in order to use in my tests to validate the output.

packages/cli/__tests__/__mocks__/snapshot/components/button.js
Jta26: Is this an option that can be passed into babel? These files are supposed to be the expected output of the CLI, so I think adding a specific option for tests degrades the integrity of the tests a bit.

nmn: Yes, this is an option that can be passed into Babel. We don't want any transform other than StyleX to run. So we probably just need to set `babelrc: false` to ensure that is the case.

nmn:
Looks like a few improvements still need to be made. I'll try to do this on my end shortly.

nmn:
@Jta26 I've fixed some of the issues, but it's not quite ready for primetime. Here are the blockers for the initial merge:

1. The CLI should have an option to take in a JSON configuration for StyleX.
2. The default configuration for StyleX should enable theming APIs
3. A working example app where the CLI is in use
   - I tried a standalone next app to test the CLI and the output was not compiled.

After the initial merge we can address the following issues before release:

1. Handling support for `node_modules` dependencies that use StyleX.
  - To simplify, this can be configured via an option in a config so it does not have to be detected manually.
2. Watch mode should only recompile files that have actually changed
3. Better logging - Log what files are being transformed.

TxHawks:
Re the discussion about support for handling `node_modules` and how best list them, in large monorepos that have code in multiple packages (sometimes hundreds), shared by several apps, it could be very very useful to be able to define a namespace, e.g., `@acme` or a glob-like pattern, e.g., `@acme/*`. Otherwise the burden of listing all relevant packages will often likely be too high.

packages/cli/src/transform.js
```diff
@@ -0,0 +1,98 @@
...
+        },
+      },
+    },
+  });
+
+  const result = await babel.transformFileAsync(fileName, {
+    babelrc: false,
+    plugins: [
+      [typescriptSyntaxPlugin, { isTSX: true }],
+      jsxSyntaxPlugin,
```
shawngustaw, line 53: I think this is missing `import jsxSyntaxPlugin from '@babel/plugin-syntax-jsx';` at the top of this file.

nmn:
> a glob-like pattern,

@TxHawks I don't think we will support that at first. It's important to note that you would only need to list dependencies that use StyleX within. As such, even in large mono-repos, there's usually only a handful of UI packages.

The challenge with supporting glob pattern is that we need something we can pre-compile without manually seeing what is being imported in the app.

packages/cli/src/index.js
Jta26: image

packages/cli/src/transform.js
```diff
@@ -0,0 +1,98 @@
...
+import typescriptSyntaxPlugin from '@babel/plugin-syntax-typescript';
+import {
+  copyFile,
+  getCssPathFromFilePath,
+  getInputDirectoryFiles,
+  isJSFile,
+  writeCompiledCSS,
+  writeCompiledJS,
+} from './files';
+import type { NodePath } from '../../babel-plugin/flow_modules/@babel/traverse';
```
Jta26, line 24:
```suggestion
import type { NodePath } from '@babel/traverse';
```

TxHawks:
> Even in large mono-repos, there's usually only a handful of UI packages.
> 
> The challenge with supporting glob pattern is that we need something we can pre-compile without manually seeing what is being imported in the app.

With things like Turborepo and NX, UI packages are often broken into very small parts to allow for better caching. Manually handling this could be very difficult. Could possibly be solved with a pre-build script generating the list of namespace dependencies, but that too isn't really ideal.

Am I correct in assuming you'd also need a list of these dependencies on the bundler level to produce a single CSS file and remove duplicate declarations? E.g. in case two packages have `fontWeight: 700` to not have two `.x1xlr1w8 { font-weight: 700; }` declarations in the final CSS file.

packages/cli/src/transform.js
nmn: @Jta26 It's still your PR, I just made some changes while testing it. Do you want me take over?

nmn:
> With things like Turborepo and NX, UI packages are often broken into very small parts to allow for better caching. Manually handling this could be very difficult.

Since we're pre-compiling we need to know all the code that *could* have StyleX usage. We *could* support globs but that would result in pre-compiling more code than necessary. The risk here is that this CLI would run *before* another build process, like Turbo+SWC. Compiling too much code may add to much runtime overhead.

> Am I correct in assuming you'd also need a list of these dependencies on the bundler level...

No, a bundler already walks a file tree. It can detect files that use StyleX across package boundaries.

TxHawks:
> We _could_ support globs but that would result in pre-compiling more code than necessary. The risk here is that this CLI would run _before_ another build process, like Turbo+SWC. Compiling too much code may add to much runtime overhead.

I think I understand the footgun here, but since glob would be a non-default opt-in option that can be warned against in the docs, wouldn't it be a good escape hatch to provide?

nmn:
> wouldn't it be a good escape hatch to provide?

It wouldn't be that simple. We need a full list of packages that need to be transpiled and the names they're imported as. Globs don't generically have as much detail.

In particular codebases, you may be able to guarantee that all `@acme/*` packages are in a particular location. In those scenarios, it should be easy to resolve the glob in the config file itself.

We're starting with a JSON config file, but we should be able to support `.js` files later which would make this relatively simple for you to set up on your own end.

packages/cli/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@stylexjs/cli",
+  "version": "0.5.1",
+  "description": "A cli to compile a folder with StyleX",
+  "main": "./src/index.js",
```
nmn, line 5: Making this change will fix the problem.

```suggestion
  "main": "./lib/index.js",
```

packages/cli/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@stylexjs/cli",
+  "version": "0.5.1",
+  "description": "A cli to compile a folder with StyleX",
+  "main": "./lib/index.js",
```
nmn, line 5:
```suggestion
  "main": "./lib/transform.js",
```

Jta26:
Closing to open a new one because something is messed up with github

PR #423, opened 2024-02-05
fix (ESLint Plugin): allow font weights between 1 and 1000

### LINKED ISSUE #419: @stylexjs/eslint-plugin: Support variable font weight
### Describe the issue

Stylex ESLint forces predetermined font weights, but [variable fonts support any number from 1 to 1000](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/Variable_fonts_guide#weight).

### Expected behavior

Stylex ESLint accepts any value for font weight, ex. 625.

### Steps to reproduce

1. Create a style with fontWeight property
2. Use value outside the type, for example 625 or 850.
3. Try linter

### Test case

```tsx
    example: {
        fontWeight: 625,
    },
});
```

### Additional comments

_No response_

yousoumar:
## What changed ?

This PR tells the ESLint plugin to allow  `fontWeight` to have any number between 1 and 1000 instead of some fixed ones like `100`, `200`... Fixes #419.

### DIFF
--- a/packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -493,7 +493,7 @@ revert`,
       ],
     },
     {
-      code: 'import stylex from "stylex"; stylex.create({default: {fontWeight: 99}});',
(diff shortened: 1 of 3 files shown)


nmn:
Thanks @yousoumar

PR #431, opened 2024-02-08
Docs: Document VarGroup and Theme types

### LINKED ISSUE #424: Document types for StyleX theming

Types for theming, like `VarGroup`, are not documented on the website.

nmn:

Document public types `VarGroup` and `Theme`. The writing could probably be better,
but sharing early for feedback.

## Linked PR/Issues

Fixes #424

### DIFF
--- a/apps/docs/docs/api/index.mdx
+++ b/apps/docs/docs/api/index.mdx
@@ -29,3 +29,5 @@ sidebar_position: 1
 - [`type StyleXStyles`](/docs/api/types/StyleXStyles/)
 - [`type StyleXStylesWithout`](/docs/api/types/StyleXStylesWithout/)
 - [`type StaticStyles`](/docs/api/types/StaticStyles/)
+- [`type VarGroup`](/docs/api/types/VarGroup/)
+- [`type Theme`](/docs/api/types/Theme/)
--- a/apps/docs/docs/api/types/Theme.mdx
+++ b/apps/docs/docs/api/types/Theme.mdx
@@ -9,7 +9,7 @@ sidebar_position: 5
 # `Theme<>`
 
 ```tsx
-type Theme<T extends VarGroup<unknown, symbol>, Tag extends symbol = symbol>
+type Theme<T extends VarGroup<unknown, symbol>>
 ```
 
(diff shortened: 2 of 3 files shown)


TxHawks:
I feel that the description and explanation of the second type argument to both `VarGroup` and `Theme` isn't clear enough, and that the example doesn't clearly illustrates what it would be used for, or what's a real world use case in which you'd reach out for this.

Otherwise it's all very clear and well articulated

NoriSte:
Thanks for working on it 😊 IMO
1. The doc is clear
2. The doc is concise
3. The last part should be explained better

Screenshot_20240208-053333

What I mean: the users could miss the use case and stuck on it because that want to fully understand it. If they can't, an idea could pop in their mind: "stylex is hard". And it's a lose-lose.

My Pareto-principle suggestion here is: just mention that "this technique is commonly referred as Branded Types" so people have something to Google for. And if the technique doesn't click for them, the result is that in their mind "oh, TypeScript is hard, but stylex is cool" pops 😊

nmn:
@NoriSte @TxHawks Thanks for the feedback! The challenge is that the second argument should be needed *very* rarely. Your feedback has been very helpful. I'll think about how to both make it easier while making it evident that's it's an advanced feature that shouldn't be needed most of the time.

TxHawks:
I think giving a concrete example of when it would be needed would help in both explaining exactly what it is, and that it's an advanced feature that should hardly ever be used.

Maybe something along the lines of

```md
The second type argument should be rarely needed, and is meant for advanced cases where type branding might be desirable.

For example, ... 
```

nmn:
@TxHawks @NoriSte I've updated the write up. Please see [this link](https://stylex-cmv0tj9k1-fbopensource.vercel.app/?vercelToolbarCode=p1z9mjTsWx94K4e) to see if it's both clearer and less overwhelming.

I removed the second argument from the core explaination and put it in a `<details>` element.

NoriSte:
Well done!! 😊

TxHawks:
This makes it completely clear in my opinion. There's a small typo though: 
However, each usage **of** stylex.defineVars

The of us missing

PR #436, opened 2024-02-09
Fix: stylex package build on windows

Jta26:
I was attempting to build on windows and was getting that `BABEL_ENV` wasn't found. Seems like adding `cross-env` like the other `NODE_ENV` in the repo uses fixes it.


nmn (approved):
I think `cross-env` needs to be included as a `devDependency` of this package.

PR #440, opened 2024-02-10
feat: Handle variable fallbacks with firstThatWorks

### LINKED ISSUE #432: API for CSS variable fallback
## Feature request

I would like to be able to use fallback for css variables.

## Suggestion 1

I suggest an API like:

```tsx
  slot: {
    color: stylex.fallback(vars.specificSlotColor, vars.textColor),
  },
});

<div {...stylex.props(styles.slot)}>text</div>
```

Which would be compiled to:

```css
.className {
  color: var(--specificSlotColor, --textColor);
}
```

## Suggestion 2

Today, `null` and `undefined` values for style properties are not used:

> Using null for a non-default condition has no effect and should be considered invalid.
https://stylexjs.com/docs/learn/styling-ui/defining-styles/#combining-conditions

Also, in this example:

```tsx
const vars = stylex.defineVars({
  textColor: 'black',
  specificSlotColor: undefined,
});

  text: {
    color: vars.textColor,
  },
  specificSlot: {
    color: vars.specificSlotColor,
  },
});

<div {...stylex.props(styles.text, styles.specificSlot)}>text</div>
```

The final color is `undefined` because "The last style applied always wins".

I suggest that "the last *defined* style applied always wins", so that the `textColor` fallback is used.

## Workaround

```tsx
<div {...stylex.props(styles.text)}>
  <div {...stylex.props(styles.specificSlot)}>text</div>
</div>
```

nmn:

This PR makes it possible to use `stylex.firstThatWorks` to also work
for fallback values for CSS variables.

Something like this should now work:

```tsx
margin: stylex.firstThatWorks(vars.foo, 200)
```

## Linked PR/Issues

Fixes #432 

## Additional Context

A unit test has been added for the new behaviour.

----

`stylex.firstThatWorks` is a lightweight transform that simply converts the arguments to an array reverses them. This transform is still unchanged. This is important so that we can convert numbers to the correct type of string.

But after the values are converted to `px` (or not), it now looks for variables within the array of values and compresses them to handle this case correctly.

It can handle `[...values, ...variables, ...values]`;

Non-contiguous variables are not allowed.

Therefore a complex value like this would work:

```tsx
height: stylex.firstThatWorks('100dvh', vars.foo, vars.bar, '100vh', 500)
```

Which will result in the following CSS:

```css
height:var(--foo,var(--bar, 500px));
height:var(--foo,var(--bar, 100vh));
height:100dvh;
```

---

The question here is if the first line should just be `500px` instead.

IMO, this depends on how CSS handles a CSS variable within an unknown fallback value.
If `var(--x, bla)` is considered a valid style then a change should be made.

### DIFF
--- a/packages/shared/__tests__/convert-to-className-test.js
+++ b/packages/shared/__tests__/convert-to-className-test.js
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import { convertStyleToClassName } from '../src/convert-to-className';
+
+const extractBody = (str: string) => str.slice(str.indexOf('{') + 1, -1);
+
+const convert = (styles: Parameters<typeof convertStyleToClassName>[0]) =>
+  extractBody(convertStyleToClassName(styles, [], [])[2].ltr);
+
+describe('convert-to-className test', () => {
+  test('converts style to className', () => {
+    expect(convert(['margin', 10])).toEqual('margin:10px');
+  });
+  test('converts margin number to px', () => {
+    expect(convert(['margin', 10])).toEqual('margin:10px');
+  });
+  test('keeps number for zIndex', () => {
+    expect(convert(['zIndex', 10])).toEqual('z-index:10');
+  });
+  test('keeps number for opacity', () => {
+    expect(convert(['opacity', 0.25])).toEqual('opacity:.25');
+  });
+  test('handles array of values', () => {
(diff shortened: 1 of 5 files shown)


nmn:
Follow up:

```css
height:var(--foo,var(--bar, 500px));
height:var(--foo,var(--bar, 100vh));
height:100dvh;
```

is equivalent to:

```css
height:var(--foo,var(--bar, 100vh));
height:100dvh;
```

So if there are multiple fallback values for a CSS variable, only the first value even matters.

As of right now, I don't see a better way to do this. It's possible to use `@supports` to get around this problem but that would need a bigger architectural change. So, I'm going to merge this as is and add documentation to explain how CSS variables are never considered "invalid", so only a single fallback value would work.

PR #457, opened 2024-02-18
fix: add `treeshakeCompensation` to esbuild plugin

### LINKED ISSUE #454: [@stylexjs/esbuild-plugin] The esbuld-plugin does not embed variables in the TypeScript build.
### Describe the issue

When I use variables with the `@stylexjs/esbuild-plugin`, the plugin does not embed variables defined by `stylex.defineVars()` into the generated CSS.  It works in the ESM build as expected, but it does not work well in the TypeScript build.  Here is an example:

Built from ESM
```css
:root{--xd1a5w:black;--x1h2un6e:white;}
@media (prefers-color-scheme: dark){:root{--xd1a5w:white;--x1h2un6e:black;}}
.xb26llm:not(#\#){background-color:var(--x1h2un6e)}
.x1bbqryo:not(#\#){color:var(--xd1a5w)}
```

Built from TypeScript
```css
.x1e8mcya{background-color:var(--x1w3s4ik)}
.x9fir3{color:var(--x5x584n)}
```

### Expected behavior

The `@stylexjs/esbuild-plugin` embed variabled defined by `stylex.defineVars()` in the TypeScript build.

### Steps to reproduce

I reproduce it by the following versions:
  - @stylexjs/esbuild-plugin: 0.5.1
  - @stylexjs/stylex: 0.5.1
  - esbuild: 0.20.0

1. Build typescripts by the following configuration

```javascript
// build.mjs
import esbuild from 'esbuild';
import stylexPlugin from '@stylexjs/esbuild-plugin';
import path from 'path';
import { fileURLToPath } from 'url';

const __dirname = path.dirname(fileURLToPath(import.meta.url));

esbuild.build({
  entryPoints: [`src/index.ts`],
  outdir: `./build`,
  bundle: true,
  plugins: [
    stylexPlugin({
      dev: false,
      generatedCSSFileName: path.resolve(__dirname, `build/stylex.css`),
      stylexImports: ['@stylexjs/stylex'],
      unstable_moduleResolution: {
        type: 'commonJS',
        rootDir: __dirname,
      },
    }),
  ],
})
```

```typescript
// src/index.ts
import { colors } from './tokens.stylex';

  container: {
    color: colors.foreground,
    backgroundColor: colors.background,
  },
});

window.dummy = stylex.props(styles.container);
```

```typescript
// src/tokens.stylex.ts

const DARK = '@media (prefers-color-scheme: dark)';

export const colors = stylex.defineVars({
  foreground: { default: 'black', [DARK]: 'white' },
  background: { default: 'white', [DARK]: 'black' },
});
```

2. Compile and  bundle TypeScript files by `node build.mjs`
3. Check a generated CSS in the `build` directory

The full example project is available here:

- https://stackblitz.com/edit/stylexjsesbuild-plugin

### Test case

_No response_

### Additional comments

_No response_

nedjulius:

This PR adds `treeshakeCompensation: true` to stylex babel plugin in esbuild plugin to avoid elision of modules. It also adds an esbuild plugin test case for `TypeScript` files.

## Linked PR/Issues

Fixes #454 

## Pre-flight checklist

### DIFF
--- a/packages/esbuild-plugin/src/index.js
+++ b/packages/esbuild-plugin/src/index.js
@@ -115,6 +115,7 @@ export default function stylexPlugin({
               jsxSyntaxPlugin,
               stylexBabelPlugin.withOptions({
                 ...options,
+                treeshakeCompensation: true,
                 dev,
                 unstable_moduleResolution,
               }),
--- a/packages/esbuild-plugin/src/index.js
+++ b/packages/esbuild-plugin/src/index.js
@@ -114,8 +114,8 @@ export default function stylexPlugin({
               ...getFlowOrTypeScriptBabelSyntaxPlugins(currFilePath),
               jsxSyntaxPlugin,
               stylexBabelPlugin.withOptions({
-                ...options,
                 treeshakeCompensation: true,
+                ...options,
                 dev,
                 unstable_moduleResolution,
               }),
--- a/apps/docs/docs/learn/03-installation.mdx
+++ b/apps/docs/docs/learn/03-installation.mdx
@@ -247,17 +247,6 @@ provides (experimental) plugins for Webpack, Rollup, Next.js and esbuild.
     ],
   })
   ```
-
-  If your project is using **TypeScript**, you will need to set `verbatimModuleSyntax` compiler option
-  to `true` in your [`tsconfig.json`](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html)
-
-  ```json title="tsconfig.json"
-  {
-    "compilerOptions": {
-      "verbatimModuleSyntax": true
-    }
-  }
-  ```
 </details>
 
 Please refer to the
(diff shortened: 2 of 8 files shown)


nmn:
Thanks for the quick fix to documentation, but could you instead setting `treeshakeCompensation: true` in the ESBuild plugin instead? That should fix this problem without needing an extra option for devs to configure.

nedjulius:
@nmn will add. thanks for letting me know 👍

nedjulius:
updated the PR

packages/esbuild-plugin/src/index.js
```diff
@@ -115,6 +115,7 @@ export default function stylexPlugin({
               jsxSyntaxPlugin,
               stylexBabelPlugin.withOptions({
                 ...options,
+                treeshakeCompensation: true,
```
nmn, line 118: Ideally, you should make `treeshakeCompensation` an optional argument that defaults to `true`.

nmn (changes requested):
Changes look good, but there is a failing test.

packages/esbuild-plugin/src/index.js
nedjulius: I think moving it above `...options` should suffice?

nmn: Yeah this should work.

nmn (approved):

PR #459, opened 2024-02-20
Add babel plugin options comment to esbuild example

fredrivett:
## What changed / motivation?

I spent half a day trying to work out why my stylex variables were getting stripped out unless I had a `console.log` logging them. It turns out there's an option for the [babel plugin options](https://stylexjs.com/docs/api/configuration/babel-plugin/) of `treeshakeCompensation` which solved the issue.

It was not obvious to me that the object here was being passed to the babel plugin behind the scenes, which I only discovered by digging through the source code. Hopefully this comment helps others gain a better understanding of how the `stylexPlugin` works under the hood.

## Additional Context

Documentation (comment) only change

## Pre-flight checklist

### DIFF
--- a/apps/esbuild-example/scripts/build.mjs
+++ b/apps/esbuild-example/scripts/build.mjs
@@ -30,6 +30,8 @@ esbuild
     outfile: OUTFILE,
     minify: true,
     plugins: [
+      // See all options in the babel plugin configuration docs:
+      // https://stylexjs.com/docs/api/configuration/babel-plugin/
       stylexPlugin({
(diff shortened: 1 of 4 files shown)


fredrivett:
Also just realised that there's a related PR (https://github.com/facebook/stylex/pull/457) which would have solved my specific issue (lack of visibility of treeshaking compensation option), but still think adding more clarity will likely benefit people on what options they can pass.

nmn:
The changes make sense. Thank you.

PR #469, opened 2024-02-29
Use stylex.props API in readme example

keeganstreet:

I noticed that the example code in the readme is using `stylex.apply` which doesn't seem to exist. Maybe an old version of the API?

## Pre-flight checklist


nmn:
Can't believe no one else caught this till now. Thanks!

PR #491, opened 2024-03-12
A CLI to compile an entire folder with StyleX

Jta26:

## Expected Behaviour

The CLI will then:
- [x] Transform every single JS file within the `source/` folder with the StyleX Babel plugin and output it at the same relative path in the `src/` folder.
- [x] While transforming it will collect the generated CSS for each file and output a CSS bundle at `src/stylex-bundle.css`.
- [x] The transformed JS files should be updated to include an import of the `stylex-bundle.css` file. (Configurable)

- [x] Have a `watch` mode so that only the files that are edited are recompiled, and the generated CSS file is updated accordingly.
- [x] Can be configured to also pre-compile a given list of `node_modules` packages and write the compiled files in some locations.
  - [x] Any processed JS files that import these packages, should be updated to rewrite their imports to point to the compiled location.

- [x] The basic configuration options should be possible with both CLI args and with a file config.
- [x] It should be possible to configure the StyleX Babel plugin *without* a `.babelrc` file at the project root. (Next goes into Babel mode if it detect a `.babelrc` file)
- [ ] It should be possible to run additional Babel transforms in addition to the StyleX Babel plugin

## Stretch Goals
- [x] Avoid recompiling unchanged files between runs (even without watch mode).

- `chokidar`
- `node-watch`

Recreated from https://github.com/facebook/stylex/pull/412


apps/cli-example/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "react": "^18",
+    "react-dom": "^18"
+  },
+  "devDependencies": {
+    "@stylexjs/cli": "0.5.1",
+    "@types/react": "^18",
+    "@types/react-dom": "^18",
+    "typescript": "^5"
+  } 
+
```
Jta26, line 18:
```suggestion
```

apps/cli-example/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "dependencies": {
+    "react": "^18",
+    "react-dom": "^18"
+  },
+  "devDependencies": {
+    "@stylexjs/cli": "0.5.1",
+    "@types/react": "^18",
+    "@types/react-dom": "^18",
+    "typescript": "^5"
+  }
```
Jta26, line 17:
```suggestion
  } 
}
```

nmn:
So @Jta26 Just the external package compilation remaining?

Jta26:
> So @Jta26 Just the external package compilation remaining?

Yep, working on it now and should have it soon.

Jta26:
Okay, so what I made now kinda works. There's a bug somewhere (probably in my code) that makes the hashes for styles that are from external `.stylex` files reference non-existing variables.

styles like this from open-props
```jsx
import {colors} from '@stylexjs/open-props/colors';
  h2Color: {
    color: colors.green
  }
});

...

<h2 {...stylex.props(styles.h2Green)}></h2>
```

Will be transformed into

```jsx
<h2 x3ifasef ... >
```

and the css will look like
```css
.x3ifasef:not(#\#):not(#\#):not(#\#){color:var(--x1masbj0)}
```

`--x1masbj0` should be a reference to the variable colors.green, but it doesn't reference anything and isn't a defined variable. The actual variable created in `colors.stylex.js` is something entirely different, and also exists in the generated css.

Something is not connected between them and I cannot figure out why.

@nmn

Jta26:
Okay so I figured it out and got it working

Babel is doing DFS, and the plugin I made was selecting on `ImportDeclaration`, but StyleX run it on `Program`, so it seems like it runs first regardless of the order of the plugins. And I think StyleX determines it's hashes based on the file paths, so to find the hash it would look at the import and hash `@stylexjs/open-props/lib/colors.stylex`, which isn't what gets hashed when the open-props module is copied over and compiled.

 it's hashing `{$rootDir}/${config.output}/stylex_compiled_modules/@stylexjs/open-props/lib/colors.stylex` when the plugin passes over the file, which is not equal to `${rootDir}/@stylexjs/open-props/lib/colors.stylex` when it hashes the import.

I fixed the issue by also using `Program` in my plugin and making it run before the StyleX plugin.

javascripter:
I'd like `importSources` config in the CLI so that I can use react-strict-dom with Next.js

nmn:
@javascripter The CLI let's you pass all the regular config options to the Babel Plugin through a JSON5 file.

Issue #478, opened 2024-03-04
flexBasis: '0%' is compiled to flex-basis: 0px, which is not the same


olivierpascal:
### Describe the issue

`flexBasis: '0%'` is compiled to `flex-basis: 0px`, which is not equivalent

### Expected behavior

`flexBasis: '0%'` to be compiled to `flex-basis: 0%`

### Steps to reproduce

```tsx
const stylex.create({
  container: {
    flexBasis: '0%',
  },
});
```

olivierpascal:
In the meantime, I use `flexBasis: '0.0%'`, which is compiled correctly.

nmn:
How are `0%` and `0px` different in practice? Can you try using a non-zero value and see that it works as expected?

We have a value minification process that tries to avoid having multiple variants of equivalent values. e.g. we want to avoid `0` and `0px` and `0%` to all be in the CSS bundle, so we try to normalize them to the same format when it doesn't change behaviour.

olivierpascal:
That's an excellent question. I don't know how they are different in practice. But they are, and my code works with '0%' and does not work with '0px' :x

olivierpascal:
I will try to make a reproducible example.

nmn:
Thanks for the example. While, this is a convincing argument for making a change, I did want to point something out.

If you read about `flex-basis` on MDN, you'll see that if you use a percentage value, it calculates the actual size based on the size of the containing element. In your example, the container doesn't have an explicit height and instead gets it's height from the flex child itself. If you set an explicit height, then 0% does behave the same as 0 and 0px.

So, at least in your example, `flex-basis: 0%` is having no effect at all since the percentage value cannot be resolved. 

I'm convinced I need to make a change to match the behavior of CSS, but it is important to know that whenever 0% behaves differently than 0 or 0px, it is likely not being used at all.

tounsoo:
@nmn I believe this can be closed as well 😄

46319943:
> Thanks for the example. While, this is a convincing argument for making a change, I did want to point something out.
> 
> If you read about `flex-basis` on MDN, you'll see that if you use a percentage value, it calculates the actual size based on the size of the containing element. In your example, the container doesn't have an explicit height and instead gets it's height from the flex child itself. If you set an explicit height, then 0% does behave the same as 0 and 0px.
> 
> So, at least in your example, `flex-basis: 0%` is having no effect at all since the percentage value cannot be resolved.
> 
[...]

PR #520, opened 2024-04-04
fix: 0 value handling for fraction and percentage

### LINKED ISSUE #498: gridTemplateRows 0fr is being compiled to 0px, which breaks css transition
### Describe the issue

Hey 👋

If I use gridTemplateRows with the value of `0fr`, this value gets replaced with `0px`. 

There's a [popular css trick](https://www.stefanjudis.com/snippets/how-to-animate-height-with-css-grid/), in order to smoothly open an accordeon, you may use a css transition from `gridTemplateRows: 0fr` to `1fr`. This css transition breaks if 0fr is being compiled to 0px, as px can not be transitioned to fr (i think?).

### Expected behavior

0fr shouldn't get compiled to 0px, and should be preserved. Basically, every transitionable property should be opted out from this optimization.

### Steps to reproduce

1. Create an example that uses the `gridTemplateRows` css transition trick from `0fr` to `1fr`, you may use the [mentioned article](https://www.stefanjudis.com/snippets/how-to-animate-height-with-css-grid/) as an example
2. Make sure if it works with transitioning from `repeat(1, 0fr)` to `repeat(1, 1fr)`
3. Replace `repeat(1, 0fr)` to `0fr` and `repeat(1, 1fr)` to `1fr`
4. Transition will break

### Test case

_No response_

### Additional comments

My current workaround is to use `repeat(1, 0fr)` instead of `0fr` and `repeat(1, 1fr)` insead of `1fr`

tounsoo:

Bug fix  🐛 🔧 

## Linked PR/Issues

Fixes #498 #478 

## Additional Context

Added a test in `convert-to-className-test.js`

## Pre-flight checklist

### DIFF
--- a/packages/shared/src/utils/normalizers/zero-dimensions.js
+++ b/packages/shared/src/utils/normalizers/zero-dimensions.js
@@ -13,6 +13,7 @@ import parser from 'postcss-value-parser';
 
 const angles = ['deg', 'grad', 'turn', 'rad'];
 const timings = ['ms', 's'];
+const fractions = ['fr'];
 
 /**
  * Remove units in zero values, except when required: in angles and timings,
@@ -43,6 +44,8 @@ export default function normalizeZeroDimensions(
       node.value = '0deg';
     } else if (timings.indexOf(dimension.unit) !== -1) {
       node.value = '0s';
+    } else if (fractions.indexOf(dimension.unit) !== -1) {
+        node.value = '0fr';
     } else if (!endFunction) {
       node.value = '0';
     }
--- a/packages/shared/src/utils/normalizers/zero-dimensions.js
+++ b/packages/shared/src/utils/normalizers/zero-dimensions.js
@@ -14,10 +14,11 @@ import parser from 'postcss-value-parser';
 const angles = ['deg', 'grad', 'turn', 'rad'];
 const timings = ['ms', 's'];
 const fraction = 'fr';
+const percentage = '%';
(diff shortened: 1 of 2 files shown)


packages/shared/src/utils/normalizers/zero-dimensions.js
```diff
@@ -43,6 +44,8 @@ export default function normalizeZeroDimensions(
       node.value = '0deg';
     } else if (timings.indexOf(dimension.unit) !== -1) {
       node.value = '0s';
+    } else if (dimension.unit === fraction) {
+        node.value = '0fr';
```
nmn, line 48: Would you mind updating this to account for `0%` as well?

nmn (approved):
LGTM. Will wait a day if you're willing to fix a bit more!

packages/shared/src/utils/normalizers/zero-dimensions.js
tounsoo: Done!

tounsoo:
@nmn I also found that the contribution guide could be improved as I wasn't running the `prettier` ahead of commit.

Potentially changing the fourth point to `4. Ensure it is ready for commit (`npm run precommit`)` could improve contributing experience.

If you agree, I can make that small change on this PR or make a new PR for it. LMK 👍

nmn (approved):
Awesome! Thanks.

PR #525, opened 2024-04-08
fix: mark open-props type as module for ESM compat

### LINKED ISSUE #505: [@styles/open-props]: Has any plan/progress to support `ESM`?
## Describe the feature request

Recently, I am trying to support astro in [unplugin-stylex](https://github.com/eryue0220/unplugin-stylex), but the `@stylexjs/open-props` didn't support ESM Module.  So is there any plan or progress to upgrade the package `@stylesjs/open-props` to support ESM Module? If no, I'd like to try.

nmn:
Fixes #505 

Marks the `@stylexjs/open-props` package as an ESM package.


eryue0220:
Sorry, I didn't notice this PR before. I committed a related pr, but the solution is different with this one. I think this will affect those non esm modules.

nmn:
@eryue0220 I saw your PR as well. I've been thinking about it and I don't think the extra work you did is needed. Since `open-props` needs to be compiled when used, it's not necessary to make CJS packages available at all.

eryue0220:
@nmn Sorry for the late response. I thought you are right, and I've closed my pr. Thank you for your answer.

PR #527, opened 2024-04-09
fix: usage of styles in a loop and non-existent styles

nmn:
Found and fixed a bug in the optimisation step for optimising `stylex.props()` and `stylex()` calls within a `for-of` loop.

The custom `evaluate` logic would fail to bail out and throw a null pointer exception instead.

The fix was to check for that null value and bail out instead.

-----

Also, internally there are some rare example of non-existent styles (`styles.unknown`) being used which can cause the Babel plugin to throw an error.

Added a check to bail out instead of throwing an error for such cases and added a couple of test cases to verify that the fix works for `stylex()`, `stylex.props()` and `stylex.attrs()`.

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-call-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-call-test.js
@@ -1910,5 +1910,28 @@ describe('@stylexjs/babel-plugin', () => {
         }[!!(sidebar == null && !isSidebar) << 2 | !!isSidebar << 1 | !!isContent << 0];"
       `);
     });
+
+    test('Using styles from a for loop', () => {
+      expect(
+        transform(
+          `
+          import stylex from '@stylexjs/stylex';
+          function test(colors, obj) {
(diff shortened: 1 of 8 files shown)


packages/babel-plugin/__tests__/stylex-transform-call-test.js
```diff
@@ -1910,5 +1910,28 @@ describe('@stylexjs/babel-plugin', () => {
...
     });
+
+    test('Using styles from a for loop', () => {
+      expect(
+        transform(
+          `
+          import stylex from '@stylexjs/stylex';
+          function test(colors, obj) {
+            for (const color of colors) {
+              obj[color.key] = stylex(color.style);
```
necolas, line 1921: Is there already a test for this against the `stylex.props` syntax?

nmn: I'll add it. For now I was testing the specific code that caused a breakage.

PR #540, opened 2024-04-16
fix: Allow all strings/numbers along with specific values

nmn:
The ESLint Plugin lets you define custom limits for various properties. So far this can be:
- `*` for any value
- `string` for all strings
- `number` for all numbers
- or an array of specific options which can be globs.

However, using `['string', 'number']` would not work as expected, and instead treat `'string'` and `'number'` as constants.

Sometimes, it might be useful to accept any number along with specific strings or globs. This PR allows for such a pattern and will interpret `'string'` to mean the type string, and `'number'` to mean the type number, even when used within an array.


necolas:
Can this be unit tested in a follow up?

PR #552, opened 2024-04-23
support font-stretch percentage values

antoinebrault:

possible values can be a percentage from 50% to 200% https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch#values

## Pre-flight checklist


necolas:
This PR allows any string, which is not correct. Since flow doesn't support typing this correctly,  maybe the compiler should perform the full check

nmn:
@necolas The ESLint rule and Babel plugin should catch this. Will make sure to update ESLint for now as the Babel validation is a longer term project.

PR #561, opened 2024-04-29
Feat/cli

Jta26:

## Expected Behaviour

The CLI will then:
- [x] Transform every single JS file within the `source/` folder with the StyleX Babel plugin and output it at the same relative path in the `src/` folder.
- [x] While transforming it will collect the generated CSS for each file and output a CSS bundle at `src/stylex-bundle.css`.
- [x] The transformed JS files should be updated to include an import of the `stylex-bundle.css` file. (Configurable)

- [x] Have a `watch` mode so that only the files that are edited are recompiled, and the generated CSS file is updated accordingly.
- [x] Can be configured to also pre-compile a given list of `node_modules` packages and write the compiled files in some locations.
  - [x] Any processed JS files that import these packages, should be updated to rewrite their imports to point to the compiled location.

- [x] The basic configuration options should be possible with both CLI args and with a file config.
- [x] It should be possible to configure the StyleX Babel plugin *without* a `.babelrc` file at the project root. (Next goes into Babel mode if it detect a `.babelrc` file)
- [x] It should be possible to run additional Babel transforms in addition to the StyleX Babel plugin

## Stretch Goals
- [x] Avoid recompiling unchanged files between runs (even without watch mode).

- `chokidar`
- `node-watch`

Recreated from https://github.com/facebook/stylex/pull/412


packages/cli/__tests__/compile-stylex-folder-test.js
```diff
@@ -123,16 +123,21 @@ describe('cli works with multiple inputs and outputs', () => {
...
   };
+  afterAll(() => clearTestDir(config));
   test('script compiles multiple directories', (done) => {
     const onClose = () => {
-      let snapshotDir = './snapshot';
+      let isSecondOutput = false;
       for (const dir of config.output) {
         if (dir.endsWith('src2')) {
-          snapshotDir = './snapshot2';
+          isSecondOutput = true;
```
nmn, line 156: Have you hardcoded support for up to 2 folders?

nmn:
This is looking good! One nit before we can ship it:

- [ ] Way to ignore certain folders in "node_modules"


- FUTURE:
  - [ ] Better logging! (This can wait)
    - Drop the file path prefix when logging
    - Don't log `[Map Iterator]`
  - [ ] Detect changed files since last run even outside of watch mode
    - There should also be a generated **JSON** file to track generated styles from each file
      so that the "merge" can happen without running compilation on all files again.
  - [ ] In watch mode, detect changes in *all* input directories
    - Currently it only detects changes in the "primary" directory
  - [ ] ... ?

packages/cli/__tests__/compile-stylex-folder-test.js
Jta26: This is only the mocks used for a simple test, it supports more than 2 directories.

nmn (approved):
Good enough for an experimental release!

We can make improvements later.

PR #569, opened 2024-05-11
feat(@stylexjs/eslint-plugin): Accept { from: string, as: string } syntax in validImports option for RSD

javascripter:

React Strict DOM [recently removed](https://github.com/facebook/react-strict-dom/issues/89
) their eslint plugin and I'm considering migrating to `@stylexjs/eslint-plugin`.

However, react-strict-dom exports `stylex` under `css` object, and it is it incompatible with current `validImports` option in the eslint plugin.

e.g. This doesn't work:
```typescript
module.exports = {
  plugins: ['@stylexjs'],
  rules: {
    '@stylexjs/sort-keys': [
      'error',
      {
        validImports: ['react-strict-dom']
      },
    ],
  },
};
```

I have added `{ from: string, as: string }` syntax similar to `importSources` option in the babel plugin for compatibility with RSD, for both `@stylexjs/sort-keys` and `@stylexjs/valid-styles` rules.

e.g. now the option accepts this format:
```
{
  validImports: [
    {
      from: 'react-strict-dom',
      as: 'css',
    },
  ],
},
```

## Additional Context

- Added unit tests
- Updated the api docs
- I also checked the behavior locally in `apps/nextjs-example` dir by replacing imports with react-strict-dom and configuring with the new options as in the screenshot.

Screenshot 2024-05-11 at 15 51 55

## Pre-flight checklist


packages/eslint-plugin/src/stylex-sort-keys.js
```diff
@@ -128,32 +145,47 @@ const stylexSortKeys = {
...
+            if (
+              specifier.type === 'ImportSpecifier' &&
+              specifier.imported.name === 'keyframes'
+            ) {
+              styleXKeyframesImports.add(specifier.local.name);
+            }
+          });
+        }
+
+        if (typeof foundImportSource === 'object') {
```
javascripter, line 180: If using the new `as` and `from` syntax, we will treat the value of the `as` option as the default import name. This is because the resolution mechanism is the same as `import stylex from 'xxx'`, where the property names of stylex are fixed to `create`, `props`, etc, and the same constraints applies to the `css` object as well.
```
import { css } from 'a'
```

nmn (approved):

necolas:
@nmn You approved this PR but it wasn't merged. Is there something else that needed work?

necolas:
See #1039 for this PR rebased on `main`

PR #577, opened 2024-05-21
Revert "fix: Keep units for 0 values (#532)"

nmn:
Not dropping units for various `0` values leads to a significant increase in the size of the generated CSS file.

So this PR, brings back this particular strategy to drop units from `0` values while keeping the `fr` and `%` units and taking a more measured approach. If any other unit needs to be maintained, we can work on that in the future.


necolas (approved):
LGTM

PR #584, opened 2024-05-23
RFC: Maintain variable names for keys that start with "--"

nmn:
# RFC - Variable name literals for keys that that start with "--"

The main change proposed is to opt out of automatic variable name generation when the object keys used within `stylex.defineVars` starts with `--`.

### 1. Using variables across StyleX and global `.css` files

There is often a desire to use variables defined within StyleX to be usable within a raw CSS file to set global styles. However, since the actual generated variable names created by StyleX are unknown until compile time, this process isn't developer friendly.

The only escape hatch today is to fallback to strings for using variables within StyleX and CSS. This means losing *all* the benefits provided by StyleX's theming APIs, such as type-safety and guaranteed default values.

### 2. Migrating from CSS strings to StyleX theming APIs

There existing codebases that use CSS variable names as strings today and don't have a clear migration path to adopt StyleX's theming APIs. While it is easier to move to StyleX's APIs for new variables, it is usually a non-starter to migrate existing variables. This new feature provides an escape hatch to make migration possible.

A typical migration process might look like:
1. Define all existing CSS variables as literals with a `stylex.defineVars` call with the default value set to `null`
2. Update all usages of that variable to use the object created with `stylex.defineVars`
3. Define the default values for the variables within StyleX
4. Define any Themes within StyleX
5. Delete the old patterns for defining the variables to make StyleX the exclusive way to define variables.
6. Optionally, if no uses of the variables exist outside of StyleX anymore, rename the keys to drop the `--`.

## Caveat

There is one major caveat or risk of this proposal. Unlike variables generated by StyleX itself, there is no way to guarantee that a variable name isn't accidentally duplicated. If the same variable name is used in two separate `stylex.defineVars` calls, it will result in the same variable being defined twice in the generated CSS file.

Ideally, in a follow-up, we should be able to *detect* such a conflict at compile time.

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-create-theme-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-create-theme-test.js
@@ -63,7 +63,7 @@ const createThemeWithDifferentOrder = `{
   cornerRadius: '6px',
 }`;
 
-describe('@stylexjs/babel-plugin', () => {
+describe('@stylexjs/babel-plugin stylex.createTheme', () => {
   beforeEach(() => {
     defineVarsOutput = transform(`
       import stylex from 'stylex';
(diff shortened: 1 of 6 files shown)


pnegahdar:
@nmn This is a cool addition. Thanks.

From an aesthetic/rename perspective theme['--bg'] is kinda meh. Would it be possible to add it to the syntax as such:

```typescript
export const colors = stylex.defineVars({
  primaryText: {default: 'black',  noRename: true},
})
```

That way dot accessors still work well.

nmn:
@pnegahdar That kind of feature is not possible due to the limitations of how the compiler works. The compiler needs to figure out the variable names not only where the variables are defined with (with `defineVars`) but also where they are imported and used.

StyleX, intentionally, does not rely on reading the imported files to keep the compiler fast and the results cacheable.

So, the only other option other than `theme['--bg']` was to use something like `theme.__bg`. I considered an approach like that but ultimately decided against it as the possible confusion it could cause was worse than the benefits it provided.

---

The `--` prefix shouldn't be needed in the vast number of cases. We're looking into ways to solve all the other usecases with other features.

PR #585, opened 2024-05-23
fix(cli): when modules experimental item is string should be copy

nonzzz:

Fix cli fetchModule can't not working as expected.

## Linked PR/Issues

None

## Additional Context

- I notice cli has been in the main branch. So i based on the example directory to test the cli without json5. And i meet some bug when the `modules_EXPERIMENTAL` children is string literal. It will break `fetchModule` logic. According [fs.cpsync](https://nodejs.org/api/fs.html#fscpsyncsrc-dest-options) so we should set a state for filter option. 
- Fix cli example incorrect babel option.

## Pre-flight checklist

### DIFF
--- a/apps/cli-example/.stylex.json
+++ b/apps/cli-example/.stylex.json
@@ -3,7 +3,7 @@
     "output": ["./src"],
     "cssBundleName": "stylex_bundle.css",
     "babelPresets": [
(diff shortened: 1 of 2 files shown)


Jta26 (approved):
Thanks for catching these

PR #593, opened 2024-05-28
Add focus-within

aspizu:
For some reason the valid-styles tests don't actually fail even if they are invalid.


nmn:
I see a prettier check failing. Could you fix that? The changes themselves look good.

PR #600, opened 2024-06-06
docs: fix stylex import source in code example

zetavg:

I tried to copy-paste and run this code snippet from "Thinking in StyleX", and found that the import source in the snippet seems incorrect.

Instead of `import * as stylex from 'stylex';`, it should be `import * as stylex from '@stylexjs/stylex';`.

## Linked PR/Issues

N/A

## Additional Context

Did a search for `import * as stylex from 'stylex';` in the codebase and did not find any other matches in `.mdx` files. 

All are `import * as stylex from '@stylexjs/stylex';`, including other code snippets in the same file. So I think this may be a mistake.

## Pre-flight checklist


nonzzz (commented):
LGTM

PR #612, opened 2024-06-23
fix: pass caller to babel within rollup plugin

### LINKED ISSUE #402: Babel plugin is not compatible with @babel/preset-env
### Describe the issue

When using rollup plugin, stylex.css stops being output when @babel/preset-env is added to a babel config in root directory

### Expected behavior

If the plugin is supposed to be compatible with @babel/preset-env, I would expect the stylex.css to still be output after adding @babel/preset-env

### Steps to reproduce

1. Use rollup-example as a starting point (I copied this folder out of npm workspace). 
2. Add the latest versions of rollup, @babel/cli, @babel/core, @babel/preset-env to package.json dependencies
3. npm install using npm 18.15.0
4. npm run build (works at this point and stylex.css is output)
5. add a babel-config.js (or json) that includes @babel/preset-env in the presets array
```json
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": {
          "node": "18"
        }
      }
    ]
  ]
}
```

And then styles stop working.  Seems any other plugins or presets works but the minute I add preset-env no styles are generated

### Test case

_No response_

### Additional comments

_No response_

nmn:
Fixes #402

There has been reports of the Rollup Plugin not playing well with `babel-env`. A fix was suggested in the issue above which is what this PR implements.

Still trying to reproduce the bug, but putting up this PR in the meantime for review.

### DIFF
--- a/packages/rollup-plugin/src/index.js
+++ b/packages/rollup-plugin/src/index.js
@@ -96,6 +96,13 @@ export default function stylexPlugin({
             unstable_moduleResolution,
           }),
         ],
+        caller: {
+          name: '@stylexjs/rollup-plugin',
+          supportsStaticESM: true,
+          supportsDynamicImport: true,
+          supportsTopLevelAwait: !inputCode.includes('require('),
+          supportsExportNamespaceFrom: true,
+        },
       });
       if (result == null) {
         console.warn('stylex: transformAsync returned null');


nmn:
I can't reproduce the issue, but the change doesn't cause any regression either.

PR #614, opened 2024-06-24
Add new lint rule for margin/padding shorthands

mellyeliu:
##  Context
This PR introduces a new lint rule for the handling of margin and padding shorthands in stylex. 

Legacy Property Mapping: Legacy margin and padding properties like marginStart and paddingEnd are now automatically mapped to their logical property counterparts (marginInlineStart, marginInlineEnd, etc.)

Shorthand Expansion: The handling of CSS shorthands for margin and padding (e.g., margin: 0 0 0 0) will be autodetected and fixable. Shorthands are now automatically expanded into separate properties (e.g., marginInlineStart, marginInlineEnd, etc.).

## Testing

Added tests to cover a number of different cases, including:
- Expanding shorthands with 1, 2, and 4 values for margin/padding
- Expanding shorthands for marginInline and paddingInline
- Replacing legacy formats
- Use of `calc` and `!important`


packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -0,0 +1,262 @@
...
+    `,
+    },
+  ],
+  invalid: [
+    {
+      code: `
+        import stylex from 'stylex';
+        const styles = stylex.create({
+          main: {
+            margin: '10px 12px 13px 14px',
```
necolas, line 45: In practice we can't really tell what the author intent was here, since 'margin' is a shorthand for physical directions. The safe expansion would be to marginTop, marginLeft, etc. But typically we want to use logical directions.

packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -0,0 +1,262 @@
...
+          },
+        });
+      `,
+      output: `
+        import stylex from 'stylex';
+        const styles = stylex.create({
+          main: {
+            marginInlineTop: '10px',
```
necolas, line 53: I assume this is meant to be marginBlockStart

packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -0,0 +1,262 @@
...
+            marginInlineEnd: '12px',
+            marginInlineBottom: '13px',
```
necolas, line 55: marginBlockEnd

packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -0,0 +1,262 @@
...
+            marginInlineStart: '14px',
+          },
+        });
+      `,
+      errors: [
+        {
+          message:
+            'Property shorthands like margin: 0 0 0 0 are not supported in styleX. Separate into marginStart, marginEnd [...].',
```
necolas, line 63: Message mentions legacy logical property names

packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -0,0 +1,262 @@
...
+        },
+      ],
+    },
+    {
+      code: `
+        import stylex from 'stylex';
+        const styles = stylex.create({
+          main: {
+            margin: '10em 1em',
```
necolas, line 99: This could be expanded to marginInline and marginBlock

packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -0,0 +1,262 @@
...
+            'Property shorthands like margin: 0 0 0 0 are not supported in styleX. Separate into marginStart, marginEnd [...].',
+        },
+      ],
+    },
+    {
+      code: `
+        import stylex from 'stylex';
+        const styles = stylex.create({
+          main: {
+            margin: '10em 1em !important',
```
necolas, line 126: I don't think we should handle values with !important, as we would aim to remove that from the StyleX API. We have some internal legacy use, but should probably only enable that for Meta behind a feature flag in the compiler. For XPR purposes we require Web code to get rid of any !important use anyway

packages/eslint-plugin/src/stylex-margin-padding-shorthand.js
```diff
@@ -72,143 +80,161 @@ const shorthandAliases = {
...
-        noLegacyFormats:
-          'Use {{correctProperty}} instead of legacy formats like {{incorrectProperty}} to adhere to logical property naming.',
-      },
-      fixable: 'code',
+  meta: {
+    messages: {
+      noMarginShorthands:
+        'Property shorthands like `{{property}}: 0 0 0 0` are not supported in styleX. Separate into individual properties.',
+      noLegacyFormats:
+        'Use {{correctProperty}} instead of legacy formats like {{incorrectProperty}} to adhere to logical property naming.',
```
necolas, line 88:
```suggestion
        `Property shorthands using multiple values like "${property}: 0 0 0 0" are not supported in StyleX. Separate into individual properties.',
      noLegacyFormats:
        `Use "${correctProperty}" instead of legacy formats like "${incorrectProperty}" to adhere to logical property naming.',
```

packages/eslint-plugin/src/stylex-margin-padding-shorthand.js
```diff
@@ -72,143 +80,161 @@ const shorthandAliases = {
...
+      } else if (property.key.type === 'Literal') {
+        key = property.key.value;
+      }
+
+      if (typeof key === 'string' && key in legacyNameMapping) {
         context.report({
           node: property,
-          message: `Property shorthands like ${key}: 0 0 0 0 are not supported in styleX. Separate into ${key}Start, ${key}End [...].`,
+          message:
+            'Use {{correctProperty}} instead of legacy formats like {{incorrectProperty}} to adhere to logical property naming.',
```
necolas, line 123:
```suggestion
            `Use "${correctProperty}" instead of legacy formats like "${incorrectProperty}" to adhere to logical property naming.',
```

packages/eslint-plugin/src/stylex-margin-padding-shorthand.js
```diff
@@ -72,143 +80,161 @@ const shorthandAliases = {
...
+      const newValues = shorthandAliases[key](property.value.value);
+
+      if (newValues.length === 1) {
+        // Single values do not need to be split
+        return;
+      }
+
+      context.report({
+        node: property,
+        message: `Property shorthands like ${key}: ${String(property.value.value)} are not supported in styleX. Separate into individual properties.`,
```
necolas, line 152:
```suggestion
        message: `Property shorthands using multiple values like "${key}: ${String(property.value.value)}" are not supported in StyleX. Separate into individual properties.`,
```

packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -35,6 +35,26 @@ eslintTester.run('stylex-sort-keys', rule.default, {
...
+        },
+      })
+    `,
+    },
+    {
+      code: `
+      import stylex from 'stylex';
+      const styles = stylex.create({
+        main: {
+          padding: 'calc(0.5 * 100px) !important',
```
necolas, line 53: Please can we remove `!important` from these fixtures so we don't need to modify them when `!important` support is removed.

packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -50,44 +70,17 @@ eslintTester.run('stylex-sort-keys', rule.default, {
         import stylex from 'stylex';
         const styles = stylex.create({
           main: {
-            marginInlineTop: '10px',
+            marginTop: '10px',
             marginInlineEnd: '12px',
-            marginInlineBottom: '13px',
+            marginBottom: '13px',
             marginInlineStart: '14px',
```
necolas, line 76: Auto-expanding into a mixture of logical and physical properties is probably not right. `margin: 1px 2px 3px 4px` is not treated as logical by browsers, so we should probably auto-expand to physical

necolas (commented):
Are we planning to handle warnings and auto-expansion for other shortforms? For example

* background
* borderColor
* borderRadius
* borderStyle
* borderWidth
* borderBlockColor
* borderBlockRadius
* borderBlockStyle
* borderBlockWidth
* borderInlineColor
* borderInlineRadius
* borderInlineStyle
* borderInlineWidth
* marginBlock
* marginInline
* overflow
* padding
* paddingBlock
* paddingInline
* transition

packages/eslint-plugin/__tests__/stylex-margin-padding.js
mellyeliu: these are the agreed upon expansions after syncing with @nmn, because of older browser compatibility. letting @nmn chip in with any thoughts

packages/eslint-plugin/__tests__/stylex-margin-padding.js
mellyeliu: synced with @nmn and we decided to leave this in as a "might as well have", but if we're planning to remove !important support in the future than I agree it may make sense to remove from the tests

packages/eslint-plugin/src/stylex-margin-padding-shorthand.js
mellyeliu: these are actually unused from me porting over this lint rule from www (will remove!), where `messageId` syntax to context.report is supported due to a newer version of eslint being used there, although {{}} is the expected format to create a message template

mellyeliu:
Other shorthands are out of scope for this specific change, but can be explored in a follow up!

packages/eslint-plugin/src/index.js
```diff
@@ -9,13 +9,16 @@
 
 import validStyles from './stylex-valid-styles';
 import sortKeys from './stylex-sort-keys';
+import marginPadding from './stylex-margin-padding-shorthand';
 
 const rules: {
   'valid-styles': typeof validStyles,
   'sort-keys': typeof sortKeys,
+  'margin-padding': typeof marginPadding,
```
necolas, line 17: Can we call this `valid-shorthands` so we can add the other shorthands later on

packages/eslint-plugin/__tests__/stylex-margin-padding.js
necolas: @nmn what is stylex currently doing under the hood when it receives values like this? are the left/right values being converted to logical styles?

packages/eslint-plugin/__tests__/stylex-margin-padding.js
```diff
@@ -0,0 +1,228 @@
...
+            marginInline: '10em 1em !important',
+          },
+        });
+      `,
+      output: `
+        import stylex from 'stylex';
+        const styles = stylex.create({
+          main: {
+            marginInlineStart: '10em 1em !important',
+            marginInlineEnd: '10em 1em !important',
```
necolas, line 155: This output is wrong

packages/eslint-plugin/src/stylex-margin-padding-shorthand.js
```diff
@@ -0,0 +1,240 @@
...
+
+const legacyNameMapping = {
+  marginStart: 'marginInlineStart',
+  marginEnd: 'marginInlineEnd',
+  paddingStart: 'paddingInlineStart',
+  paddingEnd: 'paddingInlineEnd',
+};
+
+const shorthandAliases = {
+  marginInline: (rawValue: number | string) => [
```
necolas, line 43: We should also process `marginBlock` and `paddingBlock`, so they get expanded if using more than 1 value.

packages/eslint-plugin/__tests__/stylex-margin-padding.js
nmn: @mellyeliu Yes, remove `!important` from the tests. We should just account for splitting up values even when `!important` is present. This is mostly a defensive thing to account for stuff in our codebase.

packages/eslint-plugin/__tests__/stylex-margin-padding.js
nmn: This is the correct expansion based on what StyleX does internally:

```
marginTop: '10px',
marginInlineEnd: '12px',
marginBottom: '13px',
marginInlineStart: '14px',
```

Legacy Haste reasons....

packages/eslint-plugin/__tests__/stylex-margin-padding.js
nmn: Expanding to `marginInline: '1em', marginBlock: '10em'` should be valid too.

Not a blocker for an initial implementation though.

packages/eslint-plugin/__tests__/stylex-margin-padding.js
nmn: The idea was add this to the implementation for legacy code within Meta only. @mellyeliu is it possible to make this configurable with a ESLint config option?

packages/eslint-plugin/src/index.js
nmn: Yes!

packages/eslint-plugin/src/stylex-margin-padding-shorthand.js
```diff
@@ -0,0 +1,240 @@
...
+
+/*:: import { Rule } from 'eslint'; */
+
+type Stack = null | {
+  upper: Stack,
+  prevNode: $ReadOnly<{ ...Property, ... }> | null,
+  prevName: string | null,
+  prevBlankLine: boolean,
+  numKeys: number,
+};
```
nmn, line 33: Should the object type be wrapped in a `$ReadOnly`? Unless you mutate these keys, please do so.

nmn (changes requested):
There are a few changes already suggested, but the following need to be fixed even for an initial implementation:
1. `marginInline` expansion with multiple values is broken
2. `marginBlock` and `paddingBlock` should be expanded as well. (Use `marginBlockStart` and `marginBlockEnd` for those.

packages/eslint-plugin/__tests__/stylex-margin-padding.js
necolas: I think for the purposes of the eslint plugin, we have to expand this correctly to physical properties instead of propagating what is essentially an internal StyleX bug to OSS. It means that internally we'll either have to patch the OSS eslint plugin to do the wrong thing, or better yet have a codemod do this transform and replicate the legacy behavior of StyleX. Then have the semantically correct eslint suggestion/auto-fix will work internally going forward too.

packages/eslint-plugin/__tests__/stylex-margin-padding.js
mellyeliu: this is the approach I've taken - I added an `allowImportant` config option defaulted to false. When true, the fixer preserves !important tags in the resultant code. When false, the fixer strips those values.

mellyeliu:
Updated to include the following:
1. Proper support for expanding `marginBlock`, `marginInline`, `paddingBlock`, `paddingInline`
2. A config option to specify whether to preserve `!important` tags in the fix

For the next iteration we can (1) expand this rule to include handling for other shorthands and (2) provide more specific handling for shorthands with two values (margin -> marginBlock. marginInline)

packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
```diff
@@ -0,0 +1,331 @@
...
+            marginInlineEnd: '12px',
+            marginBottom: '13px',
+            marginInlineStart: '14px',
+          },
+        });
+      `,
+      errors: [
+        {
+          message:
+            'Property shorthands using multiple values like margin: 10px 12px 13px 14px are not supported in StyleX. Separate into individual properties.',
```
necolas, line 136: nit about format of messages

```suggestion
            'Property shorthands using multiple values like "margin: 10px 12px 13px 14px" are not supported in StyleX. Separate into individual properties.',
```

packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
```diff
@@ -0,0 +1,331 @@
...
+            paddingBottom: '1em',
+            paddingInlineStart: '10em',
+            paddingInlineEnd: '1em',
+          },
+        });
+      `,
+      errors: [
+        {
+          message:
+            'Use paddingInlineStart instead of legacy formats like paddingStart to adhere to logical property naming.',
```
necolas, line 295: nit about format of messages
```suggestion
            'Use "paddingInlineStart" instead of legacy formats like "paddingStart" to adhere to logical property naming.',
```

packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
```diff
@@ -0,0 +1,331 @@
...
+    },
+    {
+      code: `
+      import stylex from 'stylex';
+      const styles = stylex.create({
+        main: {
+          marginTop: '10em',
+          marginInlineEnd: '5em',
+          marginBottom: '15em',
+          marginInlineStart: '25em',
```
necolas, line 106: I'm still convinced this not the right expansion for OSS. The [formal definition](https://developer.mozilla.org/en-US/docs/Web/CSS/margin#formal_definition) is very clear that `margin` and `padding` are shorthands for the respective top, bottom, left, right properties.

mellyeliu: Updated with a new config option `preferInline`, which is false by default to return the formal definition for margin shorthands and can be overridden to enable our current implementation

nmn (approved):
Awesome stuff @mellyeliu!

Approved!

PR #615, opened 2024-06-24
fix(babel-plugin): apply the right context for string static methods

### LINKED ISSUE #555: TypeError: String.prototype.concat called on null or undefined (v0.6.1)
### Describe the issue

Seeing the following error on v0.6.1. Code was previously working on `v0.6.0` and earlier versions
```sh
TypeError: String.prototype.concat called on null or undefined
    at concat (<anonymous>)
    at transformFile.next (<anonymous>)
    at run.next (<anonymous>)
    at transform.next (<anonymous>)
```

I was able to narrow it down to this new null check added in v0.6.1:

https://github.com/facebook/stylex/blob/c48bea12d3f4109a0b6bcce7c68c807972c16620/packages/babel-plugin/src/utils/evaluate-path.js#L693

In our case, we hit the following condition first (isStringLiteral is true)
https://github.com/facebook/stylex/blob/c48bea12d3f4109a0b6bcce7c68c807972c16620/packages/babel-plugin/src/utils/evaluate-path.js#L683-L691

^ which defines `func` as the string concat function but `context` remains undefined.

As a result, calling func later results in `TypeError: String.prototype.concat called on null or undefined` being thrown because context is undefined.
https://github.com/facebook/stylex/blob/c48bea12d3f4109a0b6bcce7c68c807972c16620/packages/babel-plugin/src/utils/evaluate-path.js#L725

I was able to workaround this by setting `context` to `val`. Reverting the `null` check also works but unsure which is the correct solution.

### Expected behavior

No errors when compiling valid code.

### Steps to reproduce

See issue description.

### Test case

_No response_

### Additional comments

_No response_

nonzzz:

Ensure the correctly binds the context when evaluating strings called with static method.

## Linked PR/Issues

Fixes #555 

## Additional Context

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/src/utils/evaluate-path.js
+++ b/packages/babel-plugin/src/utils/evaluate-path.js
@@ -690,6 +690,9 @@ function _evaluate(path: NodePath<>, state: State): any {
       ) {
         const val: number | string = object.node.value;
         func = (val as $FlowFixMe)[property.node.name];
+        if (typeof val === 'string') {
+          context = object.node.value;
+        }
       }
 
       if (func == null) {
--- a/packages/babel-plugin/__tests__/evaluation/stylex-evaluation-test.js
+++ b/packages/babel-plugin/__tests__/evaluation/stylex-evaluation-test.js
@@ -231,10 +231,15 @@ describe('custom path evaluation works as expected', () => {
     });
 
     test('Methods called by string should be bind', () => {
-      expect(evaluateFirstStatement('const x = "".concat("10px"," ").concat("10px");', {})).toBe(
-        '10px 10px',
+      expect(
+        evaluateFirstStatement(
+          'const x = "".concat("10px"," ").concat("10px");',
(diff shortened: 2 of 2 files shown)


packages/babel-plugin/src/utils/evaluate-path.js
```diff
@@ -690,6 +690,9 @@ function _evaluate(path: NodePath<>, state: State): any {
       ) {
         const val: number | string = object.node.value;
         func = (val as $FlowFixMe)[property.node.name];
+        if (typeof val === 'string') {
+          context = object.node.value;
+        }
```
nonzzz, line 695: The previous logic. we only get the String's method. but we don't setup the context for the function call.If we don't apply the context for the function. It'll cause panic. But 
the Numeric's method isn't a valid expression. So i think judge string is enough.

nmn:
Thanks!

PR #618, opened 2024-06-25
chore: Next.js + CLI example

nmn:
Added a Next.js example that uses the CLI for compilation.

NOTE: Because of some strange NPM issues, the Next build fails with some strange errors. The CLI is not the root cause and the previous Next example has fallen into this before.

So, for now, this PR is not intended to be merged. It's primarily a test to ensure that the CLI *does* work for a Next.js app.


nmn:
abandoning this for now in favour of showing the Postcss plugin.

PR #631, opened 2024-07-06
fix: BREAKING CHANGE: Themes set *all* variables in a VarGroup. always.

### LINKED ISSUE #601: Support of swapping underlying variables while using hierarchical variable references
## Background
In some design systems such as [Material Design](https://m3.material.io/foundations/design-tokens/how-to-read-tokens#efa90c6e-92da-4af3-8031-f0520540df25), [SAP Fiori](https://experience.sap.com/fiori-design-web/design-tokens/#token-hierarchy) and many more, there’s this concept called *Design Token Hierarchy*, where high level and more semantic tokens can reference low-level tokens.

<sup>(Image from https://stefaniefluin.medium.com/the-pyramid-design-token-structure-the-best-way-to-format-organize-and-name-your-design-tokens-ca81b9d8836d)</sup>

To my knowledge, we can define hierarchical tokens in StyleX like this - for system and reference tokens:

```ts
// tokens.stylex.ts

/** Reference tokens for color palette */
export const colorPalette = stylex.defineVars({
  blue: '#007AFF',
  indigo: '#5856D6',
  white: '#F2F2F7',
  // ...
});

/** System tokens for color */
export const systemColors = stylex.defineVars({
  primary: colorPalette.blue,
  onPrimary: colorPalette.white,
  // ...
});
```

And for component tokens:

```ts
// components/Button/tokens.stylex.ts

import { systemColors } from '../../tokens.stylex';

export const buttonTokens = stylex.defineVars({
  primaryColor: systemColors.primary,
  primaryLabelColor: systemColors.onPrimary,
  // ...
});
```

Which, for example, constructs a token hierarchy as:

```
(Component Token)              (System Token)            (Reference Token)

buttonTokens.primaryColor <─── systemColors.primary <─── colorPalette.blue <─── '#007AFF'
```

## The Problem
Using themes that override tokens (variables) that are referenced by other tokens (variables) might not work as expected. For example, if we want to apply a different color theme:

```jsx

import Button from './components/Button';
import { colorPalette, systemColors } from './tokens.stylex';

const indigoTheme = stylex.createTheme(systemColors, {
  primary: colorPalette.indigo,
});

function App() {
  return (
    <>
      <div {...stylex.props(indigoTheme)}>
        <h2>Button with Indigo Theme</h2>
        <p>
          In this div, a theme is applied which overrides <code>systemColors.primary</code> to <code>colorPalette.indigo</code>.
        </p>
        <Button label="Indigo button" />
      </div>
    </>
  );
}
```

We may expect an indigo button:

```
(Component Token)              (System Token)            (Reference Token)

buttonTokens.primaryColor <─── systemColors.primary <─╮x colorPalette.blue <─── '#007AFF'
                                                      │
                                                      ╰─ colorPalette.indigo <─ '#5856D6'
```

But it’s not the case. The button will still be blue, due to how CSS variables work:

```css
:root {
    --systemColors_primary: var(--colorPalette_blue);
    /* ... */          │
}                      ╰───────────────────────────────╮
                                                       │
:root {                                                ↓
    --buttonTokens_primaryColor: var(--systemColors_primary);
    /* ... */               │
}                           ╰─────────────────────────────╮
                                                          │
.indigoTheme {                                            │
    --systemColors_primary: var(--colorPalette_indigo);   │
    /* ... */                                             │
}                                      ╭──────────────────╯
                                       │
.button {                              ↓
    color: var(--buttonTokens_primaryColor);
    /* ... */
}
```
<sup>(In reality, StyleX will generate unique IDs for class names and variable names; here, we write them as recognizable names just for readability.)</sup>

To make this work, we may need to modify StyleX to re-declare all the variables that reference overwritten variables when declaring themes, so that:

```css
:root {
    --systemColors_primary: var(--colorPalette_blue);
    /* ... */
}

:root {
    --buttonTokens_primaryColor: var(--systemColors_primary);
    /* ... */
}

.indigoTheme {
    --systemColors_primary: var(--colorPalette_indigo);
                       ╰──────────────────────╮
    /* Re-declare */                          ↓
    --buttonTokens_primaryColor: var(--systemColors_primary);
    /* ... */                │
}                            ╰─────────╮
                                       │
.button {                              ↓
    color: var(--buttonTokens_primaryColor);
    /* ... */
}
```

## Repro
I made a repro here: https://github.com/zetavg/stylex-token-hierarchy-esbuild-example. It’s based on the `esbuild-example`.

---

I’m not sure if this is a thing that can be improved, is there already a solution, or it’s designed this way on purpose.

Instead of re-declaring CSS variables, there’s another solution in my mind which is done by adding class names, but I’m unsure about the performance and scalability concerns with either way.

nmn:
Partially addresses #601 

## Terminology:

- `VarGroup` - A group of variables created by `stylex.defineVars`. These variables have default values.
- `Theme` - A className generated that sets some or all of the variables in a `VarGroup` to alternate values.

## The Change

Changes the way `stylex.createTheme` such that when creating a `Theme` for a given `VarGroup`, any variables from the `VarGroup` that are *NOT* overridden will be set to their default value.

In practice, this means that the variables within a `VarGroup` will have *at most* one theme applied.

This also means that generating a theme to reset *all* variables within a `VarGroup` is as easy as:

```tsx
const resetTheme = stylex.createTheme(myVars, {});
```

`stylex.createTheme` can be used in any file, so this can be defined in the component file where a theme needs to be reset. Further, defining the same theme in multiple files will not result in duplicate CSS, and so it is safe to repeat this declaration wherever it is needed.

---

## Next Steps

A follow-up PR will detect CSS variables that point to other variables and add a minimal amount of CSS to handle those cases specifically.


necolas:
Having options means that code could be author using one option and then work totally differently if used in a project that sets a different option. Internally we're only going to use one of these so probably best to just add that one. The extra complexity here doesn't seem worth it to me

nmn:
I agree that the complexity isn't worth it. I didn't want to make a decision at the time as I don't agree with how `global` works as how themes should work. I also validated that other libraries like Vanilla Extract don't do theming the way `global` does.

Both `var` or `group` make sense to me however. `var` is currently the default and changing the behaviour to `group` sounds like a good idea to me.

PR #636, opened 2024-07-09
add esm resolver

hipstersmoothie:

I author esm and the babel plugin wasn't recognizing my imports.

## Additional Context

I tested this in my app and it resolves the files now.

The package I use is pretty lenient when It comes to imports. It might be able to be used instead of the other file path resolvers too.

## Pre-flight checklist

### DIFF
--- a/apps/docs/docs/learn/03-installation.mdx
+++ b/apps/docs/docs/learn/03-installation.mdx
@@ -243,7 +243,7 @@ provides (experimental) plugins for Webpack, Rollup, Next.js and esbuild.
         stylexImports: ['@stylexjs/stylex'],
         // Required for CSS variable support
         unstable_moduleResolution: {
-          // type: 'commonJS' | 'ESModules' | 'haste'
+          // type: 'commonJS' | 'haste'
           // default: 'commonJS'
           type: 'commonJS',
           // The absolute path to the root of your project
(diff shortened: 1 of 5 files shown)


packages/webpack-plugin/README.md
```diff
@@ -52,7 +52,7 @@ const config = (env, argv) => ({
       classNamePrefix: 'x',
       // Required for CSS variable support
       unstable_moduleResolution: {
-        // type: 'commonJS' | 'haste'
+        // type: 'commonJS' | 'esm' | 'haste'
```
nmn, line 55: My intention was to simply try both `require.resolve` *and* `import.resolve` instead of just `require.resolve`. That wouldn't need an extra option for configuration?

What do you think about that? Themes are essentially ESM only anyway.

nmn:
Github won't let me write a comment on unchanged lines of code, but what I was planning to do was to change these lines:

```tsx
for (const possiblePath of allAliases) {
  try {
    return require.resolve(possiblePath, {
      paths: [path.dirname(sourceFilePath)],
    });
  } catch {}
}
```

to:

```tsx
for (const possiblePath of allAliases) {
  try {
    return require.resolve(possiblePath, {
      paths: [path.dirname(sourceFilePath)],
    });
  } catch {
    try {
      const resolved = esmResolve(possiblePath, {
        allowImportingExtraExtensions: true,
      });
    
      if (resolved) {
        return resolved;
      }
    } catch {}
  }
}
```

hipstersmoothie:
I'll take a look later today

nmn:
This looks like a good fix for the ESM problem! Thanks.

PR #637, opened 2024-07-09
feat: Process font, background, and outline in shorthand lint rule

mellyeliu:
## Implementation

This PR iterates on a new lint rule for the handling of shorthands in stylex. 
- Here we are processing `background`, 'font' and `outline` properties.
- Because of the limitations of the parser in mapping exact styles, I'm leveraging https://www.npmjs.com/package/css-shorthand-expand to perform some of the parsing logic
- The output is transformed to stylex compatible syntax via regex parsing (stripping !important, camelCase)

## Context
On a broader scale, we want to add this fix within the js1 upgrade codemod. For the border styles, I'll explore leveraging the existing codemod instead of including it in this lint rule explicitly to avoid duplicate work.

This is a follow up to https://github.com/facebook/stylex/pull/614. 

## Testing

Added tests to cover the following:
- Style objects with `background`, `font`, and `outline`
- Above styles in isolation, as well as with other styles
- Tested !important flag, which we strip by default

## Pre-flight checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,9 +15,6 @@
         "packages/*",
         "apps/*"
       ],
-      "dependencies": {
-        "css-shorthand-expand": "^1.2.0"
-      },
       "devDependencies": {
         "@babel/cli": "^7.23.9",
         "@babel/core": "^7.23.9",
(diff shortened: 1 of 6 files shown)


necolas (approved):

mellyeliu:
Merging to sync to WWW

PR #638, opened 2024-07-09
feat: Expand configuration options to CLI

nmn:
Added a new `styleXConfig` option that can be passed to the CLI using the JSON5 file that is passed directly to the StyleX Babel plugin. There is one modification: The `aliases` can be relative instead of absolute paths. The CLI will auto-convert them to absolute paths.

Also, fixed a couple bugs:
1. The CLI was not handling the useCSSLayer option correctly
2. A few styles in open-props were incorrect

More improvements to the CLI are coming as well.


Jta26 (approved):
Looks good to me

Jta26:
Side question: is it possible for imported node modules to be under an alias. I imagine that will break the experimental module compilation if they can

nmn:
> Side question: is it possible for imported node modules to be under an alias. I imagine that will break the experimental module compilation if they can

Any aliases will need to be manually configured, and so the cli can resolve those as well.
The imported modules also have other issues, so it's a good thing they're marked as "experimental"!

PR #639, opened 2024-07-10
fix: handle watch mode errors and transform external module imports when options are specified

Jta26:
Currently the CLI when used in watch mode has 3 issues.

1. If you have babel transform errors (such as when you made a type error with the typescript plugin), then the CLI will kill itself and exit.
2. Because we copy (and the subsequently delete) the modules_experimental to the input directory, this inadvertently triggers changes in watchman.
3. The only files initially picked up when started in watch mode are JavaScript files (or typescript, etc), which means existing non-JavaScript files are not copied over.

This PR fixes the issues by adding a catch to the function that compiles the input directory, excluding external node modules from recompiling via watchman, and ensuring that the CLI will recompile on changes to non-js files.

### DIFF
--- a/packages/cli/src/index.js
+++ b/packages/cli/src/index.js
@@ -18,7 +18,7 @@ import { isDir } from './files';
 import { compileDirectory } from './transform';
 import options from './options';
 import errors from './errors';
-import watcher from './watcher';
+import { startWatcher } from './watcher';
 import fs from 'fs';
 import { clearInputModuleDir, copyNodeModules } from './modules';
 import type { CliConfig, ModuleType, TransformConfig } from './config';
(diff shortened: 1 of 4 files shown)


packages/cli/src/index.js
```diff
@@ -116,11 +116,17 @@ async function styleXCompile(cliArgsConfig: CliConfig) {
...
-      if (config.state.copiedNodeModules) {
+      try {
+        await compileDirectory(config);
+        if (config.state.copiedNodeModules) {
+          clearInputModuleDir(config);
+        }
+      } catch (err) {
+        fs.rmSync(config.output, { recursive: true, force: true });
         clearInputModuleDir(config);
+        throw err;
```
nmn, line 129: If you rethrow the error, won't the process still stop?

packages/cli/src/watcher.js
```diff
@@ -118,7 +109,20 @@ function subscribe(
...
+          (file) =>
+            // don't trigger recompile when the cli deletes the compiled modules folder
+            !file.exists && !file.name.startsWith('stylex_compiled_modules'),
+        )
+        .map((file) => file.name),
+    )
+      .then(() => {
+        clearInputModuleDir(config);
+      })
+      .catch((transformError) => {
```
nmn, line 123: NVM.

nmn (approved):
Can't see any obvious issues.

packages/cli/src/watcher.js
Jta26: Yeah, throwing the error only when it's not in watch mode. Otherwise just console.erroring it

Jta26:
@nmn requesting review again because I found another issue where I didn't properly transform modules that specified options to ignore files in their config

PR #650, opened 2024-07-22
feat: Allow "var(--" strings as values for all properties

nmn:

CSS properties are typed in StyleX, and while some properties allow arbitrary strings, many don't. These properties should still permit usages of string referencing CSS variables. The new `StringPrefix<>` type in Flow makes this possible.


necolas:
We need so many FlowFixMe to use this API 🥲

nmn:
NOTE: this will be merged after verifying how it affects our code internally.

nmn:
Update: The `StringPrefix` type still isn't ready. It has some strange bugs when being used for reads. So, this PR will have to wait for now.

necolas:
Closing old PRs

PR #664, opened 2024-08-15
fix: Fix improper variable handing in css-shorthand-expand by pre-processing whitespace

mellyeliu:
## Context
The `css-shorthand-expand` library (https://www.npmjs.com/package/css-shorthand-expand) does not handle variable spacing correctly. Property values like `rgb(0, 0, 0` and `var(--example-var, 0)` are incorrectly being split (see: D61342957). Font values with quotations perform fine.

See: [D61342957](https://www.internalfb.com/diff/D61342957)

## Implementation

Let's preprocess the values to strip the whitespace in between parentheses that occur within comma separated parameters:
- `rgb(0, 0, 0)` -> `rgb(0,0,0)`
- `var(--example-var, 0)` -> `var(--example-var,0)`

We'll add back whitespaces after the commas within parentheses after splitting values if they are absent. (Prettier should handle this as well, but just for safety)
- `rgb(0,0,0)` -> `rgb(0, 0, 0)`
- `var(--example-var,0)` -> `var(--example-var, 0)`

In cases where there are whitespaces remaining (such as in `calc(5px - 2px)` or `linear-gradient(to right, #ff7e5f, #feb47b)`), let's avoid providing an autofix for simplicity. We achieve this by checking for leftover spaces between parentheses after stripping. 

Ideally I would open a PR on the [repo](https://github.com/kapetan/css-shorthand-expand), but it has not been updated in a few years. The repo license is MIT, so another option would be to port over the code and maintain it in StyleX itself.

## Testing
Added in test cases to ensure that:
1) Single values like `borderColor: rgb(0, 0, 0)` and `borderWidth: var(--example-var, 0)`are ignored 
2) Multi-value shorthands like `borderColor: rgb(0, 0, 0) rgb(10, 10, 10)` are split appropriately
2) Multi-value shorthands like `calc(100% - 20px) calc(90% - 20px)` are marked as errors but exempt from autofix

### DIFF
--- a/packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
@@ -121,6 +121,17 @@ eslintTester.run('stylex-valid-shorthands', rule.default, {
       })
     `,
     },
+    {
+      code: `
+      import stylex from 'stylex';
+      const styles = stylex.create({
+        main: {
+          borderColor: 'rgb(0, 0, 0)',
+          borderWidth: 'var(--border-width, 10)',
+        },
+      })
+    `,
+    },
   ],
   invalid: [
     {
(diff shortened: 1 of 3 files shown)


necolas:
Looks like you need to squash your commits and rebase the diff onto main

mellyeliu:
@necolas I believe as long as we squash and merge when merging it'll automatically squash into one commit on top of main

necolas (approved):

PR #673, opened 2024-08-21
fix: process duplicated single shorthands

mellyeliu:
## Context
Last change hopefully 🤞 WWW sync is still blocked so may as well get this all in

Processing properties with 2-4 duplicate values like `margin: "10px 10px 10px"`

Implementation:
-  Take the set of split values and check if length is 1
- Previously we filtered out shorthands where the input returned singular shorthand values by default. Now we check if the value matches the original to account for this edge case

## Testing
Added tests for these use cases:
- margin: "10px 10px 10px"
- paddingInline: "20px 20px"
- padding: "25px 25px 25px 25px"

```
$ npx jest shorthands

Tests:       27 passed, 27 total
Snapshots:   0 total
Time:        1.866 s, estimated 2 s
Ran all test suites matching shorthands.
```

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
@@ -108,6 +108,19 @@ eslintTester.run('stylex-valid-shorthands', rule.default, {
       })
     `,
     },
+    {
+      code: `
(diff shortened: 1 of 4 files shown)


necolas (approved):
LGTM

Needs merge conflicts resolved

PR #675, opened 2024-08-22
Rollup example styles using 'property-specificity' style resolution

necolas:

Demonstrate impact on JS bundle size of `styleResolution` being `'property-specificity'` rather than the default of `'application-order'`.

## Additional Context

`'application-order'` generates hundreds of `null` values in the compiled JS output.

`'property-specificity'` doesn't do this and lends itself to more compact CSS bundles too.


nmn:
>  compressed | 563,563 | 504,676 | 0.90

So it looks like it saves 10% of the JS bytes in this case, if I'm reading the table right?

Should we add a new example so we can maintain a comparison going forward? There are some further optimisations that can be made to the `'application-order'` case as well.

necolas:
I think we should be working to get rid of the options and standardize on property-specificity

necolas:
Closing. Related PR #1064

PR #691, opened 2024-09-22
Install Lightning CSS dependency and process rollup-plugin CSS using …

Prakshal-Jain:
Using Lightning CSS to process the output of `stylexBabelPlugin.processStylexRules`, which outputs CSS rules. As a starting point, I've installed the package as dependency, and currently only being used in `rollup-plugin`.

## Additional Context
Using Lightning CSS `transform` function helps with:
- optimizing CSS for performance by minifying and reducing file sizes
- ensuring cross-browser compatibility through automatic prefixing and polyfilling.

The tests `packages/rollup-plugin/__tests__/index-test.js` is updated with the new output from `transform()` function, and successfully passes.
image

## Resources
1. Lightning CSS Documentation: https://lightningcss.dev/docs.html
2. Github: https://github.com/parcel-bundler/lightningcss

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "flow-bin": "^0.245.2",
     "hermes-eslint": "^0.23.1",
     "jest": "^30.0.0-alpha.6",
+    "lightningcss": "^1.27.0",
     "prettier": "^3.3.3",
     "prettier-plugin-hermes-parser": "^0.23.1",
     "rimraf": "^5.0.5",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "flow-bin": "^0.245.2",
     "hermes-eslint": "^0.23.1",
     "jest": "^30.0.0-alpha.6",
-    "lightningcss": "^1.27.0",
     "prettier": "^3.3.3",
     "prettier-plugin-hermes-parser": "^0.23.1",
     "rimraf": "^5.0.5",
--- a/packages/rollup-plugin/__tests__/index-test.js
+++ b/packages/rollup-plugin/__tests__/index-test.js
@@ -20,7 +20,7 @@ import { browserslistToTargets } from 'lightningcss';
 
 describe('rollup-plugin-stylex', () => {
   async function runStylex(options) {
-    let targets = browserslistToTargets(browserslist('>= 0.25%'));
+    const targets = browserslistToTargets(browserslist('>= 0.25%'));
 
     // Configure a rollup bundle
     const bundle = await rollup.rollup({
--- a/packages/rollup-plugin/__tests__/index-test.js
+++ b/packages/rollup-plugin/__tests__/index-test.js
@@ -61,66 +61,9 @@ describe('rollup-plugin-stylex', () => {
   it('extracts CSS and removes stylex.inject calls', async () => {
     const { css, js } = await runStylex({ fileName: 'stylex.css' });
 
-    expect(css).toMatchInlineSnapshot(`
-      "@layer priority1 {
-        @keyframes xgnty7z-B {
-          0% {
-            opacity: .25;
-          }
-
-          100% {
-            opacity: 1;
-          }
-        }
-      }
-
-      @layer priority2 {
-        .x1oz5o6v:hover {
-          background: red;
-        }
-      }
-
-      @layer priority3 {
-        .xeuoslp {
-          animation-name: xgnty7z-B;
-        }
-
-        .x1lliihq {
-          display: block;
-        }
-
-        .x78zum5 {
-          display: flex;
-        }
-
-        .xt0psk2 {
-          display: inline;
-        }
-
-        .x1hm9lzh {
-          margin-inline-start: 10px;
-        }
-      }
-
-      @layer priority4 {
-        .x1egiwwb {
-          height: 500px;
-        }
-
-        .xlrshdv {
-          margin-top: 99px;
-        }
-
-        .xh8yej3 {
-          width: 100%;
-        }
-
-        .x3hqpx7 {
-          width: 50%;
-        }
-      }
-      "
-    `);
+    expect(css).toMatchInlineSnapshot(
+      `"@layer priority1{@keyframes xgnty7z-B{0%{opacity:.25}to{opacity:1}}}@layer priority2{.x1oz5o6v:hover{background:red}}@layer priority3{.xeuoslp{animation-name:xgnty7z-B}.x1lliihq{display:block}.x78zum5{display:flex}.xt0psk2{display:inline}.x1hm9lzh{margin-inline-start:10px}}@layer priority4{.x1egiwwb{height:500px}.xlrshdv{margin-top:99px}.xh8yej3{width:100%}.x3hqpx7{width:50%}}"`,
+    );
 
     expect(js).toMatchInlineSnapshot(`
       "import stylex from 'stylex';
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -51,6 +51,7 @@
     "flow-bin": "^0.245.2",
     "hermes-eslint": "^0.23.1",
     "jest": "^30.0.0-alpha.6",
+    "lightningcss": "^1.27.0",
```
nmn, line 54: This should be added as dependency to the `rollup-plugin` and not the monorepo itself.

packages/rollup-plugin/__tests__/index-test.js
```diff
@@ -62,27 +62,64 @@ describe('rollup-plugin-stylex', () => {
     const { css, js } = await runStylex({ fileName: 'stylex.css' });
 
     expect(css).toMatchInlineSnapshot(`
-      "
-      @layer priority1, priority2, priority3, priority4;
-      @layer priority1{
-      @keyframes xgnty7z-B{0%{opacity:.25;}100%{opacity:1;}}
+      "@layer priority1 {
```
nmn, line 65: Look into the options for lightningcss so we format CSS in a more minimal formay.

nmn: I couldn't find such an option, so maybe it's OK to leave it expanded.

packages/rollup-plugin/src/index.js
```diff
@@ -59,11 +60,20 @@ export default function stylexPlugin({
...
         );
 
+        // Process the CSS using lightningcss
+        const { code } = transform({
+          filename: 'stylex.css',
+          code: Buffer.from(collectedCSS),
+        });
+
+        // Convert the Buffer back to a string
+        const processedCSS = code.toString();
```
nmn, line 70: This is a good start, but you must enable the option for vendor prefixing for this to be useful:

https://lightningcss.dev/transpilation.html#browser-targets

nmn (changes requested):
Good start. A few more changes needed before it can be merged.

packages/rollup-plugin/__tests__/index-test.js
Prakshal-Jain: We can use `minify: true` option for this.
For example:
```
import { transform } from 'lightningcss';

let { code, map } = transform({
  filename: 'style.css',
  code: Buffer.from('.foo { color: red }'),
  minify: true,
  sourceMap: true
});
```

Documentation: https://lightningcss.dev/docs.html

nmn: @Prakshal-Jain `minify: true` is more aggressive than we want. It will also get rid of all the line breaks and make the CSS generally unreadable. It should not be turned on by default.

However, the `rollup-plugin` should be updated accept props that are passed into `lightningcss` and it should be possible to pass in `minify: true` within `lightningcssArgs: {...}` by the developer.

packages/rollup-plugin/__tests__/index-test.js
```diff
@@ -61,29 +61,9 @@ describe('rollup-plugin-stylex', () => {
...
-      }
-      @layer priority4{
-      .x1egiwwb{height:500px}
-      .xlrshdv{margin-top:99px}
-      .xh8yej3{width:100%}
-      .x3hqpx7{width:50%}
-      }"
-    `);
+    expect(css).toMatchInlineSnapshot(
+      `"@layer priority1{@keyframes xgnty7z-B{0%{opacity:.25}to{opacity:1}}}@layer priority2{.x1oz5o6v:hover{background:red}}@layer priority3{.xeuoslp{animation-name:xgnty7z-B}.x1lliihq{display:block}.x78zum5{display:flex}.xt0psk2{display:inline}.x1hm9lzh{margin-inline-start:10px}}@layer priority4{.x1egiwwb{height:500px}.xlrshdv{margin-top:99px}.xh8yej3{width:100%}.x3hqpx7{width:50%}}"`,
```
nmn, line 65: Yeah this is worse than adding even more whitespace.

packages/rollup-plugin/src/index.js
```diff
@@ -59,11 +60,21 @@ export default function stylexPlugin({
           useCSSLayers,
         );
 
+        // Process the CSS using lightningcss
+        const { code } = transform({
+          filename: 'stylex.css',
+          code: Buffer.from(collectedCSS),
+          minify: true,
```
nmn, line 67: remove `minify: true`, but enable "syntax lowering". You can use a default value for `browsers` but the user should be able to override it.

nmn (changes requested):
Nice work getting most of the signals to be green.

packages/rollup-plugin/__tests__/index-test.js
Prakshal-Jain: Okay. Let me disable this for now. I'll add the lightningcssArgs in the next commit to add support for this argument.

packages/rollup-plugin/src/index.js
Prakshal-Jain: For the vendor list, would you recommend doing this:
```
let targets = browserslistToTargets(browserslist('>= 0.25%'));
```

packages/rollup-plugin/src/index.js
Prakshal-Jain: I think "syntax lowering" will be enabled when we add the supported browsers in the vendor list
https://lightningcss.dev/transpilation.html#syntax-lowering

packages/rollup-plugin/__tests__/index-test.js
Prakshal-Jain: Removed this in the newer commit. Back to the original format

packages/rollup-plugin/__tests__/index-test.js
Prakshal-Jain: Resolved. Back to original format.

packages/rollup-plugin/src/index.js
nmn: Yes. That's what I'm saying. You should choose a good default list for browsers to support. Perhaps `>0.5%` like the example in the docs? But then users should be able to customise it with their own list of supported browsers.

packages/rollup-plugin/src/index.js
```diff
@@ -25,6 +27,11 @@ const IS_DEV_ENV =
   process.env.NODE_ENV === 'development' ||
   process.env.BABEL_ENV === 'development';
 
+type LightningcssTransformOptions = $ReadOnly<{
```
Prakshal-Jain, line 30: @nmn We currently need to define this explicitly because Flow complains if I:
```
import type { TransformOptions } from 'lightningcss';
```

and use it like:
```
export type PluginOptions = $ReadOnly<{
  ...Partial<Options>,
  fileName?: string,
  babelConfig?: $ReadOnly<{
    plugins?: $ReadOnlyArray<PluginItem>,
    presets?: $ReadOnlyArray<PluginItem>,
  }>,
  useCSSLayers?: boolean,
  lightningcssOptions?: TransformOptions<{}>,      <--------Using the type
  ...
}>;
```


Do you have any suggestions to do this more elegantly?

nmn:
The changes are starting look good. Work on fixing the signals next. Run ESLint etc locally.

packages/rollup-plugin/src/index.js
Prakshal-Jain: Synced. Setting this to >1.0

necolas:
@Prakshal-Jain Please don't merge branches, especially if they haven't been approved yet. You'll have to fix the lint errors in a follow up

PR #693, opened 2024-09-23
fix: Update allowed styles with Flow and ESLint

nmn:
Update the StyleX Flow types to allow additional valid `scrollSnapType` values.

Update the ESLint rule to allow `containIntrinsic*` properties.

### DIFF
--- a/packages/eslint-plugin/src/stylex-valid-styles.js
+++ b/packages/eslint-plugin/src/stylex-valid-styles.js
@@ -1793,6 +1793,11 @@ const CSSProperties = {
   columnWidth: columnWidth,
   columns: columns,
   contain: contain,
+  containIntrinsicSize: makeUnionRule(isNumber, isString),
+  containIntrinsicBlockSize: makeUnionRule(isNumber, isString),
+  containIntrinsicInlineSize: makeUnionRule(isNumber, isString),
+  containIntrinsicHeightSize: makeUnionRule(isNumber, isString),
+  containIntrinsicWidthSize: makeUnionRule(isNumber, isString),
   containerType: makeUnionRule('normal', 'size', 'inline-size'),
   containerName: isString,
   content: content,
@@ -2356,8 +2361,7 @@ const stylexValidStyles = {
 
     const legacyProps: PropLimits = {
       'grid*': { limit: null, reason: legacyReason },
-      rowGap: { limit: null, reason: legacyReason },
-      columnGap: { limit: null, reason: legacyReason },
+      gap: { limit: null, reason: 'Use `rowGap` and `columnGap` instead' },
       'mask+([a-zA-Z])': { limit: null, reason: legacyReason },
       blockOverflow: { limit: null, reason: legacyReason },
       inlineOverflow: { limit: null, reason: legacyReason },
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -1793,6 +1793,11 @@ const CSSProperties = {
   columnWidth: columnWidth,
   columns: columns,
   contain: contain,
+  containIntrinsicSize: makeUnionRule(isNumber, isString),
+  containIntrinsicBlockSize: makeUnionRule(isNumber, isString),
+  containIntrinsicInlineSize: makeUnionRule(isNumber, isString),
+  containIntrinsicHeightSize: makeUnionRule(isNumber, isString),
```
necolas, line 1799:
```suggestion
  containIntrinsicHeight: makeUnionRule(isNumber, isString),
```

packages/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -1793,6 +1793,11 @@ const CSSProperties = {
...
+  containIntrinsicWidthSize: makeUnionRule(isNumber, isString),
```
necolas, line 1800:
```suggestion
  containIntrinsicWidth: makeUnionRule(isNumber, isString),
```

packages/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -2345,8 +2350,7 @@ const stylexValidStyles = {
 
     const legacyProps: PropLimits = {
       'grid*': { limit: null, reason: legacyReason },
-      rowGap: { limit: null, reason: legacyReason },
-      columnGap: { limit: null, reason: legacyReason },
+      gap: { limit: null, reason: 'Use `rowGap` and `columnGap` instead' },
```
necolas, line 2353: `gap` should be allowed

packages/stylex/src/StyleXCSSTypes.js
```diff
@@ -583,7 +583,21 @@ type rubyMerge = 'separate' | 'collapse' | 'auto';
...
+  | 'block mandatory'
+  | 'block proximity'
+  | 'block'
+  | 'both mandatory'
+  | 'both proximity'
+  | 'both'
+  | 'inline mandatory'
+  | 'inline proximity'
+  | 'inline'
+  | 'x mandatory'
```
necolas, line 597:
```suggestion
  | 'x'
  | 'x mandatory'
```

packages/stylex/src/StyleXCSSTypes.js
```diff
@@ -583,7 +583,21 @@ type rubyMerge = 'separate' | 'collapse' | 'auto';
...
+  | 'x proximity'
+  | 'y mandatory'
```
necolas, line 599:
```suggestion
  | 'y'
  | 'y mandatory'
```

necolas (changes requested):
Requires fixes and unit tests please.

packages/eslint-plugin/src/stylex-valid-styles.js
nmn: will be done in a follow-up. Need to auto-expand `gap` to `row-gap` and `column-gap` to deal with out-of-order css issues internally.

Issue #137, opened 2023-12-12
Target '.class[pseudo class] [pseudo element]', ie. '.button:hover ::before', or ':before:hover:prop' in tw


olivierpascal:
**Is your feature request related to a problem? Please describe.**

I would like to target `.class[pseudo class] [pseudo element]`, ie. `.button:hover ::before`.

**Describe a solution you'd like**

```ts
const buttonStyles = stylex.create({
  shadow: {
    boxShadow: {
      default: null,
      ':hover': {
        '::before': `0 4px 4px 0 ${colorScheme.shadow}`,
        '::after': `0 8px 12px 6px ${colorScheme.shadow}`,
      },
    },
  },
});
```

**Describe alternatives you've considered**

I have no alternative yet.

necolas:
Did you mean `.button:hover::before`?

nmn:
What you are asking for is essentially a descendent selector. We have documented that this is not supported for now.

https://stylexjs.com/docs/learn/thinking-in-stylex/#encapsulation

The current workaround for such situations is to set the value of a variable in the parent and read it in the child.

```ts
import {myVars} from './variables.stylex';

const buttonStyles = stylex.create({
  shadow: {
    [myVars.shadowStart]: {
      default: null,
      ':hover': `0 4px 4px 0 ${colorScheme.shadow}`,
    },
  },
  child: {
    boxShadow: myVars.shadowStart,
  }
});
```

olivierpascal:
That's genius. Thanks.

olivierpascal:
@nmn  Hum after all I think I don't fully get how your workaround works.

How does `myVars.shadowStart` is set in your code?
Does `[myVars.shadowStart]: { ... }` can really set a var content?

I get `eslint: Computed key cannot be resolved` when I am using your syntax.

Also, can `stylex.create` be used to set an arbitrary var like `--my-var: 'my-content';`? I get `eslint: This is not a key that is allowed by stylex` when trying to do so.

Thanks.

nmn:
> Does [myVars.shadowStart]: { ... } can really set a var content?

Yes! StyleX can use variables in both the key and var position. This approach is meant for one-offs, as opposed to `createTheme` which is meant to override a bunch of variables together.

> I get eslint: Computed key cannot be resolved when I am using your syntax.

This is bug in the ESLint plugin. I tried to fix it just before release, but it needs some more work.

> ...can stylex.create be used to set an arbitrary var like --my-var: 'my-content';?

[...]

PR #697, opened 2024-09-24
fix: Add missing CSS pseudo elements to types and ESLint

nikeee:
## What changed / motivation?
Found some missing pseudo classes while commenting on #137.

## Linked PR/Issues
Probably falls somewhere under #135.

## Additional Context
See diff.

Refs on pseudo elements:
- https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-slider-thumb
- https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-slider-runnable-track
- https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-range-thumb
- https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-range-track
- https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-range-progress
- https://www.smashingmagazine.com/2021/12/create-custom-range-input-consistent-browsers/
- https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/
- https://stackoverflow.com/questions/18389224/how-to-style-html5-range-input-to-have-different-color-before-and-after-slider

Sadly, there doesn't seem to be a standardized version without a prefix.

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -156,7 +156,7 @@ eslintTester.run('stylex-valid-styles', rule.default, {
           '100%': {
             opacity: 1,
           },
-        }); 
+        });
         const styles = create({
           main: {
             animationName: fadeIn,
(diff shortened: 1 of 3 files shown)


nmn (approved):
Changes look good. Thanks!

PR #703, opened 2024-09-30
Fix nested usage of `@starting-style`

### LINKED ISSUE #698: ESLint rule trips up on nested objects within at-rules such as `@starting-style`

I tried to use [`@starting-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style) and I think StyleX doesn't support it (yet)? Is something like this on the roadmap?

nikeee:
See #698

## Linked PR/Issues
Fixes #698

### DIFF
--- a/packages/eslint-plugin/src/stylex-valid-styles.js
+++ b/packages/eslint-plugin/src/stylex-valid-styles.js
@@ -2262,7 +2262,7 @@ const pseudoClassesAndAtRules = makeUnionRule(
   makeRegExRule(/^@media/, 'a media query'),
   makeRegExRule(/^@container/, 'a media query'),
   makeRegExRule(/^@supports/, 'a supports query'),
-  makeRegExRule(/^@starting-style/, 'a starting-style query'),
+  makeLiteralRule('@starting-style'),
 );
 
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -2262,6 +2262,7 @@ const pseudoClassesAndAtRules = makeUnionRule(
   makeRegExRule(/^@media/, 'a media query'),
   makeRegExRule(/^@container/, 'a media query'),
   makeRegExRule(/^@supports/, 'a supports query'),
+  makeRegExRule(/^@starting-style/, 'a starting-style query'),
```
nmn, line 2265:
```suggestion
  makeLiteralRule('@starting-style'),
```

Please change the `RegEx` to a `Literal` here since `@starting-style` is the exact string that must be used is not like `@media` or `@supports` queries where the query is added after the keyword.

nmn (changes requested):
One small change needed. But thanks for the quick PR.

nmn (approved):
Thanks!

PR #704, opened 2024-10-01
feat: new lint rule to enforce stylex file extension

mellyeliu:
## Context
This ESLint rule ensures consistency in file naming conventions when exporting `stylex.defineVars`. 

Specifically:
- Files that export `stylex.defineVars` must end with `.stylex.jsx` or `.stylex.tsx` 
- Files that do not export `stylex.defineVars` must not use the `.stylex.jsx` or `.stylex.tsx` extensions
- Users can configure the file extension via a themeFileExtension option (e.g., `.custom.jsx`), defaulting to `.stylex.jsx`.
## Test Cases:
Valid cases:
- Files with `stylex.defineVars` export correctly using `.stylex.jsx`, `.stylex.tsx`, or custom extensions.
- Files without `stylex.defineVars` export using non-.stylex extensions.

Invalid cases:
- Files exporting `stylex.defineVars` without the correct extension.
- Files not exporting `stylex.defineVars` but using `.stylex.jsx/.tsx` or custom extensions incorrectly.

### DIFF
--- a/packages/eslint-plugin/__tests__/stylex-enforce-extension-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-enforce-extension-test.js
@@ -16,6 +16,13 @@ const ruleTester = new RuleTester({
   parserOptions: { ecmaVersion: 2020, sourceType: 'module' },
 });
 
+const invalidFilenameWithDefineVars =
+  'Files that export StyleX variables defined with `stylex.defineVars()` must end with the `.stylex.jsx` or `.stylex.tsx` extension.';
+const invalidFilenameWithoutDefineVars =
+  'Only StyleX variables defined with `stylex.defineVars()` can be exported from a file with the `.stylex.jsx` or `.stylex.tsx` extension.';
+const invalidExportWithDefineVars =
+  'Files that export `stylex.defineVars()` must not export anything else.';
+
 ruleTester.run('stylex-enforce-extension', rule.default, {
   valid: [
     {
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/__tests__/stylex-enforce-extension-test.js
```diff
@@ -0,0 +1,118 @@
...
+        },
+      ],
+    },
+    {
+      code: 'export const somethingElse = {};',
+      filename: 'testComponent.stylex.tsx',
+      errors: [
+        {
+          message:
+            'Files that do not export `stylex.defineVars` must not have a `.stylex.jsx` or `.stylex.tsx` extension.',
```
nmn, line 91: Change the error message please:

```
Only StyleX variables defined with `stylex.defineVars()` can be exported from a file with the ".stylex.tsx" extension.
```

packages/eslint-plugin/src/stylex-enforce-extension.js
```diff
@@ -0,0 +1,91 @@
...
+
+            if (
+              init &&
+              init.callee &&
+              init.callee.object &&
+              init.callee.object.name === 'stylex' &&
+              init.callee.property &&
+              init.callee.property.name === 'defineVars'
+            ) {
+              hasDefineVarsExport = true;
```
nmn, line 65: Not quite. This just verifies that any `.stylex.js` file has at least one `stylex.defineVars`. We want to make sure there *no other exports*.

You shouldn't need the `Program:exit` at all. You should only need `ExportNamedDeclrataion` and `ExportDefaultDeclaration`.

nmn (changes requested):
Not quite what we want.

packages/eslint-plugin/src/stylex-enforce-extension.js
```diff
@@ -0,0 +1,91 @@
...
+        if (hasDefineVarsExport && !isStylexFile) {
+          context.report({
+            node,
+            message: `Files that export \`stylex.defineVars\` must have a \`${themeFileExtension}\` or \`${themeTsxExtension}\` extension.`,
+          });
+        } else if (!hasDefineVarsExport && isStylexFile) {
+          context.report({
+            node,
+            message: `Files that do not export \`stylex.defineVars\` must not have a \`${themeFileExtension}\` or \`${themeTsxExtension}\` extension.`,
+          });
```
nmn, line 84: We also want the errors to be reported on the export statements and not on the file level.

packages/eslint-plugin/src/stylex-enforce-extension.js
mellyeliu: thanks for the requirements clarification here - updated to explicitly forbid mixed imports

packages/eslint-plugin/src/stylex-enforce-extension.js
```diff
@@ -0,0 +1,96 @@
...
+              init.callee.object.name === 'stylex' &&
+              init.callee.property &&
+              init.callee.property.name === 'defineVars'
+            ) {
+              hasStyleXVarsExports = true;
+            } else {
+              hasOtherExports = true;
+            }
+          });
+        }
```
nmn, line 65: You can do something simple like this.

```suggestion
        if (node.declaration && node.declaration.declarations) {
          node.declaration.declarations.forEach((declaration) => {
            const init = declaration.init;

            if (
              init &&
              init.callee &&
              init.callee.object &&
              init.callee.object.name === 'stylex' &&
              init.callee.property &&
              init.callee.property.name === 'defineVars' &&
              !isStyleXFile
            ) {
              context.report({
                node: declaration.init,
                message: "...Must be in a .stylex.js file"
              });
            } else if (isStyleXFile) {
              context.report({
                node: declaration.init,
                message: "...Must NOT be in a .stylex.js file"
              });
            }
          });
        }
```

nmn (changes requested):
You missed Default Exports. Default exports should be treated all other exports.

mellyeliu:
Processing for default exports, general refactor to shared functions

nmn (approved):
Nice big improvement. Thanks!

PR #709, opened 2024-10-03
fix: more efficient dynamic styles

nmn:
# More efficient dynamic styles

Dynamic Styles in StyleX are expressed as minimal arrow functions that return an arrow expression. These styles are then compiled into a function that returns a tuple of two objects — an object of static classNames and an object of inline styles that *only* sets some CSS variables.

The CSS variables defined as inline styles are used in the statically generated CSS as values for the properties that are to have a dynamic value.

## What changed

In this PR, the biggest change is that the object of static classNames is slightly more dynamic than before. Instead of statically returning classNames for all properties, the classNames are now only returned conditionally *when* the dynamic value passed into the function for that particular party is *not* null or undefined.

## What does that mean?

In short this means that after this change, if a property's dynamic value resolves to `null` or `undefined`, no className will be applied for that property at all.

This makes null dynamic values more efficient.

## But also,

Previously, the inline styles would default to `initial` when the dynamic value was `null` or `undefined`. Now they default to `undefined`. This will mean that when a value is `null` or `undefined`, instead of setting a CSS variable to `initial`, we will simply omit it from the inline styles which has the same effect.

This is also a small efficiency improvement.

## ~~What's still missing~~

Not anymore!

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-create-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-create-test.js
@@ -847,12 +847,12 @@ describe('@stylexjs/babel-plugin', () => {
         _inject2(".xbsl7fq{border-style:dashed}", 2000);
         _inject2(".xn43iik{border-width:0 0 2px 0}", 2000);
         _inject2(".xmkeg23{border-width:1px}", 2000);
-        _inject2(".xa309fb{border-bottom-width:5px}", 4000);
         _inject2(".x1y0btm7{border-style:solid}", 2000);
-        _inject2(".x1q0q8m5{border-bottom-style:solid}", 4000);
         _inject2(".x1lh7sze{border-color:var(--divider)}", 2000);
-        _inject2(".xud65wk{border-bottom-color:red}", 4000);
         _inject2(".x12oqio5{border-radius:4px}", 2000);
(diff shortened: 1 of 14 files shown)


packages/babel-plugin/__tests__/stylex-transform-create-test.js
```diff
@@ -1402,15 +1406,17 @@ describe('@stylexjs/babel-plugin', () => {
...
+        _inject2(".xrkmrrc{background-color:red}", 3000);
+        _inject2(".x9lz66z{color:var(--4xs81a,revert)}", 3000);
+        _inject2("@media (min-width: 1000px){.xtljkjt.xtljkjt:hover{color:green}}", 3330);
+        _inject2(".x17z2mba:hover{color:blue}", 3130);
         export const styles = {
           default: color => [{
-            ":hover_backgroundColor": "x1gykpug",
-            ":hover_color": "x11bf1mc",
+            backgroundColor: "xrkmrrc",
+            color: "x9lz66z xtljkjt x17z2mba",
```
nmn, line 1416: This is what we actually want in this case:

```suggestion
            color: color == null ? "" : "x9lz66z " + "xtljkjt x17z2mba",
```

Issue #739, opened 2024-10-10
Prefix className with propName in development.


mellyeliu:
### Describe the feature request

Right now, we hash the <list of “conditions”>+<propertyName>+<value> to generate classNames for everything!
Instead: For common, key-value pairs, such as margin:0, generate small readable classNames, such as: m-0
We'll want to add this in as a config option to `StyleXOptions`

Some examples:
```
Margin:0 -> m-0 / `margin-0`
marginTop: 8px -> m-t-8
marginTop: 1rem -> m-t-1rem / `margin-top-1rem`
marginTop: calc(1rem - 2px); -> `m-t-ksjhc` / `margin-top-jksdhjkh`
display: block -> d-block
flex: 1` -> `f-1`
Float: left -> `fl-left`
Filter: saturate() -> `fi-hshjsfg3`

```

necolas:
I wrote a stylex patch in RSD to prefix class names with the prop name: https://github.com/facebook/react-strict-dom/blob/main/patches/%40stylexjs%2Bbabel-plugin%2B0.7.5.patch#L52

I would be wary of trying to generate tailwind-like classes, even if it's only in dev, because people will simply try to write selectors against them. We already had an issue internally where a patch was landed to target a StyleX dev classname.

It was communicated to the CSS BE team that we'd switch to dev `data-*` props (https://github.com/facebook/stylex/issues/723), which would make that a lot less likely for an engineer to try to target with a CSS selector. The GH issue frames it as providing info for a dev tools integration, but it's first an improvement over the current debugging info we generate.

nmn:
The initial goal is only to make the classes a bit more readable in development. We can start with prefixing class names with the prop name. and leave the hashes for the value part.

nmn:
See also #184

mellyeliu:
Done in https://github.com/facebook/stylex/pull/742

PR #742, opened 2024-10-10
feat: Prefix className with propName in development

mellyeliu:

Simple change to prefix dev classnames with the property name for debugging purposes. 

## Linked PR/Issues
See discussion: https://github.com/facebook/stylex/issues/739

## Screenshots, Tests, Anything Else
Added tests to ensure prefixing for dev mode and previous behaviour for production modes.

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-call-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-call-test.js
@@ -793,7 +793,7 @@ describe('@stylexjs/babel-plugin', () => {
                 padding: 5,
                 paddingEnd: 10,
               },
-  
+
               bar: {
                 padding: 2,
                 paddingStart: 10,
@@ -822,7 +822,7 @@ describe('@stylexjs/babel-plugin', () => {
                 padding: 5,
                 paddingEnd: 10,
               },
-  
+
               bar: {
                 padding: 2,
                 paddingStart: null,
@@ -996,8 +996,8 @@ describe('@stylexjs/babel-plugin', () => {
           "import _inject from "@stylexjs/stylex/lib/stylex-inject";
           var _inject2 = _inject;
           import stylex from 'stylex';
-          _inject2(".x1e2nbdu{color:red}", 3000);
-          "FooBar__styles.default x1e2nbdu";"
+          _inject2(".color-x1e2nbdu{color:red}", 3000);
+          "FooBar__styles.default color-x1e2nbdu";"
         `);
       });
 
@@ -1029,11 +1029,11 @@ describe('@stylexjs/babel-plugin', () => {
           "import _inject from "@stylexjs/stylex/lib/stylex-inject";
           var _inject2 = _inject;
           import stylex from 'stylex';
-          _inject2(".x1e2nbdu{color:red}", 3000);
-          _inject2(".x1t391ir{background-color:blue}", 3000);
+          _inject2(".color-x1e2nbdu{color:red}", 3000);
+          _inject2(".backgroundColor-x1t391ir{background-color:blue}", 3000);
           ({
-            0: "FooBar__styles.default x1e2nbdu",
-            1: "FooBar__styles.default x1e2nbdu FooBar__otherStyles.default x1t391ir"
+            0: "FooBar__styles.default color-x1e2nbdu",
+            1: "FooBar__styles.default color-x1e2nbdu FooBar__otherStyles.default backgroundColor-x1t391ir"
           })[!!isActive << 0];"
         `);
       });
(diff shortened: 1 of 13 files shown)


packages/shared/src/convert-to-className.js
```diff
@@ -58,7 +58,9 @@ export function convertStyleToClassName(
 
   // NOTE: '<>' is used to keep existing hashes stable.
   // This should be removed in a future version.
-  const className = classNamePrefix + createHash('<>' + stringToHash);
+  const className = dev
+    ? `${classNamePrefix}${dashedKey}_${createHash('<>' + stringToHash)}`
```
nmn, line 62: Let's put a separator in dev for this:

```suggestion
    ? `${classNamePrefix}-${dashedKey}_${createHash('<>' + stringToHash)}`
```

OR:

```suggestion
    ? `${dashedKey}-${classNamePrefix}${createHash('<>' + stringToHash)}`
```

nmn: There might also be some value in adding pseudo class or `_media` as prefixes, but let's save that for a follow-up.

packages/shared/src/convert-to-className.js
```diff
@@ -29,7 +29,7 @@ export function convertStyleToClassName(
   atRules: $ReadOnlyArray<string>,
   options: StyleXOptions = defaultOptions,
 ): StyleRule {
-  const { classNamePrefix = 'x' } = options;
+  const { classNamePrefix = 'x', dev = false } = options;
```
nmn, line 32: We should add a new config option for this ability.
The new option should be optional and should default to the value of `dev`.

nmn (commented):
Almost there! These changes look good. Let's add the new config option as well!

packages/shared/src/convert-to-className.js
necolas: Let's make it follow this pattern: `${key}-${classNamePrefix}${hash}`

1. We keep the prefix+hashed part the same, whether it's dev or prod.
2. The key matches the authored key name.
3. Simpler code

packages/shared/src/convert-to-className.js
necolas: I disagree. This should be the default for dev, and replace the other dev class names

necolas (changes requested):
We should also do the following:

1. Remove the existing dev class names and replace them with this approach
2. This needs to be tested e2e against the public API, not just tested against part of the internals, as we may or may not have bugs where this code is not used properly in the compiler.

packages/shared/src/convert-to-className.js
mellyeliu: I think the question on whether we want to add an override has some considerations around:
1. Do we want users to be able to disable dev classnames? ie. is there a use case where users may want to test with the same classnames as in production? I think @nmn mentioned there may be some debug cases where this may be useful, will wait for weigh in here
2. Are we opinionated in allowing/restricting this in production as well? 

In either case, having an override could be helpful. Otherwise agree with mapping to dev value.

nmn:
> 1. Remove the existing dev class names and replace them with this approach

The idea was to remove those classNames once we added the `data-*` attributes. If you feel strongly about this, we can ensure that we add those new `data attributes` before we land this PR.


> 2. This needs to be tested e2e against the public API...

Yes, we need to add more tests in the `babel-plugin` package and ensure that we run various tests with both `dev` set to `true` and `false`.

packages/shared/src/convert-to-className.js
necolas: If we want to separate "dev" and "debug" capabilities, we can change the human-readable name to be generated when a "debug" field is `true` instead. RSD makes this kind of distinction too.

necolas:
> The idea was to remove those classNames once we added the data-* attributes.

That makes sense. Please disregard my comment about that being done in this patch.

packages/shared/src/convert-to-className.js
mellyeliu: let's go with this approach then - we'll prefix the className in "debug" mode via a new config and have the dev/prod behaviour remain the same

nmn: Yes. `debug` is a far better name. Let's go with that!

packages/shared/src/convert-to-className.js
```diff
@@ -58,7 +58,9 @@ export function convertStyleToClassName(
 
   // NOTE: '<>' is used to keep existing hashes stable.
   // This should be removed in a future version.
-  const className = classNamePrefix + createHash('<>' + stringToHash);
+  const className = debug
+    ? `${dashedKey}-${classNamePrefix}${createHash('<>' + stringToHash)}`
```
necolas, line 62: I think it would be slightly clearer if we don't use the dashedKey here, as it will match the property name in the source code. We also wouldn't typically introduce dashes into the var names, which we'll want give the same treatment when debug is true.

packages/babel-plugin/__tests__/stylex-transform-call-test.js
```diff
@@ -2002,5 +2002,51 @@ describe('@stylexjs/babel-plugin', () => {
...
+        ),
+      ).toMatchInlineSnapshot(`
+        "import _inject from "@stylexjs/stylex/lib/stylex-inject";
+        var _inject2 = _inject;
+        import stylex from '@stylexjs/stylex';
+        _inject2(".margin-right-x1wsuqlk{margin-right:12px}", 4000);
+        const styles = {};
+        stylex.props(styles.unknown);"
+      `);
+    });
```
nmn, line 2050: Could you add some more tests for various different properties?

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -146,6 +146,12 @@ export default class StateManager {
   }
 
   setOptions(options: { +[string]: mixed }): StyleXStateOptions {
+    const debug: StyleXStateOptions['debug'] = z.logAndDefault(
+      z.boolean(),
+      options.debug ?? false,
```
nmn, line 151: Put this below dev and default it's value to `dev`.

```suggestion
      options.debug ?? dev,
```

packages/shared/src/convert-to-className.js
nmn:
```suggestion
    ? `${key}-${classNamePrefix}${createHash('<>' + stringToHash)}`
```

nmn (commented):
A couple of changes and type errors still.

necolas (approved):

PR #748, opened 2024-10-15
use code frames for errors for babel plugin

EvanBacon:

Using code frame errors will show the dev generally where the problem is:

## Additional Context

- I'm not familiar with the repo. Wouldn't be offended if you close the PR and replicate with preferred style.

Screenshot 2024-10-14 at 7 31 59 PM

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/src/visitors/stylex-create-theme.js
+++ b/packages/babel-plugin/src/visitors/stylex-create-theme.js
@@ -73,7 +73,7 @@ export default function transformStyleXCreateTheme(
       state,
     );
     if (!confident1) {
-      throw new Error(messages.NON_STATIC_VALUE);
+      throw callExpressionPath.buildCodeFrameError((messages.NON_STATIC_VALUE: any));
     }
 
     const injectedKeyframes: { [animationName: string]: InjectableStyle } = {};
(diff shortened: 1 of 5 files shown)


nmn:
Thank you so much @EvanBacon 

I was about to start working on this task myself, and your solution is honestly better than what I had in mind.

There is probably further improvements that can be made further down the line, but this should already be a huge improvement in the meantime.

nmn (changes requested):
If you can fix the Flow errors and failing tests then this looks good.

If you're not sure how to deal with Flow, I'm happy to do this for you!

EvanBacon:
@nmn I think you should drive this. I just took a guess at where the errors are but I'm still getting used to stylex so I'm not sure what the DX should feel like. Specifically we point to the general `path` object when things go wrong, which is already an improvement, but it's possible that we could assert on the precise dynamic variables that are triggering the errors.

Will continue to brush up on stylex and the repo though so I can be more helpful in the future.

nmn:
@EvanBacon Thank you for putting this up and getting it so far! I'll drive it from here.

nmn (approved):

PR #772, opened 2024-11-12
chore: extract eslint get souce code to util

Samantha-Zhan:

As per reviewer comment left by @mellyeliu on PR [`feat: add no-unused eslint rule to find unused styles`](https://github.com/facebook/stylex/pull/767#discussion_r1833640097), multiple eslint rule definition files are using identical conditional logic to get source code from Rule context such that we are backwards comparable with older eslint versions. This PR extract this logic into a separate util function so that it can be reused.
 
## Linked PR/Issues

Fixes #767 

## Additional Context

- `npx jest *` running all test files inside `estlint-plugin` folder, all tests still passing

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/src/stylex-sort-keys.js
+++ b/packages/eslint-plugin/src/stylex-sort-keys.js
@@ -9,7 +9,6 @@
 
 'use strict';
 
-import getSourceCode from './utils/getSourceCode';
 import type { Token } from 'eslint/eslint-ast';
 import type { RuleFixer, SourceCode } from 'eslint/eslint-rule';
 import type {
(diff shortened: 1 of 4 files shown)


mellyeliu (approved):
LGTM! fingers crossed for the day we have ESLint parity between StyleX and WWW

packages/eslint-plugin/src/stylex-sort-keys.js
```diff
@@ -9,6 +9,7 @@
 
 'use strict';
 
+import getSourceCode from './utils/getSourceCode';
```
necolas, line 12: nit: we typically import modules after importing types. see the other util imports below

necolas (approved):

nmn (approved):

PR #775, opened 2024-11-14
Interpolate size

nikeee:
## Linked PR/Issues
https://github.com/facebook/stylex/issues/135#issuecomment-2430087345

## Additional Context
Ref:
https://developer.mozilla.org/en-US/docs/Web/CSS/interpolate-size

### DIFF
--- a/packages/stylex/src/StyleXCSSTypes.js
+++ b/packages/stylex/src/StyleXCSSTypes.js
@@ -1244,7 +1244,7 @@ export type CSSProperties = $ReadOnly<{
   initialLetterAlign?: all | initialLetterAlign,
   inlineSize?: all | inlineSize,
 
-  interpolateSize: all | interpolateSize,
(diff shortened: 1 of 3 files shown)


packages/stylex/src/StyleXCSSTypes.js
```diff
@@ -1243,6 +1244,8 @@ export type CSSProperties = $ReadOnly<{
   initialLetterAlign?: all | initialLetterAlign,
   inlineSize?: all | inlineSize,
 
+  interpolateSize: all | interpolateSize,
```
nmn, line 1247: You're missing a question mark here, which is causing type errors. Please fix it so I can merge it. Thanks!

```suggestion
  interpolateSize?: all | interpolateSize,
```

nmn (changes requested):

nmn:
Please also fix the formatting issues.

Issue #779, opened 2024-11-24
[Bundlers] Investigate creating a postCSS plugin to create more reliable Bundler integrations


nmn:
### Describe the feature request

See: https://github.com/facebook/react-strict-dom/issues/34#issuecomment-2493900014

There are community efforts using a PostCSS plugin to create a Webpack plugin that works reliably in more scenarios, including Next.js.

Our current Next.js plugin is not of a high-enough quality and leads to many issues during development. A better implementation that works reliably is a big need at the moment and the results so far are promising.

@javascripter would you be interested in tackling this?

javascripter:
@nmn Yes, and I've submitted [a PR](https://github.com/facebook/stylex/pull/781).

PR #781, opened 2024-11-24
feat: add postcss-plugin

javascripter:

- Adds a PostCSS Plugin named `@stylexjs/postcss-plugin` as described in the linked issue.
- Updates nextjs-example to use` @stylexjs/postcss-plugin`

## Linked PR/Issues

Fixes https://github.com/facebook/stylex/issues/779

## Additional Context

The implementation is mostly based on [postcss-react-strict-dom](https://github.com/javascripter/postcss-react-strict-dom) that I wrote but have decided to convert it to JavaScript instead of TypeScript to make it suitable for this repository.

The plugin is rather simple so I didn't add Flow types for now (same as the current nextjs-plugin).

I've described the overview of this plugin approach [here](https://github.com/facebook/react-strict-dom/issues/236#issuecomment-2474722120).

This PR:
Screenshot 2024-11-24 at 16 14 35
Original:
Screenshot 2024-11-24 at 16 14 58

## Pre-flight checklist

### DIFF
--- a/apps/nextjs-example/app/stylex.css
+++ b/apps/nextjs-example/app/stylex.css
@@ -12,6 +12,4 @@
  * It is automatically replaced with generated CSS during builds.
  */
 
-@layer stylex {
-  @stylex;
-}
+@stylex;
--- a/apps/nextjs-example/app/stylex.css
+++ b/apps/nextjs-example/app/stylex.css
@@ -7,8 +7,11 @@
  *
  */
 
-/** @type {import('next').NextConfig} */
+/**
+ * The @stylex directive is used by the @stylexjs/postcss-plugin.
+ * It is automatically replaced with generated CSS during builds.
+ */
 
-module.exports = {
-  transpilePackages: ['@stylexjs/open-props'],
-};
+@layer stylex {
+  @stylex;
+}
--- a/apps/nextjs-example/babel.config.js
+++ b/apps/nextjs-example/babel.config.js
@@ -1,3 +1,11 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
 const path = require('path');
 module.exports = {
   presets: ['next/babel'],
@@ -8,6 +16,7 @@ module.exports = {
       // https://stylexjs.com/docs/api/configuration/babel-plugin/
       {
         dev: process.env.NODE_ENV === 'development',
+        runtimeInjection: false,
         genConditionalClasses: true,
         treeshakeCompensation: true,
         aliases: {
--- a/apps/nextjs-example/postcss.config.js
+++ b/apps/nextjs-example/postcss.config.js
@@ -7,14 +7,48 @@
  *
  */
 
+const fs = require('fs');
+const path = require('path');
+
+const projectRoot = __dirname;
+const monorepoRoot = path.join(projectRoot, '../../');
+
+function getPackageIncludePaths(packageName, nodeModulePaths) {
+  let packagePath = null;
+
+  for (const nodeModulePath of nodeModulePaths) {
+    const packageJsonPath = path.resolve(
+      nodeModulePath,
+      packageName,
+      'package.json',
+    );
+    if (fs.existsSync(packageJsonPath)) {
+      packagePath = path.dirname(packageJsonPath);
+      break;
+    }
+  }
+  if (!packagePath) {
+    throw new Error(`Could not find package ${packageName}`);
+  }
+
+  return [
+    path.join(packagePath, '**/*.{js,mjs}'),
+    '!' + path.join(packagePath, 'node_modules/**/*.{js,mjs}'),
+  ];
+}
+
+const openPropsIncludePaths = getPackageIncludePaths('@stylexjs/open-props', [
+  path.join(projectRoot, 'node_modules'),
+  path.join(monorepoRoot, 'node_modules'),
+]);
+
 module.exports = {
   plugins: {
     '@stylexjs/postcss-plugin': {
       include: [
         'app/**/*.{js,jsx,ts,tsx}',
         'components/**/*.{js,jsx,ts,tsx}',
-        // TODO: Figure out a better way to write this
-        './node_modules/@stylexjs/open-props/lib/**/*.{js,mjs}',
+        ...openPropsIncludePaths,
       ],
       useCSSLayers: true,
     },
--- a/apps/nextjs-example/postcss.config.js
+++ b/apps/nextjs-example/postcss.config.js
@@ -14,7 +14,7 @@ module.exports = {
         'app/**/*.{js,jsx,ts,tsx}',
         'components/**/*.{js,jsx,ts,tsx}',
         // TODO: Figure out a better way to write this
-        './node_modules/@stylexjs/open-props/lib/**/*.{js,jsx,ts,tsx}',
+        './node_modules/@stylexjs/open-props/lib/**/*.{js,mjs}',
       ],
       useCSSLayers: true,
     },
--- a/apps/nextjs-example/postcss.config.js
+++ b/apps/nextjs-example/postcss.config.js
@@ -1,0 +1,23 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
+
+module.exports = {
+  plugins: {
+    '@stylexjs/postcss-plugin': {
+      include: [
+        'app/**/*.{js,jsx,ts,tsx}',
+        'components/**/*.{js,jsx,ts,tsx}',
+        // TODO: Figure out a better way to write this
+        './node_modules/@stylexjs/open-props/lib/**/*.{js,jsx,ts,tsx}',
+      ],
+      useCSSLayers: true,
+    },
+    autoprefixer: {},
+  },
+};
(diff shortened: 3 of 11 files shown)


apps/nextjs-example/app/stylex.css
```diff
@@ -0,0 +1,17 @@
...
+ *
+ *
+ */
+
+/**
+ * The @stylex directive is used by the @stylexjs/postcss-plugin.
+ * It is automatically replaced with generated CSS during builds.
+ */
+
+@layer stylex {
```
javascripter, line 15: This diverges from the README from `@stylexjs/postcss-plugin` because the reset uses CSS layers so I need to apply layers here as well.

I can think of two options:
a. Leave it as is and maybe add some explanation here
b Remove using CSS Layers entirely to align it with README.md

apps/nextjs-example/babel.config.js
```diff
@@ -8,6 +16,7 @@ module.exports = {
       // https://stylexjs.com/docs/api/configuration/babel-plugin/
       {
         dev: process.env.NODE_ENV === 'development',
+        runtimeInjection: false,
```
javascripter, line 19: This is now always false because even in dev we use postcss to extract CSS statically

apps/nextjs-example/postcss.config.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+module.exports = {
+  plugins: {
+    '@stylexjs/postcss-plugin': {
+      include: [
+        'app/**/*.{js,jsx,ts,tsx}',
+        'components/**/*.{js,jsx,ts,tsx}',
+        // TODO: Figure out a better way to write this
+        './node_modules/@stylexjs/open-props/lib/**/*.{js,jsx,ts,tsx}',
```
javascripter, line 17: I tried to use something like this initially
```
require.resolve('@stylexjs/open-props')
```
as suggested [here](https://github.com/facebook/react-strict-dom/pull/238#discussion_r1842825089) but that didn't work because of individual package.json exports. For now I manually specified the files to transform and the example is rendering correctly.

apps/nextjs-example/postcss.config.js
```diff
@@ -0,0 +1,23 @@
...
+      ],
+      useCSSLayers: true,
```
javascripter, line 19: Should we use useCSSLayers true or false?

packages/postcss-plugin/src/bundler.js
```diff
@@ -0,0 +1,73 @@
...
+const babel = require('@babel/core');
+const stylexBabelPlugin = require('@stylexjs/babel-plugin');
+
+// Creates a stateful bundler for processing StyleX rules using Babel.
+module.exports = function createBundler() {
+  const styleXRulesMap = new Map();
+
+  // Determines if the source code should be transformed based on the presence of StyleX imports.
+  function shouldTransform(sourceCode) {
+    return sourceCode.includes('stylex');
```
javascripter, line 19: I did not include RSD related code in this plugin since react-strict-dom seems to have moved to its own babel/preset and we can use postcss-react-strict-dom for it already.

apps/nextjs-example/postcss.config.js
necolas: We shouldn't ever have to process non `.js` or `.mjs` files from `node_modules` packages, as everything gets compiled to js before publishing.

A more reliable way to get the package exports for any package (inc if they change their exports) is to use a script. Something like this:

```
const fs = require('fs');
const path = require('path');

function getPackageExports(packageName) {
  const packageJsonPath = require.resolve(`${packageName}/package.json`);
  const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));

  if (packageJson.exports) {
    const exports = {};
    for (const key in packageJson.exports) {
      if (typeof packageJson.exports[key] === 'string') {
        exports[key] = require.resolve(path.join(packageName, packageJson.exports[key]));
      } else if (typeof packageJson.exports[key] === 'object') {
        exports[key] = getPackageExports(path.join(packageName, key));
      }
    }
    return exports;
  } else {
    return require.resolve(packageName);
  }
}

const openPropsExports = getPackageExports('@stylexjs/open-props');
```

javascripter: Currently,
```javascript
require.resolve('@stylexjs/open-props/package.json')
```
throws this error:
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /Users/tsubasa/Projects/github.com/javascripter/stylex/apps/nextjs-example/node_modules/@stylexjs/open-props/package.json
```

[This](https://gist.github.com/javascripter/2d5992e31fa4ed6c08a77a30d3aa3db9) is the postcss.config.js I tried and I found some comments about this:
https://github.com/react-native-community/cli/issues/1168#issuecomment-1192741058
>  Please don't open issues/PRs on open source projects about this. It is not expected that packages export their package.json file. This is a problem React Native needs to fix, not every single package out there.

Is there's a reliable way to find the package root path?

javascripter: I've updated open-props include paths so that it no longer hard-code internal paths (e.g. /lib) in https://github.com/facebook/stylex/pull/781/commits/378c80e5811912511f94a1ad50b85419eb78f687
Not as accurate as I'd like as it doesn't look for exports but it now accounts for node_modules hoisting and should be more stable and explicit about what it does.

necolas: https://nodejs.org/api/packages.html#package-entry-points

> Existing packages introducing the ["exports"](https://nodejs.org/api/packages.html#exports) field will prevent consumers of the package from using any entry points that are not defined, including the [package.json](https://nodejs.org/api/packages.html#nodejs-packagejson-field-definitions) (e.g. require('your-package/package.json')). This will likely be a breaking change. To make the introduction of ["exports"](https://nodejs.org/api/packages.html#exports) non-breaking, ensure that every previously supported entry point is exported. It is best to explicitly specify entry points so that the package's public API is well-defined.

It's strange that package.json isn't reachable by default, seems like a Node.js issue tbh. But we can export package.json from packages we control

apps/nextjs-example/postcss.config.js
```diff
@@ -0,0 +1,57 @@
...
+]);
+
+module.exports = {
+  plugins: {
+    '@stylexjs/postcss-plugin': {
+      include: [
+        'app/**/*.{js,jsx,ts,tsx}',
+        'components/**/*.{js,jsx,ts,tsx}',
+        ...openPropsIncludePaths,
+      ],
```
nmn, line 52: Just so I can verify how the postcss plugin works:

We're fining all files that match these globs and compiling them as a list.

Since we're reading all the files manually anyway, if we wanted to we could accept a list of "entrypoints" instead and resolve the dependency graph, right?

packages/postcss-plugin/src/bundler.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  // Transforms the source code using Babel, extracting StyleX rules and storing them.
+  async function transform(id, sourceCode, babelConfig, options) {
```
nmn, line 23: I think we could share some code between the postcss plugin and the CLI. Although this plugin should make the CLI redundant in most cases.

nmn (approved):
This looks really good and much simpler than I thought it would be.

Thank you so much!

apps/nextjs-example/postcss.config.js
javascripter: > Since we're reading all the files manually anyway, if we wanted to we could accept a list of "entrypoints" instead and resolve the dependency graph, right?

Yes. The postcss plugin globs and compile all matching files to extract css.
This means we may be over-compiling in some cases similar to how Tailwind CLI does it with the `content` glob.

We can accept entry points and resolve dependency graphs  ourselves but we need to take into account aliases, platform extensions (.web.ts), etc. to replicate bundler dependency graphs so I avoided that to allow greater flexibility to users and to be bundler-agnostic in general.

apps/nextjs-example/app/stylex.css
javascripter: Indeed, it was redundant and I tested before adding useCSSLayers option.
Fixed in https://github.com/facebook/stylex/pull/781/commits/e2ab3e3530c2e33250119e2e9b6347bed0d1e7e4

apps/nextjs-example/postcss.config.js
javascripter: I see. We keep it to true and removed redundant `@layer stylex` wrapper

packages/postcss-plugin/src/bundler.js
javascripter: Yes. The goal of this PR is to make the plugin simple, reliable and production-ready for Next.js and other bundler users so I made the plugin simple and self-contained.

It's out of the scope of this PR because of that but I have some success in making this work for Next.js + Turbopack setup as well ([gist](https://gist.github.com/javascripter/17f0fe9cc54cca0087c273aecc4412a6)). It depends on some internals and am not using it in my production setup but I see a path where StyleX CLI becomes unnecessary for Turbopack users too so it's worth decoupling this plugin from the CLI.

nmn: I just mean we can refactor the CLI to share some of the code between it and the postcss plugin. Anyway, definitely out of scope for this PR.

nmn:
Thank you so much @javascripter! This will massively improve the experience of using StyleX.

packages/postcss-plugin/src/bundler.js
necolas: I think @javascripter's point is that it's simpler keeping the packages decoupled. There's no real value to sharing code here

PR #782, opened 2024-11-25
fix/stylex-content-quotes

### LINKED ISSUE #771: Don't wrap `content: counters(...)`  in quotes.
StyleX wraps values passed to `content` in quotes for you. However, this is sometimes problematic because it is also legal to use the `counters()` function instead of a string literal.

The compiler should be updated to detect this function, (and any other such valid value) and not wrap it in quotes. It would better to err on the side not putting quotes than putting quotes too often because it will always be possible to put the quotes in manually.

## Where to change the code

The relevant for this is the `transformValue` function in `traansform-value.js` inside of `@stylexjs/shared`

p0nch000:
# Fix StyleX content property quotes handling

## What changed / motivation
Updated `transformValue` function to correctly handle CSS `content` property values:
- Prevents wrapping CSS functions (counter, attr, url) in quotes
- Preserves CSS keywords (normal, none, open-quote)
- Supports mixed content values
- Only adds quotes for plain text strings

Previously, StyleX incorrectly wrapped CSS functions like `counters()` in quotes, breaking their functionality.

## Linked PR/Issues
Fixes #771 - Don't wrap `content: counters(...)` in quotes

### Test Coverage
```javascript
describe('transformValue content property tests', () => {
  test('preserves CSS functions without quotes', () => {
    const functions = [
      ['counters(div, ".")', 'counters(div, ".")'],
      ['counter(chapter)', 'counter(chapter)'],
      ['attr(href)', 'attr(href)'],
      ['url(image.jpg)', 'url(image.jpg)']
    ];
    functions.forEach(([input, expected]) => {
      expect(transformValue('content', input, {})).toBe(expected);
    });
  });

  test('preserves CSS keywords without quotes', () => {
    const keywords = ['normal', 'none', 'open-quote'];
    keywords.forEach(keyword => {
      expect(transformValue('content', keyword, {})).toBe(keyword);
    });
  });

  test('handles mixed content values', () => {
    expect(transformValue('content', 'open-quote counter(chapter)', {}))
      .toBe('open-quote counter(chapter)');
  });
});
```

### Implementation Notes
- Pattern matches CSS functions and keywords
- Errs on side of not adding quotes when uncertain
- Returns unchanged value if parsing unclear
- No breaking changes - maintains backward compatibility

## Pre-flight checklist

### DIFF
--- a/packages/shared/src/transform-value.js
+++ b/packages/shared/src/transform-value.js
@@ -60,7 +60,8 @@ export default function transformValue(
     );
     const isKeyword = cssContentKeywords.has(val);
     const hasMatchingQuotes =
-      val.match(/"/g)?.length >= 2 || val.match(/'/g)?.length >= 2;
+      (val.match(/"/g)?.length ?? 0) >= 2 ||
+      (val.match(/'/g)?.length ?? 0) >= 2;
 
     if (isCssFunction || isKeyword || hasMatchingQuotes) {
       return val;


p0nch000:
@nmn this was my first approach for the content issue, lmk if need to do some fixes :)

packages/shared/__tests__/stylex-transform-value.js
```diff
@@ -0,0 +1,93 @@
...
+      'unset',
+    ];
+
+    keywords.forEach((keyword) => {
+      expect(transformValue('content', keyword, {})).toBe(keyword);
+    });
+  });
+
+  test('handles mixed content values', () => {
+    const mixedValues = [
```
nmn, line 55: Same here. Values are left unchanged, so you can use an array of strings rather than an array of pairs.

packages/shared/src/transform-value.js
```diff
@@ -24,26 +24,67 @@ export default function transformValue(
...
+      'no-open-quote',
+      'no-close-quote',
+      'inherit',
+      'initial',
+      'revert',
+      'revert-layer',
+      'unset',
+    ]);
+
+    const hasCssFunction = cssContentFunctions.some(
```
nmn, line 58:
```suggestion
    const isCssFunction = cssContentFunctions.some(
```

packages/shared/src/transform-value.js
```diff
@@ -24,26 +24,67 @@ export default function transformValue(
...
+      (func) => val.startsWith(`${func}(`) && val.endsWith(')'),
+    );
+
+    const isKeyword = cssContentKeywords.has(val);
+
+    const containsMixedContent = val.split(' ').some((part) => {
```
nmn, line 64: Sadly, this won't work. I double checked to be sure and the values for `content` don't have to be separated by spaces.

There is also the concern that functions and string can contain spaces within.

nmn:
@p0nch000 Thanks for the great work. Turns out values for `content` don't need to be separated by spaces, so we might need to adjust the code a bit.

p0nch000:
Heyy took your advice i think is better now, lmk if it needs extra work

nmn (approved):
This looks great. Thanks!

p0nch000:
thanks for the fast review and the great feedback again! I'll look for my next issue to work on @nmn

PR #784, opened 2024-11-26
chore: Clean up Nextjs-example with postcss-plugin

nmn:
Chore: Just makes some small tweaks to the Next.js example to simplify the setup now that we have the postcss-plugin.


apps/nextjs-example/app/stylex.css
```diff
@@ -1,15 +0,0 @@
...
- * Copyright (c) Meta Platforms, Inc. and affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- *
- *
- */
-
-/**
- * The @stylex directive is used by the @stylexjs/postcss-plugin.
```
necolas, line 11: You should have kept this comment when moving the directive, otherwise it's not going to be clear what it means to people just browsing the nextjs code example

PR #786, opened 2024-11-27
fix: avoid traversing on `Program:enter`

nmn:
## Better compatibility with pre-transformation

StyleX is sometimes used alongside other code transformation that can generate StyleX code. Examples include the React-Strict-DOM Babel plugin and `tailwind-to-stylex`.

The StyleX compiler has to do multiple passes of the entire tree to compile and then optimize the code. Previously, some of the transformations happened using a `path.traverse` in `Program:enter`. Because of how Babel runs multiple plugins, this meant that StyleX would often transform *before* previous transforms were completed. This could result incomplete compilation.

This change moves some code around to move things to the normal tree traversal or `Program:exit`. This will ensure that StyleX can consume any code generated by previous transforms.

> [!NOTE]
> There is no functional change to the output of the Babel plugin.

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-pre-plugin-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-pre-plugin-test.js
@@ -1,0 +1,149 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
+
+'use strict';
+
+jest.autoMockOff();
+
+import * as t from '@babel/types';
+const { transformSync } = require('@babel/core');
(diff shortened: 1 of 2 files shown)


necolas:
FWIW the RSD preset doesn't generate any StyleX code

necolas:
Is this something you can unit test to validate?

nmn:
> FWIW the RSD preset doesn't generate any StyleX code

I should explain better, but RSD will transform `stylex={}` into `{...stylex.props()}` right? StyleX normally optimizes them away if all the styles are defined locally. When the `stylex.props` is inserted by a Babel plugin, it may fail to do so.

Is this something you can unit test to validate?

Yes, I don't want to add a new dependency just for a test, so I'll try coming up with a minimal Babel plugin just for the test.

*(I tested this independently on top of [tailwind-to-stylex](https://www.npmjs.com/package/tailwind-to-stylex).)*

PR #790, opened 2024-12-02
[feat] add caching for CLI file transform

mellyeliu:
This PR introduces a caching mechanism to the CLI file transform for performance.

1. Input and Output File Caching
   - Caches the input file hash, output file hash, and compiled CSS for each file
   - Ensures files are only recompiled if:
     - The input file content has changed
     - The output file content does not match the expected hash
2. Cache Storage
   - Stores cache data defaulting to `node_modules/.stylex-cache`, 
   - Will add fast follow for configurable directory (`cachePath`)
   - Each file's cache data includes:
     - `inputHash`: Hash of the input file content
     - `outputHash`: Hash of the compiled output file content
     - `collectedCSS`: The compiled CSS
3. Cache Integration
   - Before recompiling, the CLI checks the cache:
     - If hashes match, the cached CSS is reused
     - If not, the file is recompiled, and the cache is updated

## Testing

Added tests to ensure:
- Files are skipped when cache is valid
- Files are recompiled when inputs or outputs change
- Cache is correctly written and read from the specified path
## Pre-flight checklist


packages/cli/package.json
```diff
@@ -12,7 +12,7 @@
     "postbuild:esm": "rollup -c ./rollup.config.mjs",
     "build": "npm run build:cjs && npm run build:esm",
     "build-haste": "rewrite-imports -i src/ -o lib/",
-    "test": "jest"
+    "test": "NODE_ENV=test jest"
```
nmn, line 15: Please use the `cross-env` package to set the environment variable so it can also work on Windows. :(

packages/cli/__tests__/compile-stylex-folder-test.js
```diff
@@ -196,3 +203,83 @@ describe('individual testing of util functions', () => {
...
+  });
+
+  afterAll(() => {
+    fs.rmSync(config.output, { recursive: true, force: true });
+    fs.rmSync(cachePath, { recursive: true, force: true });
+  });
+
+  test('first compilation populates the cache', async () => {
+    fs.mkdirSync(config.output, { recursive: true });
+    fs.mkdirSync(cachePath, { recursive: true });
```
nmn, line 240: NIT: Use the `async` versions of these functions so more tests can run in parallel.

packages/cli/src/cache.js
```diff
@@ -0,0 +1,75 @@
...
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
+import fs from 'fs';
+import path from 'path';
+import crypto from 'crypto';
```
nmn, line 11: TODO: check if `crypto` is faster than `murmur`hash which is what we use for hashing classNames etc.

We should use whatever is fastest because we don't care about the cryptographic properties of hashing algorithms that are hard to reverse-engineer.

packages/cli/src/cache.js
```diff
@@ -0,0 +1,75 @@
...
+// $FlowFixMe
+import { mkdirp } from 'mkdirp';
```
nmn, line 13: Maybe we replace `mkdirp` everywhere with:
```
fs.mkdirSync(config.output, { recursive: true });
```

?

packages/cli/src/cache.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Default cache directory in `node_modules/.stylex-cache`
+export function getDefaultCachePath() {
+  return process.env.NODE_ENV === 'test'
+    ? path.join('node_modules', '.stylex-cache-test')
+    : path.join('node_modules', '.stylex-cache');
+}
```
nmn, line 20: Why do we need two different cache paths for test and prod?

packages/cli/src/cache.js
```diff
@@ -0,0 +1,75 @@
...
+
+function getCacheFilePath(cachePath, filePath) {
+  const fileName = filePath.replace(/[\\/]/g, '__');
```
nmn, line 23: Can we change the filePath to be relative to the project root instead? (This can be a follow-up PR)

This would make the caches portable and re-usable on different machines.

packages/cli/src/cache.js
```diff
@@ -0,0 +1,75 @@
...
+  return path.join(cachePath, `${fileName}.json`);
+}
+
+export function readCache(filePath) {
+  const cacheFile = getCacheFilePath(getDefaultCachePath(), filePath);
```
nmn, line 28: Do you have a way to override `getDefaultCachePath` already? If not, please add some comments that the "defaultCachePath` is currently the only cache path.

packages/cli/src/cache.js
```diff
@@ -0,0 +1,75 @@
...
+  if (fs.existsSync(cacheFile)) {
+    return JSON.parse(fs.readFileSync(cacheFile, 'utf-8'));
+  }
+  return null;
```
nmn, line 32: Consider making all file read/write operations asynchronous if it's not too complicated. This will make the code more easy to change in the future to support additional ways to cache.

packages/cli/src/transform.js
```diff
@@ -94,16 +100,56 @@ export async function compileFile(
...
+      oldOutputHash = computeHash(outputFilePath);
+    } catch (err) {
+      console.error(
+        `[stylex] Failed to compute hash for: ${outputFilePath}`,
+        err,
+      );
+      oldOutputHash = null;
+    }
+  } else {
+    console.log(`[stylex] Output file does not exist: ${outputFilePath}`);
```
nmn, line 118: Don't log this here. We will already log when we generate this file.

packages/cli/src/transform.js
```diff
@@ -94,16 +100,56 @@ export async function compileFile(
 ): Promise<?string> {
   const inputFilePath = path.join(config.input, filePath);
   const outputFilePath = path.join(config.output, filePath);
+  const cachePath = getDefaultCachePath();
+
+  const inputHash = computeHash(inputFilePath);
```
nmn, line 105: Yup. Making things async will be an improvement as you can compute the hash *and* read the cache *and* check if the output file exists, all in parallel.

packages/cli/src/transform.js
```diff
@@ -94,16 +100,56 @@ export async function compileFile(
...
+  let oldOutputHash = null;
+  if (fs.existsSync(outputFilePath)) {
+    try {
+      oldOutputHash = computeHash(outputFilePath);
+    } catch (err) {
+      console.error(
```
nmn, line 111: Why do we need the `try-catch` here but not when the hash is being calculated for the input file?

packages/cli/src/transform.js
```diff
@@ -94,16 +100,56 @@ export async function compileFile(
...
+  } else {
+    console.log(`[stylex] Output file does not exist: ${outputFilePath}`);
+  }
+
+  const cacheData = readCache(filePath);
+
+  if (
+    cacheData &&
+    cacheData.inputHash === inputHash &&
+    fs.existsSync(outputFilePath) &&
```
nmn, line 126: This computation is being repeated from above. (line 107)

nmn (approved):
This is a pretty solid implementation. Please cleanup the logging so it isn't so noisy at the very least.

There are improvements that can be made by making more of the code asynchronous, but nothing that should block this PR.

packages/cli/src/transform.js
mellyeliu: whoops leftover from debugging, will rm

packages/cli/src/cache.js
mellyeliu: adding this in a fast follow!

mellyeliu:
Addressed nits and async caching logic

To address in a follow up PR:
- Specifying a custom `cachePath`
- Testing on `murmurHash` vs `crypto`
- Make file path relative to project root

PR #792, opened 2024-12-03
[feat] add support for custom CLI cache path

mellyeliu:
## Context

Follow up to https://github.com/facebook/stylex/pull/790 to add custom cache path

## Testing

Added test for custom cache dir

## Pre-flight checklist


packages/cli/src/config.js
```diff
@@ -31,6 +31,7 @@ export type TransformConfig = {
   ...CliConfig,
   input: string,
   output: string,
+  cachePath?: string,
```
nmn, line 34:
```suggestion
  cachePath?: string = getDefaultCachePath(),
```

packages/cli/src/transform.js
```diff
@@ -100,7 +100,7 @@ export async function compileFile(
 ): Promise<?string> {
   const inputFilePath = path.join(config.input, filePath);
   const outputFilePath = path.join(config.output, filePath);
-  const cachePath = getDefaultCachePath();
+  const cachePath = config.cachePath || getDefaultCachePath();
```
nmn, line 103:
```suggestion
  const cachePath = config.cachePath;
```

nmn (approved):
Thanks for making this improvement!

packages/cli/src/config.js
mellyeliu: i believe typescript doesn't allow defaults in a type definition, but could create a factory function to set the default if that's preferable

nmn: This is a mistake, it should be:

```
cachePath: string = getDefaultCachePath(),
```

The `?` is not needed because it is implied when the default value is provided.

PR #794, opened 2024-12-03
feat: dynamic styles set inherits to false

### LINKED ISSUE #734: Mark variables used for dynamic styles as `inherits: false`
Generate `@property` declarations for all variables that are created for dynamic styles and declare them as not inheriting.

This improves the performance of variables in the browser.

Samantha-Zhan:

Addresses task https://github.com/facebook/stylex/issues/734?fbclid=IwZXh0bgNhZW0CMTEAAR2-QvBRO5HXoBYI-LmCO1benFCf76i8rsQqF7eQ9C3elQYx859Zpkr4rRI_aem_hLD_maBEZrcTD5svWp7J3A.

Generate `@property` declarations for all variables that are created for dynamic styles and declare them as not inheriting. Increase stylex variable compiler efficiency by disabling all dynamic styles from cascading down from the node defining the variable.

## Linked PR/Issues

Fixes #734 

## Additional Context
`npx jest stylex-transform-create-test`

Test file [stylex-transform-create-test.js](https://github.com/facebook/stylex/compare/feat/dynamic-styles-inherits-false?expand=1#diff-1ae52714c572f4282c394af92227b4198e93ffdccd6bf7c38129b36f09250280) should have `@property` injection added for all stylex dynamic styles.

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/src/visitors/stylex-create/index.js
+++ b/packages/babel-plugin/src/visitors/stylex-create/index.js
@@ -113,6 +113,19 @@ export default function transformStyleXCreate(
       memberExpressions,
     });
 
+    // add injection that mark variables used for dynamic styles as `inherits: false`
+    const injectedInheritStyles: { [string]: InjectableStyle } = {};
+    if(fns != null){
+      const dynamicFnsNames = Object.values(fns)?.map(entry => Object.keys(entry[1])).flat();
+      dynamicFnsNames.forEach(fnsName => {      
+        injectedInheritStyles[fnsName] = {
+          priority: 0,
+          ltr: `@property ${fnsName} { inherits: false }`,
+          rtl: null,
+        };
+      })
+    }
+
     if (!confident) {
       throw path.buildCodeFrameError(messages.NON_STATIC_VALUE, SyntaxError);
     }
@@ -124,6 +137,7 @@ export default function transformStyleXCreate(
     const injectedStyles = {
       ...injectedKeyframes,
       ...injectedStylesSansKeyframes,
+      ...injectedInheritStyles,
     };
 
     let varName = null;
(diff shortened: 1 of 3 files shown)


packages/babel-plugin/src/visitors/stylex-create/index.js
```diff
@@ -113,6 +113,21 @@ export default function transformStyleXCreate(
...
+          priority: 0,
+          ltr: `@property ${fnsName} { inherits: false }`,
+          rtl: null,
+        };
+      });
+    }
+
     if (!confident) {
       throw path.buildCodeFrameError(messages.NON_STATIC_VALUE, SyntaxError);
     }
```
nmn, line 133: Make sure you add this code *after* the `if` condition that checks `confident`. Otherwise you'll end up throwing errors because `fns` will be `null` or `undefined`.

```suggestion
    if (!confident) {
      throw path.buildCodeFrameError(messages.NON_STATIC_VALUE, SyntaxError);
    }

    // add injection that mark variables used for dynamic styles as `inherits: false`
    const injectedInheritStyles: { [string]: InjectableStyle } = {};
    if (fns != null) {
      const dynamicFnsNames = Object.values(fns)
        ?.map((entry) => Object.keys(entry[1]))
        .flat();
      dynamicFnsNames.forEach((fnsName) => {
        injectedInheritStyles[fnsName] = {
          priority: 0,
          ltr: `@property ${fnsName} { inherits: false }`,
          rtl: null,
        };
      });
    }
```

nmn (approved):

packages/babel-plugin/src/visitors/stylex-create/index.js
Samantha-Zhan: Great call, thanks for fixing it!

necolas:
Created new tasks: https://github.com/facebook/stylex/issues/799, https://github.com/facebook/stylex/issues/800

PR #803, opened 2024-12-06
[feat] better dev theme names

mellyeliu:

Please include relevant motivation and context

## Linked PR/Issues

Fixes # (issue)

## Additional Context

## Pre-flight checklist


packages/shared/src/stylex-define-vars.js
```diff
@@ -51,9 +51,17 @@ export default function styleXDefineVars<Vars: VarsConfig>(
...
+    +[string]: {
+      +nameHash: string,
+      +value: VarsConfigValue,
+      +themeNamePrefix: string,
+    },
   } = objMap(variables, (value, key) => {
     // Created hashed variable names with fileName//themeName//key
+
+    const processedKey = key.startsWith('--') ? key.slice(2) : key;
+    const themeNamePrefix = debug ? `${processedKey}` : '';
```
mellyeliu, line 63: let's sync on what we want to prefix here, I'm putting `processedKey` as a placeholder but can be modified to be themeName-processedKey etc. I noticed in the testing that some of the themeNames include special characters beyond alphanumeric and dash symbols, which could potentially break things, so it'd be good to validate

packages/shared/__tests__/define-vars/stylex-define-vars-test.js
```diff
@@ -186,6 +186,156 @@ describe('stylex-define-vars test', () => {
...
+          "priority": 0.1,
+          "rtl": null,
+        },
+        "bgColorDisabled-x568ih9-bdddrq": {
+          "ltr": "@media print{:root, .bgColorDisabled-x568ih9{--xgck17p:white;}}",
+          "priority": 0.1,
+          "rtl": null,
+        },
+        "bgColorDisabled-x568ih9-kpd015": {
+          "ltr": "@supports (color: oklab(0 0 0)){:root, .bgColorDisabled-x568ih9{--xpegid5:oklab(0.7 -0.3 -0.4);}}",
```
necolas, line 281: Is this output correct? If looks like the source is a "supports" nested in a "media" for dark mode, but the output is missing the media part.

It's a little difficult to read the test because there are several patterns in the input and a lot of generated content as a result. It might be worth breaking it up so it's easier to check the output by eye

packages/shared/src/stylex-define-vars.js
nmn: If the key starts with `--` don't add a prefix at all. In those cases we don't generate a variable name and use the key name as is.

packages/shared/__tests__/define-vars/stylex-define-vars-test.js
nmn: Yes, `--xpegid5:` doesn't have the prefix.

Instead the className that defines the variables has the prefix which is likely wrong.

packages/shared/__tests__/define-vars/stylex-define-vars-test.js
```diff
@@ -186,6 +186,156 @@ describe('stylex-define-vars test', () => {
...
+        default: 'pink',
+      },
+    };
+    const [jsOutput, cssOutput] = styleXDefineVars(defaultVars, {
+      themeName,
+      debug: true,
+    });
+
+    expect(jsOutput).toEqual({
+      __themeName__: classNamePrefix + createHash(themeName),
```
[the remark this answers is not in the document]
mellyeliu: good callout! will address this in a follow up as it seems to be the default pattern across all the tests in the test suite

mellyeliu:
updated to prefix the _variable names_ with key as discussed with @nmn

packages/babel-plugin/__tests__/stylex-transform-define-vars-test.js
```diff
@@ -279,6 +280,41 @@ describe('@stylexjs/babel-plugin', () => {
...
+        ).code,
+      ).toMatchInlineSnapshot(`
+        "import stylex from 'stylex';
+        export const buttonTheme = {
+          bgColor: "var(--xgck17p)",
+          bgColorDisabled: "var(--xpegid5)",
+          cornerRadius: "var(--xrqfjmn)",
+          fgColor: "var(--x4y59db)",
+          __themeName__: "x568ih9"
+        };"
```
nmn, line 314: The variables don't have the prefixes in this test?

The expected output should be:

```
import stylex from 'stylex';
export const buttonTheme = {
 bgColor: "var(--bgColor-xgck17p)",
 bgColorDisabled: "var(--bgColorDisabled-xpegid5)",
 cornerRadius: "var(--cornerRadius-xrqfjmn)",
 fgColor: "var(--fgColor-x4y59db)",
 __themeName__: "x568ih9"
};
```

You changes to `stylex-define-vars` look correct, so will need to debug why that isn't working here.

packages/shared/__tests__/define-vars/stylex-define-vars-test.js
```diff
@@ -186,6 +186,85 @@ describe('stylex-define-vars test', () => {
...
+        classNamePrefix +
+        createHash(`${themeName}.bgColorDisabled`)
+      })`,
+      cornerRadius: `var(--${
+        'cornerRadius-' +
+        classNamePrefix +
+        createHash(`${themeName}.cornerRadius`)
+      })`,
+      fgColor: `var(--${'fgColor-' + classNamePrefix + createHash(`${themeName}.fgColor`)})`,
+    });
```
nmn, line 235: Can be a bit simpler this way:

```suggestion
    expect(jsOutput).toEqual({
      __themeName__: classNamePrefix + createHash(themeName),
      bgColor: `var(--bgColor-${classNamePrefix + createHash(`${themeName}.bgColor`)})`,
      bgColorDisabled: `var(--bgColorDisabled-${
        classNamePrefix +
        createHash(`${themeName}.bgColorDisabled`)
      })`,
      cornerRadius: `var(--cornerRadius-${
        classNamePrefix +
        createHash(`${themeName}.cornerRadius`)
      })`,
      fgColor: `var(--fgColor-${classNamePrefix + createHash(`${themeName}.fgColor`)})`,
    });
```

packages/babel-plugin/__tests__/stylex-transform-define-vars-test.js
```diff
@@ -279,6 +280,41 @@ describe('@stylexjs/babel-plugin', () => {
...
+              default: 'grey',
+              '@media (prefers-color-scheme: dark)': 'rgba(0, 0, 0, 0.8)',
+            },
+            cornerRadius: 10,
+            fgColor: {
+              default: 'pink',
+            },
+          });
+        `,
+          { debug: true, ...defaultOpts },
```
nmn, line 304: Oh ignore the other comment. This is the problem.

```suggestion
          { ...defaultOpts, debug: true },
```

nmn (changes requested):
I think there's only a bug in the test. The code changes look good.

nmn (approved):

PR #804, opened 2024-12-06
chore: refactor cli cache tests to be async

mellyeliu:
## Context

Follow up to https://github.com/facebook/stylex/pull/790 and https://github.com/facebook/stylex/pull/792

Let's clean up the tests and move to async functions 

## Pre-flight checklist

### DIFF
--- a/packages/cli/__tests__/compile-stylex-folder-test.js
+++ b/packages/cli/__tests__/compile-stylex-folder-test.js
@@ -15,7 +15,7 @@ import { compileDirectory } from '../src/transform';
 import * as cacheModule from '../src/cache';
 import { getDefaultCachePath } from '../src/cache';
 
-import fs from 'fs';
+const fs = require('fs').promises;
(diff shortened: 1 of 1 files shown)


packages/cli/__tests__/compile-stylex-folder-test.js
```diff
@@ -15,44 +15,56 @@ import { compileDirectory } from '../src/transform';
 import * as cacheModule from '../src/cache';
 import { getDefaultCachePath } from '../src/cache';
 
-import fs from 'fs';
+const fs = require('fs').promises;
```
nmn, line 18:
```suggestion
const fs = require('fs/promises')
```

nmn (approved):
lgtm

packages/cli/__tests__/compile-stylex-folder-test.js
mellyeliu: current type definitions are outdated so flow is erroring on this, will keep for now

PR #805, opened 2024-12-10
feat: add autofix for logical to inline properties in valid-styles rule

mellyeliu:
## Context

As per internal discussions, let's add an autofix into the valid-styles rule. Some of this logic exists in the valid-shorthands rule, but it makes sense to move the erroring here for the repos in which valid-shorthands is not enabled.

I've kept the descriptor in the autofix short and simple to avoid clutter on a lint error that will likely crop up a lot, but can also link to https://fb.workplace.com/groups/stylexfyi/posts/1769833193785624 or something related as context. Open to feedback here!

As a next step, we should work towards splitting up the valid-styles rule into more discrete chunks to increase the granularity of error reporting/messaging (perhaps after the codemodding)

## Tests

Added tests for the following styles: 
- marginStart -> marginInlineStart
- marginEnd -> marginInlineEnd
- marginHorizontal -> marginInline
- marginVertical -> marginBlock
- paddingStart -> paddingInlineStart
- paddingEnd -> paddingInlineEnd
- paddingHorizontal -> paddingInline
- paddingVertical -> paddingBlock

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -584,6 +584,72 @@ eslintTester.run('stylex-valid-styles', rule.default, {
       ],
     },
     {
+      code: `
+        import stylex from 'stylex';
+        const styles = stylex.create({
+          invalidStyle: {
+            marginStart: '10px',
+            marginEnd: '5px',
+            marginHorizontal: '15px',
+            marginVertical: '15px',
+            paddingStart: '10px',
+            paddingEnd: '5px',
+            paddingHorizontal: '5px',
+            paddingVertical: '15px',
+          },
+        });
+      `,
+      errors: [
+        {
+          message:
+            'The key "marginStart" is not allowed by stylex. Use "marginInlineStart" instead.',
+        },
+        {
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -583,6 +598,72 @@ eslintTester.run('stylex-valid-styles', rule.default, {
...
+            paddingEnd: '5px',
+            paddingHorizontal: '5px',
+            paddingVertical: '15px',
+          },
+        });
+      `,
+      errors: [
+        {
+          message:
+            'The key "marginStart" is not allowed by stylex. Use "marginInlineStart" instead.',
```
necolas, line 620: Maybe something like this:

"The key 'marginStart' is not a standard CSS property. Did you mean 'marginInlineStart'?"

Communicating that it's non-standard is useful context for people

necolas (approved):
LGTM (once message bikeshedding is complete)

mellyeliu:
> LGTM (once message bikeshedding is complete)

I had to google this #til Parkinson's Law of Triviality

Issue #799, opened 2024-12-05
Add dynamic styles to benchmarks fixture


necolas:
### Describe the feature request

We don't have any dynamic styles syntax in the fixture data used by the bundle-size benchmark. This means we don't capture the impact on generated CSS from the current approach to implementing dynamic styles.. Include a significant number of dynamic styles in the fixture to resolve this issue.

Samantha-Zhan:
I can attempt this task if no one is on it since it relates to a previous task I was working on. However, I haven't worked with benchmarks fixture before, would you mind give me some pointers on where to get started?

necolas:
These are the relevant files/scripts:

* The rollup-example app contains a (very big) fixture called [`lotsOfStyles.js`](https://github.com/facebook/stylex/blob/main/apps/rollup-example/lotsOfStyles.js)
* The [benchmark workflow](https://github.com/facebook/stylex/blob/main/.github/workflows/performance.yml) builds the packages/apps.
* It then calls the [size.js script](https://github.com/facebook/stylex/blob/main/packages/scripts/size.js) (which you can also run locally)
* The size script generates size data using the build in the rollup example.

If we add [dynamic styles](https://stylexjs.com/docs/learn/styling-ui/defining-styles/#dynamic-styles) to the `lotsOfStyles.js` file, we should see an increase in the rollup bundle size data generated by the script. Maybe simplest to add a large block of dynamic styles at the bottom of the fixture, but open to suggestions.

nmn:
Thanks for taking this on @Samantha-Zhan!

The benchmark script has been flaky recently so this task might involve a bit of investigation to figure out why and to fix it.

nmn:
@Samantha-Zhan The flakiness is with the github action specifically, not the script itself. When you're done with your PR that's when that will become relevant.

necolas:
I don't think it's flakey. It doesn't run on PR's from forks

Samantha-Zhan:
@necolas I added some styles inside the `lotsOfStyles.js` fixture, and was trying to rebuild the `script.js` file. What I did:
- `npm install --save-dev @stylexjs/rollup-plugin`
- 1npm run build`
However, I got following error on build:
```
(!) Unresolved dependencies
https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency
@stylexjs/stylex (imported by "index.js" and "lotsOfStyles.js")
[!] (plugin rollup-plugin-stylex) SyntaxError: Invalid @ rule body
[...]

PR #806, opened 2024-12-11
chore: fix @property inherits syntax

Samantha-Zhan:

This PR is a quick fix to resolve `AtRuleBodyInvalid` `SyntaxError: Invalid @ rule body` error caught in while doing task #799 ([see discussion](https://github.com/facebook/stylex/issues/799#issuecomment-2537296024)). This is a bug caused by previous PR [feat: dynamic styles set inherits to false #794](https://github.com/facebook/stylex/pull/794/files#diff-879871ad922cfbbd8d59d46cd33adfccaa6305fb88c5b22dfdd5c06f5c945bf5R130), where we did not include mandatory descriptors `syntax` and `initial-value` for `@property` (see MDS docs [@property](https://developer.mozilla.org/en-US/docs/Web/CSS/@property) [initial-value](https://developer.mozilla.org/en-US/docs/Web/CSS/@property/initial-value) [syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/@property/syntax)).

> BEFORE
```
injectedInheritStyles[fnsName] = {
          priority: 0,
          ltr: `@property ${fnsName} { inherits: false }`,
          rtl: null,
        };
});
```

> AFTER
```
injectedInheritStyles[fnsName] = {
          priority: 0,
          ltr: `@property ${fnsName} { syntax: "*"; inherits: false; initial-value: "*";}`,
          rtl: null,
};
```

## Linked PR/Issues

Fixes #794 

## Additional Context

running rollup generate bundle.js with no error: `npx rollup --config ./rollup.config.mjs`

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/__tests__/evaluation/stylex-import-evaluation-test.js
+++ b/packages/babel-plugin/__tests__/evaluation/stylex-import-evaluation-test.js
@@ -369,7 +369,7 @@ describe('Evaluation of imported values works based on configuration', () => {
         import 'otherFile.stylex';
         import { MyTheme } from 'otherFile.stylex';
         _inject2(".__hashed_var__b69i2g{--__hashed_var__1jqb1tb:var(----__hashed_var__1jqb1tb)}", 1);
-        _inject2("@property ----__hashed_var__1jqb1tb { inherits: false }", 0);
+        _inject2("@property ----__hashed_var__1jqb1tb { syntax: \\"*\\"; inherits: false; initial-value: \\"*\\";}", 0);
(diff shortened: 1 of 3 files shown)


necolas (approved):
Good find.

BTW, to run the rollup build in the stylex repo, you can run the relevant worksplace script:

```
npm run build -w rollup-example
```

The benefit of doing this is that it wraps the specifics of the script (config path, etc). In general, all scripts within the repo can be called using `npm run <script-name> -w <workspace-name>` from the root

PR #807, opened 2024-12-12
feat: add dynamic styles to benchmarks fixture

### LINKED ISSUE #799: Add dynamic styles to benchmarks fixture

Samantha-Zhan:

This PR added a feature as requested by task #799 , adding dynamic styles to benchmark  fixture, which can then be used by [benchmark workflow](https://github.com/facebook/stylex/blob/main/.github/workflows/performance.yml) on each PR to check performance. This new addition can also help us quantify any performance improvement made to compile dynamic styles, in addition to existing support for static style(e.g. [feat: dynamic styles set inherits to false #794](https://github.com/facebook/stylex/pull/794)).

## Linked PR/Issues

Fixes #799 

## Additional Context

We do:
- adding additional style examples to fixture
- run `npm run build -w rollup-example` to update `bundle.js`
- we observe generated `bundle.js` to contain the new dynamic styles and observe a size increase when running `./packages/scripts/size.js` script
**[BEFORE]**
image  

**[AFTER]**
image

 

## Pre-flight checklist

### DIFF
--- a/apps/rollup-example/lotsOfStylesDynamic.js
+++ b/apps/rollup-example/lotsOfStylesDynamic.js
@@ -174,8 +174,8 @@ export const lotsOfStylesDynamic = [
     }),
   }),
   stylex.create({
-    dynamicContainer: (marginEnd) => ({
-      marginEnd,
+    dynamicContainer: (marginInlineEnd) => ({
+      marginInlineEnd,
     }),
     dynamicKeyInfo: (
       backgroundColor,
@@ -183,7 +183,7 @@ export const lotsOfStylesDynamic = [
       borderStyle,
       borderColor,
       borderRadius,
-      marginEnd,
+      marginInlineEnd,
       padding,
     ) => ({
       backgroundColor,
@@ -191,7 +191,7 @@ export const lotsOfStylesDynamic = [
       borderStyle,
       borderColor,
       borderRadius,
-      marginEnd,
+      marginInlineEnd,
       padding,
     }),
     dynamicKeyInfoItem: (marginTop) => ({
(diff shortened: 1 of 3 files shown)


apps/rollup-example/lotsOfStylesDynamic.js
```diff
@@ -0,0 +1,44 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
+
+'use strict';
+
+import * as stylex from '@stylexjs/stylex';
+
+export const lotsOfStylesDynamic = [
```
necolas, line 14: Good to see this setup working. I think we'll need 10-20x more dynamic styles for their impact to show up enough in the bundle benchmark

Samantha-Zhan: Got it, I'm adding more styles to the file.

apps/rollup-example/lotsOfStylesDynamic.js
```diff
@@ -0,0 +1,807 @@
...
+      borderStyle,
+      borderColor,
+      borderRadius,
+      display,
+      lineHeight,
+      margin,
+      minWidth,
+      padding,
+      paddingEnd,
+      paddingStart,
```
necolas, line 261: We should use the standard properties (paddingInlineEnd etc) in fixtures going forward.

necolas (approved):
LGTM. Thanks

apps/rollup-example/lotsOfStylesDynamic.js
Samantha-Zhan: That's a great call, using standardize naming would definitely help since these serves as examples for users. I made my best attempt fixing all non-standard property names (according to gpt); I will followup with a chore PR since I also made the same changes to `lotsOfStyles.js` for both dynamic&static fixture to be in sync.

PR #814, opened 2024-12-17
fix: fix valid-shorthands object casting error on css-shorthand-expand null return

mellyeliu:

Certain complex shorthand values like `'4px solid var(--fds-gray-10)'` seem to trip up the css-shorthand-expand package, which defaults to returning null. Let's handle the cases by prematurely returning and not providing an autofix in such cases.

## Testing
**_Still identifying repros to make the test more robust_**

### DIFF
--- a/packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
@@ -173,6 +173,22 @@ eslintTester.run('stylex-valid-shorthands', rule.default, {
     },
     {
       code: `
+      import stylex from 'stylex';
+      const styles = stylex.create({
+        main: {
+          borderRight: '4px solid var(--fds-gray-10)'
+        },
+      })
+    `,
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
```diff
@@ -171,6 +171,22 @@ eslintTester.run('stylex-valid-shorthands', rule.default, {
         },
       ],
     },
+    {
+      code: `
+      import stylex from 'stylex';
+      const styles = stylex.create({
+        main: {
+          borderRight: '4px solid var(--fds-gray-10)'
```
mellyeliu, line 179: still identifying repros to make the test more robust

Prakshal-Jain (approved):
Assuming we are aligned that we should return if css shorthand expand fails, the changes LGTM.
@nmn please feel free to chime in since you have more context 🙂

mellyeliu:
Merging as SEV follow up

PR #815, opened 2024-12-18
feat: rollup plugin support generate file with pattern

### LINKED ISSUE #778: [hash] in fileName

Hello, I'm using the rollup plugin and I went to output the css file with a hash, I tried adding `[hash]` in `fileName` but it doesn't get replaced by a hash, can this feature be supported, if yes, how to do it (I can also implement it and make PR).

Many thanks.

```js
import stylexPlugin from '@stylexjs/rollup-plugin';

const config = {
  input: './index.js',
  output: {
    file: './.build/bundle.js',
    format: 'es',
  },
  // Ensure that the stylex plugin is used before Babel
  plugins: [stylexPlugin({
    // Required. File path for the generated CSS file.
    fileName: './.build/stylex.[hash].css',
    // default: false
    dev: false,
    // prefix for all generated classNames
    classNamePrefix: 'x',
    // Required for CSS variable support
    unstable_moduleResolution: {
      // type: 'commonJS' | 'haste'
      // default: 'commonJS'
      type: 'commonJS',
      // The absolute path to the root directory of your project
      rootDir: __dirname,
    },
  })],
};

export default config;
```

nonzzz:

Until now I found that no one has implemented it.

## Linked PR/Issues

Fixes #778 

## Additional Context

## Pre-flight checklist

### DIFF
--- a/packages/rollup-plugin/src/index.js
+++ b/packages/rollup-plugin/src/index.js
@@ -24,6 +24,7 @@ import type {
 } from 'rollup';
 import browserslist from 'browserslist';
 import { browserslistToTargets } from 'lightningcss';
+import crypto from 'crypto';
 
 const IS_DEV_ENV =
   process.env.NODE_ENV === 'development' ||
@@ -43,6 +44,15 @@ export type PluginOptions = $ReadOnly<{
   >,
   ...
 }>;
+
+function replaceFileName(original: string) {
+  const hash = crypto
+    .createHash('sha256')
+    .update(crypto.randomBytes(16))
+    .digest('hex')
+    .slice(0, 8);
+  return original.replace(/\[hash\]/g, hash);
+}
 
 export default function stylexPlugin({
   dev = IS_DEV_ENV,
@@ -72,7 +82,7 @@ export default function stylexPlugin({
         const { code } = transform({
           targets: browserslistToTargets(browserslist('>= 1%')),
           ...lightningcssOptions,
-          filename: fileName,
+          filename: 'stylex.css',
           code: Buffer.from(collectedCSS),
         });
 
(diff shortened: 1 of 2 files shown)


packages/rollup-plugin/src/index.js
```diff
@@ -44,6 +45,15 @@ export type PluginOptions = $ReadOnly<{
...
 }>;
 
+function replaceFileName(original: string) {
+  const hash = crypto
+    .createHash('sha256')
+    .update(crypto.randomBytes(16))
+    .digest('hex')
+    .slice(0, 8);
+  return original.replace(/\[hash\]/g, hash);
+}
```
nmn, line 55:
```suggestion
function replaceFileName(original: string) {
  if (!original.includes('[hash]')) {
    return original;
  }
  const hash = crypto
    .createHash('sha256')
    .update(crypto.randomBytes(16))
    .digest('hex')
    .slice(0, 8);
  return original.replace(/\[hash\]/g, hash);
}
```

nmn: Also, shouldn't the hash be hashing the contents of the generated file?

nmn (changes requested):
Some small changes suggested.

packages/rollup-plugin/src/index.js
nonzzz: No, in my opinion that feature request is on hashing the generated file name not the contents of the generated file. By the way following rollup document, we can use `assetFileNames` do this thing but it need to split chunk by themselves. (But we need change the emit logic)

nmn: from everything I've read the point of putting a hash in the file name is for a cache breaker, which should likely be based on the file contents.

nonzzz: My bad, now i update the code :)

nmn:
Thanks!

PR #817, opened 2024-12-20
fix: add inline processing for borderWidth, borderColor, borderStyle to valid-shorthands

mellyeliu:
## Context
When using the `preferInline` config, we were expanding margin/padding to equivalent inline logical properties. However, we overlooked applying the same transform to certain border properties.

The change introduced here extends the inline logical transformation to border properties (specifically borderWidth, borderStyle, and borderColor) by applying the same logic used for margin and padding. We take the output from `css-shorthand-expand` and apply a transform to the physical properties returned.

## Testing

Added test expansions for borderWidth, borderStyle, and borderColor with `preferInline` set to default (false) and true

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/src/utils/splitShorthands.js
+++ b/packages/eslint-plugin/src/utils/splitShorthands.js
@@ -29,6 +29,7 @@ export const createSpecificTransformer = (
       rawValue.toString(),
       allowImportant,
       typeof rawValue === 'number',
+      _preferInline,
     );
   };
 };
@@ -37,7 +38,6 @@ export const createDirectionalTransformer = (
   baseProperty: string,
   blockSuffix: string,
   inlineSuffix: string,
-  basePropertySuffix: string = '',
 ): ((
   rawValue: number | string,
   allowImportant?: boolean,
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/utils/splitShorthands.js
```diff
@@ -189,13 +191,33 @@ export function splitSpecificShorthands(
...
+  const directionMap = {
+    left: 'inline-start',
+    right: 'inline-end',
+  };
+
+  // List directional properties for later inline transform if needed
+  const directionalProperties = [
+    'border-width',
+    'border-color',
+    'border-style',
```
necolas, line 203: We should include `borderRadius` too. Is there anything else? Where is the configuration for enabling this for `padding` and `margin` located? Do we know if OSS users get consistent behavior across properties (it should default for physical properties for every shorthand)

mellyeliu: These are the supported properties, I believe it's just these that are directional:

borderColor
borderWidth
borderStyle
borderRadius
margin
padding
marginBlock
marginInline
paddingBlock
paddingInline

And these that are left as is:
borderTop
borderRight
borderBottom
borderLeft
background
font
outline

There's a global config `preferInline` that's set when configuring the lint rule; it's default false for all properties, so by default all properties use physical properties when this config is not set.

necolas (approved):

PR #834, opened 2025-01-04
fix: bypass kebab-case conversion for CSS variable names

### LINKED ISSUE #821: CamelCased variable names are unexpected turned into kebab-case when setting them
Previously, all variable names are always hashes that only included small-case letters.

Now, with the `debug` option set to true, variables are prefixed with their in-source key name.

This behavior has exposed a bug in how StyleX set the values of CSS variables. We always convert camelCase to kebab-case for CSS properties. We need to exclude CSS variables.

yasuhiro-yamamoto:
## What changed / motivation?
Fixed an issue where CSS variable names in camelCase were being incorrectly converted to kebab-case. The bug occurred because StyleX was applying its camelCase to kebab-case conversion logic to all properties, including CSS custom properties (variables). Added a check to preserve the original case for CSS variables that start with '--'.

## Linked PR/Issues
Fixes #821

## Pre-flight checklist

### DIFF
--- a/apps/nextjs-example/app/CardTokens.stylex.ts
+++ b/apps/nextjs-example/app/CardTokens.stylex.ts
@@ -10,5 +10,5 @@
 import * as stylex from '@stylexjs/stylex';
 
 export const tokens = stylex.defineVars({
-  arrow_transform: 'translateX(0)',
+  arrowTransform: 'translateX(0)',
 });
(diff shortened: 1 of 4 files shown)


nmn:
Sorry, I noticed your PR after I fixed this problem myself.

yasuhiro-yamamoto:
@nmn Thanks for letting me know! Glad the issue is resolved.

Issue #769, opened 2024-11-11
vars used in `stylex.create` generate error with eslint(@stylexjs/valid-styles)


Mikescops:
### Describe the issue

My project is in typescript and I use the `@stylexjs/eslint-plugin` package.

I'm defining variables as follow:

```ts
export const tokens = stylex.defineVars({
    // Background and Foreground
    background: "#ffffff",
    foreground: "#171717",
```

Whenever I use my created variable as below, eslint complains about the variable not being the expected type.

```ts
const styles = stylex.create({
    primary: {
        backgroundColor: tokens.background
    }
})
```

<img width="378" alt="image" src="https://github.com/user-attachments/assets/f9400abd-8b8d-4908-aa11-2743e1aac769">

> warning  color value must be one of: a string literal ...

<img width="314" alt="image" src="https://github.com/user-attachments/assets/378a2cf0-32d4-4f2a-93f8-a75d611d8f32">

### Expected behavior

eslint should not complain about the variable as it is meant to be written that way

### Steps to reproduce

Create a project with the eslint plugin and create variables as described in the problem statement above.
When used in `stylex.create` eslint should display an error (or warning depending on your eslint settings).

### Test case

```ts
// tokens.stylex.ts
 export const tokens = stylex.defineVars({
    background: "#ffffff"
    });

// App.tsx
const styles = stylex.create({
    primary: {
        backgroundColor: tokens.background
    }
});
```

nmn:
`stylex.defineVars` must be defined in it's own file and the file must be named with a `.stylex.ts` extension.

When you import the variables directly from that file, you should not see this error.

It is currently not legal to define variables with `defineVars` in the same file as where they are used.

Mikescops:
@nmn I may have simplified how I use it for writing this issue, but initially I indeed defined them as you mentioned and I still have the error.

<img width="759" alt="image" src="https://github.com/user-attachments/assets/4382a8f9-850b-4e8a-a1af-9425bfa41873">

Mikescops:
So I've figure out something interesting, I removed the `"allowImportingTsExtensions": false` in tsconfig.

And here is the behaviour:
<img width="721" alt="image" src="https://github.com/user-attachments/assets/83e102fe-4a07-4e74-b295-722c4fde7f20">

without `.js` in the import

<img width="753" alt="image" src="https://github.com/user-attachments/assets/0b0b3061-faa1-4af5-9019-c02aae716ec2">
[...]

PR #839, opened 2025-01-05
fix(@stylexjs/babel-plugin): Support .js resolved file extension imports from .ts files

beaumontjonathan:

In a TypeScript project using `moduleResolution: 'node16' | 'nodenext'`, relative imports must include the file extension because of the Node.js ESM resolver [requires file extensions](https://nodejs.org/docs/latest-v18.x/api/esm.html#mandatory-file-extensions).

Consider the following:

```ts
import { vars } from '../vars.stylex.js';
import { theme } from '../theme.stylex.js';

  root: {
    backgroundColor: vars.primary,
  },
});
```

The babel plugin currently attempts to verify whether these relative import files actually exist. It does this by trying to resolve the import string relative to the source file. First by checking the import string itself, then checking each of the valid file extensions in turn (`js,ts,tsx,jsx,mjs,cjs`).

The case above is rejected because it will attempt to resolve e.g. `../vars.stylex.js`, `../vars.stylex.js.ts` but not `../vars.stylex.ts`. This results in a `Only static values are allowed inside of a stylex.create() call` compile error.

This change first checks whether the file extension in the relative import string is one of the recognized extensions, and if so, will remove the extension before attempting to resolve the valid extensions.

## Linked PR/Issues

Related to https://github.com/facebook/stylex/issues/769, which has a similar issue but with the ESLint plugin instead. I'm happy to take a look at the ESLint plugin if this change is accepted.

## Additional Context

I've added a test for this, but not sure it matches the existing code style because it makes use of `jest.mock(..)` for `@dual-bundle/import-meta-resolve`. I don't love this, but I could not find any other test examples for `moduleResolution: 'commonJS'`.

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/__tests__/evaluation/stylex-import-evaluation-test.js
+++ b/packages/babel-plugin/__tests__/evaluation/stylex-import-evaluation-test.js
@@ -406,9 +406,16 @@ describe('Evaluation of imported values works based on configuration', () => {
   });
 
   describe('Module resolution commonJS', () => {
+    afterEach(() => {
+      moduleResolve.mockReset();
+    });
+
     test('Recognizes .ts stylex imports when resolving .js relative imports', () => {
-      moduleResolve.mockReturnValue({
-        pathname: '/project/otherFile.stylex.ts',
+      moduleResolve.mockImplementation((value) => {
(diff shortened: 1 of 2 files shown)


nmn:
Thanks! Code changes look good. I'll do some final testing before merging.

Issue #829, opened 2025-01-02
Allow `stylex.create` nested within TS namespaces (and more?)


nmn:
### Discussed in https://github.com/facebook/stylex/discussions/827

<div type='discussions-op-text'>

<sup>Originally posted by **zaydek** December 30, 2024</sup>
I love using TS namespaces to create conceptual boundaries so it's easier to reason about what different chunks of my code do. Like a `Button` namespace so I can aggregate relevant variables and components for buttons. One thing I've noticed is that StyleX is incompatible with TS namespaces so I can't do something like `namespace Foo { const styles = stylex.create({ ... }); }`.

I believe what's going on is that TS namespaces are syntax sugar for objects and therefore it may be impossible to add support for this. That being said, is there a way for me to colocate styles behind something like a namespace so I can do something like the following in the future?

```tsx
namespace A {
  const styles = stylex.create({
    foo: { ... },
    bar: { ... },
  });
  
  export function Foo() { return <div {...stylex.props(styles.foo)} /> }
  export function Bar() { return <div {...stylex.props(styles.bar)} /> }
}

namespace B {
  const styles = stylex.create({
    foo: { ... },
    bar: { ... },
  });
  
  export function Foo() { return <div {...stylex.props(styles.foo)} /> }
  export function Bar() { return <div {...stylex.props(styles.bar)} /> }
}

...

<A.Foo />
<B.Foo />
```

I know I can simply create boundary lines using files but I find namespaces allow me to do the job of several files in one file, but when it comes to StyleX APIs, it breaks this convention and I need to prefix styles, which is OK but slightly annoying.

Any ideas or recommendations?

Here is an example of some code where I'd love to have the ability to nest the styles under the namespace instead of prefixing the styles name:

<img width="643" alt="Screenshot 2024-12-29 at 8 52 24 PM" src="https://github.com/user-attachments/assets/0d4f4e04-cbc4-43c1-9d1e-7b186a3d27bd" /></div>

zaydek:
It's worth commenting that, in case anyone works on this, it would be ideal if this works for all StyleX APIs and not just `stylex.create`.

I'd really enjoy being able to do something like:

```ts
export namespace Theme {
  export const Spacing = stylex.defineVars({
    ...
  })
  
  ...
}

nonzzz:
Let me do it.

PR #841, opened 2025-01-07
Feat:  Allow `stylex.create` nested within TS namespaces

nonzzz:

Now current work is WIP, but i need to clarify some situations.

## Linked PR/Issues

#829 

## Additional Context

> Q: Should we add a `manipulateOptions` for babel plugin to infer the coming file is typescript or others?
> Q: Should the `export expression` check of Theme's API be cancelled? If embedded in typescript namespace.

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-create-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-create-test.js
@@ -36,6 +36,26 @@ function transform(source, opts = {}) {
   }).code;
 }
 
+function transformTypescript(source, opts = {}) {
+  return transformSync(source, {
+    filename: opts.filename,
+    parserOpts: {
+      plugins: ['typescript'],
+    },
+    babelrc: false,
+    plugins: [
+      [
+        stylexPlugin,
+        {
+          runtimeInjection: true,
+          unstable_moduleResolution: { type: 'haste' },
+          ...opts,
+        },
+      ],
+    ],
+  }).code;
+}
+
 describe('@stylexjs/babel-plugin', () => {
   describe('[transform] stylex.create()', () => {
     test('transforms style object', () => {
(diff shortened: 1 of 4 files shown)


packages/babel-plugin/__tests__/stylex-transform-create-test.js
```diff
@@ -36,6 +36,26 @@ function transform(source, opts = {}) {
...
+        {
+          runtimeInjection: true,
+          unstable_moduleResolution: { type: 'haste' },
+          ...opts,
+        },
+      ],
+    ],
+  }).code;
+}
+
```
nmn, line 58: Can we update the existing `transform` function to use typescript based on `opts.filename` instead?

That way, we won't need two mostly duplicate functions.

nmn (approved):
Could you make one quick change to the tests and then I can merge

packages/babel-plugin/__tests__/stylex-transform-create-test.js
nonzzz: BTW, I think that's not enough ( Maybe shouldn't judge whether it is a `tsBlock`) The final conversion result of each compiler

```js

import stylex from "@stylexjs/stylex";
export var X;
((X2) => {
  X2.styles = stylex.create({
    t: {
      color: "red"
    }
  });
})(X || (X = {}));

```

If we want to support namespace, then we should also support this.

nmn: If this is the output then this is not yet going to be feasible.

We can *hoist* up the generated objects to the top level of the module and replace the original `stylex.create` with a reference to that object. That way we can allow defining `stylex.create` anywhere you want.

nmn:
#1137 should fix this issue.

mellyeliu:
Addressed in https://github.com/facebook/stylex/pull/1137

PR #847, opened 2025-01-11
fix: cli should rewrite the aliases

### LINKED ISSUE #846: [CLI] Rewrite aliases when compiling a folder
It is common to use the `@/` alias in Next.js. When using the CLI, the tsconfig is usually changed to update the alias to point to `source` instead of `src`.

Today, this means that any `@/` import in the `src/` folder will resolve to the uncompiled files in `source` and cause runtime errors.

We should rewrite aliases as part of the CLI to avoid dealing with this problem manually.

referece: #844

nmn:

The CLI will now rewrite imports based on the `aliases` that are passed into `stylexConfig`.

Resolves #846 

### Side note:

It might make sense to extract some of the utilities that were copied from the `babel-plugin` to a separate `utils` package for such use-cases.


packages/cli/src/config.js
```diff
@@ -24,7 +24,7 @@ export type CliConfig = {
   babelPluginsPost?: $ReadOnlyArray<any>,
   modules_EXPERIMENTAL: $ReadOnlyArray<ModuleType>,
   useCSSLayers?: boolean,
-  styleXConfig?: { +[string]: mixed },
+  styleXConfig?: StyleXOptions,
```
Jta26, line 27: nit: match naming

packages/cli/src/plugins.js
```diff
@@ -116,3 +121,143 @@ export const createModuleImportModifierPlugin = (
...
+        return moduleResolve(possiblePath, url.pathToFileURL(sourceFilePath))
+          .pathname;
+      } catch {
+        continue;
+      }
+    }
+  }
+  // Failed to resolve the file path
+  return null;
+};
```
Jta26, line 232: Did you intend to use the newly exported functions from the babel plugin here, or is there a special edge case that makes them incompatible?

packages/cli/src/plugins.js
```diff
@@ -116,3 +121,143 @@ export const createModuleImportModifierPlugin = (
...
+  filename.endsWith(`${allowedSuffix}.jsx`) ||
+  filename.endsWith(`${allowedSuffix}.mjs`) ||
+  filename.endsWith(`${allowedSuffix}.cjs`) ||
+  filename.endsWith(allowedSuffix);
+
+const EXTENSIONS = ['.js', '.ts', '.tsx', '.jsx', '.mjs', '.cjs'];
+const filePathResolver = (
+  relativeFilePath: string,
+  sourceFilePath: string,
+  aliases: $ReadOnly<{ [string]: $ReadOnlyArray<string> }>,
```
Jta26, line 202: Are the aliases in here going to have to be something you specify in the `stylex.json` where you have to explicitly tell the CLI where the aliases resolve to?

Jta26 (approved):
Looks good, just a few questions

packages/cli/src/plugins.js
```diff
@@ -116,3 +121,143 @@ export const createModuleImportModifierPlugin = (
...
+            },
+          });
+        },
+      },
+    },
+  };
+};
+
+const matchesFileSuffix = (allowedSuffix: string) => (filename: string) =>
+  filename.endsWith(`${allowedSuffix}.js`) ||
```
mellyeliu, line 190: can we use a loop ie. `EXTENSIONS.some(ext => filename.endsWith(`${allowedSuffix}${ext}`))`

mellyeliu (approved):

packages/cli/src/plugins.js
nmn: The Babel plugin is compiled into a single bundle which makes using the individual exports impossible. I will need to refactor some of these utility functions to a separate package.

For now I just copied the relevant bits.

packages/cli/src/plugins.js
nmn: Yes. This is already the case in the `cli-example`.

nmn: However, there is a PR up right now which discovers aliases automatically. That PR will need to be updated to do that work in the CLI as well.

It might also make sense to move alias rewriting to the main StyleX plugin as it already has all the relevant logic. (This might be the right thing to do actually....)

PR #850, opened 2025-01-13
feat: replace crypto with murmurhash for cli caching

mellyeliu:

`murmurhash` is in fact faster than `crypto`, and we don't need the added security for caching. let's switch for consistency

## Testing

tests pass, yolo

## Pre-flight checklist

### DIFF
--- a/packages/cli/src/cache.js
+++ b/packages/cli/src/cache.js
@@ -8,7 +8,9 @@
  */
 import fs from 'fs/promises';
 import path from 'path';
-import crypto from 'crypto';
+const { utils } = require('@stylexjs/shared');
+
+const hash = utils.hash;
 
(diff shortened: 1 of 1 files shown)


necolas (approved):

packages/cli/src/cache.js
```diff
@@ -8,7 +8,9 @@
  */
 import fs from 'fs/promises';
 import path from 'path';
-import crypto from 'crypto';
+const { utils } = require('@stylexjs/shared');
```
nmn, line 11: Why is this using CommonJS instead of ESM here?

mellyeliu: copied from elsewhere in the repo, will fix in follow ups ...

PR #851, opened 2025-01-13
fix: fix readCache to use custom cache path when set

mellyeliu:

Oversight in https://github.com/facebook/stylex/pull/792 where `readCache` is still reading from default cache path

## Testing

Added more comprehensive testing, previously we were just ensuring the correct cache files were written, not checking the read functionality

## Pre-flight checklist

### DIFF
--- a/packages/cli/src/cache.js
+++ b/packages/cli/src/cache.js
@@ -8,7 +8,7 @@
  */
 import fs from 'fs/promises';
 import path from 'path';
-const { utils } = require('@stylexjs/shared');
+import { utils } from '@stylexjs/shared';
 
 const hash = utils.hash;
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/cache.js
```diff
@@ -8,20 +8,22 @@
  */
 import fs from 'fs/promises';
 import path from 'path';
-import crypto from 'crypto';
+const { utils } = require('@stylexjs/shared');
```
nmn, line 11: Don't use `require`. Use ESM

nmn (changes requested):
Small nit and also failing signal.

nmn (approved):

PR #852, opened 2025-01-13
feat: make filepath relative for cache portability

mellyeliu:

Final follow up to https://github.com/facebook/stylex/pull/790 for cache portability/reusability

## Testing

tests pass~~

### DIFF
--- a/packages/cli/src/cache.js
+++ b/packages/cli/src/cache.js
@@ -18,7 +18,9 @@ export function getDefaultCachePath() {
 }
 
 export async function getCacheFilePath(cachePath, filePath) {
-  const fileName = filePath.replace(/[\\/]/g, '__');
+  const projectRoot = path.resolve();
+  const relativePath = path.relative(projectRoot, filePath);
+  const fileName = relativePath.replace(/[\\/]/g, '__');
   return path.join(cachePath, `${fileName}.json`);
 }
 
--- a/packages/cli/src/cache.js
+++ b/packages/cli/src/cache.js
@@ -17,8 +17,22 @@ export function getDefaultCachePath() {
   return path.join('node_modules', '.stylex-cache');
 }
 
+async function findProjectRoot(startDir = __dirname) {
+  let currentDir = path.resolve(startDir);
+  while (currentDir !== path.parse(currentDir).root) {
+    const packageJsonPath = path.join(currentDir, 'package.json');
+    try {
+      await fs.access(packageJsonPath);
+      return currentDir;
+    } catch (error) {
+      currentDir = path.dirname(currentDir);
+    }
+  }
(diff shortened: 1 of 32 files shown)


nmn (changes requested):
Also a breaking test.

packages/cli/src/cache.js
```diff
@@ -18,7 +18,9 @@ export function getDefaultCachePath() {
 }
 
 export async function getCacheFilePath(cachePath, filePath) {
-  const fileName = filePath.replace(/[\\/]/g, '__');
+  const projectRoot = path.resolve();
```
nmn, line 21: This is equivalent to `__dirname`. I don't think this will always work to give you the projectRoot which is likely what you want.

nmn (changes requested):

packages/cli/src/cache.js
mellyeliu: fixed!

packages/cli/src/cache.js
```diff
@@ -18,7 +18,10 @@ export function getDefaultCachePath() {
 }
 
 export async function getCacheFilePath(cachePath, filePath) {
-  const fileName = filePath.replace(/[\\/]/g, '__');
+  const projectRoot = path.resolve(__dirname, '../../../');
```
nmn, line 21: This doesn't look right. The path should not be hardcoded. You probably need to walk up the tree to find the nearest `node_modules` or `package.json` folder.

nmn (changes requested):
Doesn't quite look right. We can chat further if I'm wrong.

packages/cli/src/cache.js
mellyeliu: I was going after an assumed path that this would be within node_modules/@stylex/ etc path or regardless be _consistent_ across files but you're right it's not guaranteed to be the project root, updated

packages/cli/src/cache.js
```diff
@@ -17,8 +17,25 @@ export function getDefaultCachePath() {
   return path.join('node_modules', '.stylex-cache');
 }
 
+async function findProjectRoot(startDir = __dirname) {
+  let currentDir = path.resolve(startDir);
+  while (currentDir !== path.parse(currentDir).root) {
+    const packageJsonPath = path.join(currentDir, 'package.json');
+    try {
+      await fs.access(packageJsonPath);
```
nmn, line 25: TODO:

Check for `deno.json` and other similar files that may exist. This is fine for now.

nmn (approved):

nmn:
This PR has more changes than just what is expected. Can you please rebase, or change the branch it merges into?

I think you have stacked PRs but this is on the top of the stack and reapplies changes from your other PRs.

PR #810, opened 2024-12-16
Feat: Automatically Discover StyleX Aliases from Configuration Files

p0nch000:
# **[Feature] Automatically Discover StyleX Aliases from Configuration Files**

## Linked Issues
Fixes #765

---

## Overview
This PR enhances the **StyleX Babel plugin** to automatically discover alias configurations from **Node.js native imports** (`package.json`), **`tsconfig.json`**, and **`deno.json`**. It ensures normalized paths for **cross-platform compatibility** and maintains **backward compatibility** with existing manual configurations.

---

## Implementation Details
- **JSON5 Parser**: Handles configuration files with comments or trailing commas.
- **Alias Discovery**:
  - Parses `package.json` for **Node.js native subpath imports** (aliases with `#` prefix).
  - Parses `tsconfig.json` for `compilerOptions.paths` and resolves paths relative to `baseUrl`.
  - Parses `deno.json` for the `imports` field.
- **Merge Configurations**:
  - Combines configurations from all sources:
    **Manual > `package.json` > `tsconfig.json` > `deno.json`**.
- **Code Reuse**:
  - Replaced redundant `findProjectRoot` with `getPackageNameAndPath` for project root discovery.
- **Normalized Paths**:
  - Standardizes path separators for cross-platform compatibility.
- **Error Handling**:
  - Gracefully handles missing or malformed configuration files.

---

## Test Coverage
The following scenarios are tested:

### 1. **Alias Discovery**:
   - **Node.js native imports** from `package.json` (subpath imports).
   - Aliases from `tsconfig.json` paths.
   - Aliases from `deno.json` imports.

### 2. **Configuration Merging**:
   - Merges aliases from all sources with correct priority.

### 3. **Manual Overrides**:
   - Ensures manual alias configurations take precedence over discovered aliases.

### 4. **Error Handling**:
   - Handles missing or invalid configuration files gracefully.

---

### 1. `state-manager.js` (Alias Discovery)
- **Updated Functionality**:
  - `loadAliases`: 
    - Now supports **Node.js native imports** and **deno.json**.
    - Removed custom `stylex.aliases` support.
    - Simplified project root discovery using `getPackageNameAndPath`.
- Added support for **`deno.json` imports**.

### 2. `stylex-transform-alias-config-test.js`
- **Updated Unit Tests**:
  - Tests for Node.js native imports discovery.
  - Tests for `deno.json` imports.
  - Tests for merged configuration from all sources.
  - Tests for proper handling of invalid configurations.

---

[...]

PR #856, opened 2025-01-13
Rebased PR of #810

nmn:

#810 already exists but needs fixes and to be rebased. This is a PR to do that work so it can be merged.

Making a separate PR in case the original PR is updated in the meantime.

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-alias-config-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-alias-config-test.js
@@ -7,7 +7,6 @@
  *
  */
 
-
 'use strict';
 
 import path from 'path';
@@ -59,7 +58,7 @@ describe('StyleX Alias Configuration', () => {
     const manager = new StateManager(state);
 
     expect(manager.options.aliases).toEqual({
-      'components': ['./src/components'],
+      components: ['./src/components'],
       'utils/*': ['./src/utils/*'],
     });
   });
@@ -131,7 +130,7 @@ describe('StyleX Alias Configuration', () => {
     const manager = new StateManager(state);
 
     expect(manager.options.aliases).toEqual({
-      'components': ['./src/components'],
+      components: ['./src/components'],
       '@utils': ['src/utils'],
       '@styles/': ['./src/styles/'],
     });
@@ -149,14 +148,14 @@ describe('StyleX Alias Configuration', () => {
 
     state.opts = {
       aliases: {
-        'components': './custom/path',
+        components: './custom/path',
       },
     };
 
     const manager = new StateManager(state);
 
     expect(manager.options.aliases).toEqual({
-      'components': ['./custom/path'],
+      components: ['./custom/path'],
     });
   });
 
--- a/packages/babel-plugin/__tests__/stylex-transform-alias-config-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-alias-config-test.js
@@ -1,0 +1,158 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
+
+
+'use strict';
+
+import path from 'path';
+import fs from 'fs';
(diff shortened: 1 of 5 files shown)


packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -630,6 +621,141 @@ export default class StateManager {
...
+
+    const pkgInfo = this.getPackageNameAndPath(this.filename);
+    if (!pkgInfo) {
+      return manualAliases ? this.normalizeAliases(manualAliases) : null;
+    }
+
+    const [_packageName, projectDir] = pkgInfo;
+
+    const resolveAliasPaths = (value: string | $ReadOnlyArray<string>) =>
+      (Array.isArray(value) ? value : [value]).map((p) => {
```
mellyeliu, line 644: nit: we should consider moving some of this logic to a separate files utils, there's some path functions in the `cache.js` utils that I also want to extract

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -630,6 +621,141 @@ export default class StateManager {
...
+    } catch (err) {
+      console.warn('Failed to load aliases from deno.json:', err.message);
+    }
+
+    // Merge aliases in priority: manual > package.json > tsconfig.json > deno.json
+    const mergedAliases = {
+      ...denoAliases,
+      ...tsconfigAliases,
+      ...packageAliases,
+      ...(manualAliases || {}),
```
necolas, line 735: Do we still need manualAliases?

nmn: I didn't want to remove it in the same commit, but we could probably remove them in the future once we are sure this feature is working for the wide variety of project types.

We could deprecate it in the next release and delete it entirely after.

packages/babel-plugin/__tests__/stylex-transform-alias-config-test.js
```diff
@@ -0,0 +1,178 @@
...
+ */
+
+'use strict';
+
+import path from 'path';
+import fs from 'fs';
+import os from 'os';
+import StateManager from '../src/utils/state-manager';
+
+describe('StyleX Alias Configuration', () => {
```
necolas, line 17: Please write tests against the public API. If we refactor the internals, tests like this would likely need to be rewritten. And this test doesn't actually validate that real code is correctly processed.

necolas (changes requested):

packages/babel-plugin/__tests__/stylex-transform-alias-config-test.js
nmn: Will do. Will need to simulate a filesystem or similar to make this happen, but should be doable.

necolas:
Are you still working on this @nmn ?

nmn:
I will rebase this PR and rewrite the tests soon. I was travelling for a while and had little time.

packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -725,6 +716,124 @@ export default class StateManager {
...
+        },
+      ],
+    ].map(
+      ([fileName, getConfig]): $ReadOnly<{
+        [string]: string | $ReadOnlyArray<string>,
+      }> => {
+        try {
+          const filePath = path.join(projectDir, fileName);
+          if (fs.existsSync(filePath)) {
+            const rawConfig: mixed = JSON5.parse(
```
necolas, line 783: Why do we need the JSON5 dependency?

nmn: It's just to allow the config file to have comments, trailing commas etc.

It's like `JSONC`, but a more permissive superset.

hungpm372: Okk

KingAmo:
any update on this?

MoOx:
@nmn do you think your work done here could help with #1045 (aliases not working with a Windows setup)

alexandre-ferreiradasilva:
Hi @nmn , are you still working/pushing this forward?

PR #865, opened 2025-01-17
feat: cache invalidation on .babelrc update

mellyeliu:

Let's add cache invalidation on .babelrc config update
- Iteratively search upwards through the file hierarchy until we find the closest .babelrc file
- Hash file and add to cache

## Testing
Added tests writing .babelrc update

## Pre-flight checklist

### DIFF
--- a/packages/cli/src/cache.js
+++ b/packages/cli/src/cache.js
@@ -18,7 +18,9 @@ export function getDefaultCachePath() {
 }
 
 export async function getCacheFilePath(cachePath, filePath) {
-  const fileName = filePath.replace(/[\\/]/g, '__');
+  const projectRoot = path.resolve();
+  const relativePath = path.relative(projectRoot, filePath);
+  const fileName = relativePath.replace(/[\\/]/g, '__');
   return path.join(cachePath, `${fileName}.json`);
 }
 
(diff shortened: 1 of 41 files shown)


packages/cli/src/cache.js
```diff
@@ -17,6 +17,23 @@ export function getDefaultCachePath() {
...
+
+  while (currentDir !== path.parse(currentDir).root) {
+    const babelrcPath = path.join(currentDir, '.babelrc');
+    try {
+      await fs.access(babelrcPath);
+      console.log('Found babelrc:', babelrcPath);
+      return babelrcPath;
+    } catch {
+      currentDir = path.dirname(currentDir);
+    }
```
nmn, line 31: Don't use `fs.access`. Use `fs.exists` instead. That returns a boolean and won't throw an error.

nmn (changes requested):
One nitpick before I merge this.

nmn:
Also needs to be rebased.

packages/cli/src/cache.js
mellyeliu: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback `fs.exists` is deprecated and may cause issues with modern node versions

nmn (approved):

nmn:
This PR is still not ready to merge. It has a bunch of extra commits from other PRs that have already merged. I think it needs to be rebased.

mellyeliu:
this should be safe to merge, iirc when you squash, it commits only the changes seen in the file diff. so even though the commit history of the PR is wonky, Github is aware of the latest state of origin/main.

PR #867, opened 2025-01-18
feat: exported return types of 'stylex.types.*' fns

### LINKED ISSUE #857: TypeScript error while using `stylex.types` - `the inferred type cannot be named without a reference`
### Describe the issue

Hello 👋

`stylex.types` functions are typed to return specific class instances for each of the types - all of these are imported from `VarTypes` file. However, these aren't exported anywhere so TypeScript can't properly typecheck/generate declaration for this. If I didn't use `stylex.types` and just use standard string value - declaration file would look like this:

Image

In cases where I use `stylex.types` there's no type that it can use in place for a given var value - because as mentioned, it's not exported.

Here you can see TS error that's caused by this:

Image

Image

### Expected behavior

Project can be properly typechecked via `tsc` when declaration generation is enabled and `stylex.types` are used on exported member

### Steps to reproduce

1. Install StyleX
2. Enable declaration generating in your `tsconfig` by setting `declaration` to `true`
3. Try to declare and export `stylex.defineVars` from a file
4. Observe error in IDE/`tsc` failing

I've created small reproduction: https://github.com/pawelblaszczyk5/stylex-incorrect-types-repro

### Test case

Enable `declaration` in your `tsconfig` and try this code anywhere:

```tsx

export const example = stylex.defineVars({
  etst: stylex.types.angle("20deg"),
});
```

### Additional comments

I'm pretty sure this could be fixed by just exporting the classes that are used as a return type from `stylex.types` functions

nmn:

Fixes #857

New types are exported such as `stylex.Color` and `stylex.Length`.

These are the return types of the `stylex.types.*` function. So, for example,

- `stylex.types.color("red")` is of type `stylex.Color<string>`

## Question

Should the types be under `stylex.types` instead?

- `stylex.types.color("red")` would be of type `stylex.types.Color<string>`

### DIFF
--- a/packages/stylex/src/VarTypes.js
+++ b/packages/stylex/src/VarTypes.js
@@ -7,12 +7,7 @@
  * @flow strict
  */
 
-export type ValueWithDefault<+T> =
-  | T
-  | $ReadOnly<{
-      default: ValueWithDefault<T>,
-      [string]: ValueWithDefault<T>,
-    }>;
+import type { ValueWithDefault } from './util-types';
 
 export type CSSSyntax =
   | '*'
--- a/packages/stylex/src/stylex.js
+++ b/packages/stylex/src/stylex.js
@@ -21,7 +21,6 @@ import type {
   MapNamespaces,
 } from './StyleXTypes';
 import type {
-  ValueWithDefault,
   Angle,
   Color,
   Url,
@@ -36,22 +35,8 @@ import type {
   TransformFunction,
   TransformList,
 } from './VarTypes';
-export type {
-  ValueWithDefault,
-  Angle,
-  Color,
-  Url,
-  Image,
-  Integer,
-  LengthPercentage,
-  Length,
-  Percentage,
-  Num,
-  Resolution,
-  Time,
-  TransformFunction,
-  TransformList,
-} from './VarTypes';
+import type { ValueWithDefault } from './util-types';
+export * as Types from './VarTypes';
 
 export type {
   VarGroup,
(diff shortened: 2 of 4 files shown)


Jta26:
> Should the types be under stylex.types instead?

I think this is a better organization in terms of usage with `stylex.types` functions, however it's kinda confusing when you have `stylex.StaticStyles`, which is also a "type". So the naming between TS/Flow types and CSS value types is kinda confusing.

This is also kinda awkward, but maybe `valueTypes` is better. This aligns with the actual name for such things.

https://www.w3.org/TR/css3-values/#component-types

nmn:
> This is also kinda awkward, but maybe valueTypes is better. This aligns with the actual name for such things.

I don't know if it's worth renaming the existing public APIs such as `stylex.types.color()` etc. And so it might make sense to keep that name.

`stylex.types` is about CSS types and not TS/Flow types such as `stylex.StyleXStyles`. Can we clarify this with documentation or is it too confusing and something that warrants an API change?

Jta26 (approved):

Jta26:
> Can we clarify this with documentation 

Yeah documentation is probably the best way to disambiguate this.

nmn:
There was no clean way to embed types under `stylex.types` that worked for both Flow and TS, so the types are now under `stylex.Types.*`.

This raises the question, is that the right name? 

Should it be `stylex.VarTypes.*`? Or is `stylex.Types` better because it matches `stylex.types`?

Jta26:
`stylex.Types` is next best imo.

Will the usage look like this?
```
xstyle: stylex.StaticStyles<{
 backgroundColor: stylex.Types.Color, (null | string)
 marginInlineStart: stylex.Types.Length, (null | string | number)
}>
```

nmn:
@Jta26, these types are the return types of `stylex.types.color` etc. So not quite. This is for specialized variable types only. They'd only be needed for theming.

PR #869, opened 2025-01-18
docs: Add theming recipes

nmn:

Adds three more recipes all related to theming.

### 1. Reset Themes

Explains how using `stylex.createTheme(vars, {})` creates a theme that resets variables to their default values.

### 2. Merge Themes

Shows how to define object constants outside `stylex.createTheme` calls to be able to reuse and merge objects and avoid repetition.

### 3. `light-dark`

Talks about using the `light-dark()` function to control the color-scheme instead of defining three `light`, `dark` and `system` themes explicitly.

### DIFF
--- a/apps/docs/docs/learn/06-recipes/03-descendant styles.mdx
+++ b/apps/docs/docs/learn/06-recipes/03-descendant styles.mdx
@@ -6,9 +6,6 @@
 sidebar_position: 3
 ---
 
-import Tabs from '@theme/Tabs';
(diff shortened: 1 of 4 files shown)


necolas:
A theme is actually just a different set of values for a given bag of vars. So you can merge themes, just not themes for the same vars.

nmn:
@necolas Let me take a crack at rewording the recipes so that fact is clearer.

necolas (approved):

PR #884, opened 2025-02-04
Remove 'stylex.include' from public API

necolas:

Removes the `stylex.include` API which is undocumented in OSS and unused at Meta.

## Pre-flight checklist


nmn:
Would it make sense to do a smaller release without the API first?

packages/stylex/src/stylex.js
```diff
@@ -114,138 +72,136 @@ export function attrs(
...
   throw errorForFn('create');
-}
-
-function stylexDefineVars(_styles: $FlowFixMe) {
-  throw errorForFn('defineVars');
-}
+};
 
-const stylexCreateTheme: StyleX$CreateTheme = (_baseTokens, _overrides) => {
+export const createTheme: StyleX$CreateTheme = (_baseTokens, _overrides) => {
```
mellyeliu, line 81: optional, but we could split this refactoring into another PR for clarity

nmn (approved):

packages/stylex/src/stylex.js
```diff
@@ -114,138 +72,136 @@ export function attrs(
...
+    result.style = style;
+  }
+  return result;
+}
 
 export const types = {
-  angle: <T: string | 0 = string | 0>(_v: ValueWithDefault<T>): Angle<T> => {
+  angle: <T: string | 0 = string | 0>(
+    _v: ValueWithDefault<T>,
+  ): Types.Angle<T> => {
```
mellyeliu, line 129: +1 we could make this a separate PR of importing namespace

mellyeliu (approved):

packages/stylex/src/stylex.js
necolas: It was done as a separate commit

PR #886, opened 2025-02-05
Remove default export from 'stylex' module

necolas:

The default export is a legacy API that is undocumented and not intended for use. Meta is still in the process of removing all use of 'stylex()', so the compiler optimizations for that API are left in tact. The legacy API is exported as 'legacyMerge'.

## Pre-flight checklist


nmn:
Can we undo the Babel changes and keep the rest? We can ship that today as we already maintain a bit for a fork for the runtime internally. Otherwise the changes look good.

necolas:
Which babel changes are you referring to?

packages/babel-plugin/src/visitors/stylex-merge.js
```diff
@@ -10,10 +10,10 @@
 import type { NodePath } from '@babel/traverse';
 
 import * as t from '@babel/types';
+import * as babelPathUtils from '../babel-path-utils';
 import StateManager from '../utils/state-manager';
-import stylex from '@stylexjs/stylex';
 import { evaluate } from '../utils/evaluate-path';
-import * as babelPathUtils from '../babel-path-utils';
+import { legacyMerge } from '@stylexjs/stylex';
```
nmn, line 16: My mistake. This is already doing what I was trying to suggest!

nmn (approved):
PR already does what I was asking to do. Ignore my previous comments. Looks good. Let's verify internally before merging though.

PR #887, opened 2025-02-07
Runtime debugging refactor

### LINKED ISSUE #723: Richer debugging information that can power a dev-tools extension
We currently generate a unique className for every `stylex.create` call the identifies the file, constant name and rule name. We should have a more rich system for adding debugging information
that can then be used by a dev-tools extension.

What we would like is a list of all Style Objects applied with fileNames and line numbers, such as:

```html
<div
  data-style-src="package-name:path/to/file:12,8-16-8; package-name:path/to/other:56,4-66,4"
  class="margin-xrjyskp padding-xkgsvji"
>
```

here the `data-style-src` attribute contains a list with the following information:
- The package name where the file can be found. (we should walk up to the nearest `package.json` file to find the `name` of the package)
- The relative path to *within* the package.
- The range where the Style object can be found (Starting line number and column number to the end)

All of this information should provide a future dev-tools extension to pull up the actual source for all the styles that are applied to an element in the order they are applied.

---

## How it would need to work

It should be easy to *find* the needed metadata for this feature, but we would also need to find a way to attach the relevant metadata each compiled style rule.

- We need to be careful that this metadata is not included in the applied styles by `stylex.props`
- The compiler will need to inject the `data-style-src` attribute whenever `{...stylex.props}` is used.

## Possible pitfalls

If the end developer uses `stylex.props` separately from the markup to then apply `className` and `style` manually, the compiler may not be able to reliably set the `data-` attribute.

We should have a separate lint rule that discourages this pattern.

---

## Alt: Richer information in classNames

- We should improve the quality of the debug classNames as well.
- We should investigate if we can generate empty CSS rules for debug classNames with source maps in a way that achieves the same purpose.

necolas:

Add sourceMap information about the file, line, and column responsible for providing styles on a given element. The information populates the 'data-style-src' prop in the DOM, and can be enabled by setting 'debug' to 'true' in the StyleX compiler config.

Removes "debug class names".

## Linked PR/Issues

Fixes #723

## Additional Context

Depends on https://github.com/necolas/styleq/pull/24

## Pre-flight checklist


packages/babel-plugin/__tests__/stylex-transform-create-test.js
```diff
@@ -38,6 +38,37 @@ function transform(source, opts = {}) {
...
+        ),
+      ).toMatchInlineSnapshot(`
+        "import _inject from "@stylexjs/stylex/lib/stylex-inject";
+        var _inject2 = _inject;
+        import stylex from 'stylex';
+        _inject2(".position-x15oojuh{position:fixed;position:sticky}", 3000);
+        export const styles = {
+          foo: {
+            position: "position-x15oojuh",
+            $$css: "npm-package:components/Foo.react.js:3"
```
necolas, line 66: How npm package sources are presented

packages/babel-plugin/__tests__/stylex-transform-stylex-attrs-test.js
```diff
@@ -835,17 +835,11 @@ describe('@stylexjs/babel-plugin', () => {
           import stylex from 'stylex';
           _inject2(".color-x1e2nbdu{color:red}", 3000);
           ({
-            class: "FooBar__styles.default color-x1e2nbdu"
+            class: "color-x1e2nbdu",
+            "data-style-src": "js/FooBar.react.js:3"
```
necolas, line 839: At Meta, I don't think it will add value to display any part of the path before the component name since every file name is unique. In codebases that go the other way, and make heavy use generic component names within folders (e.g., `../Menu/Button/index.js`), the display path might not be long enough to disambiguate from others. So I wonder if it might be necessary to include some kind of configuration for this.

packages/babel-plugin/__tests__/stylex-transform-stylex-attrs-test.js
```diff
@@ -959,10 +946,12 @@ describe('@stylexjs/babel-plugin', () => {
...
           ({
             0: {
-              class: "FooBar__styles.default color-x1e2nbdu"
+              class: "color-x1e2nbdu",
+              "data-style-src": "js/FooBar.react.js:3"
             },
             1: {
-              class: "FooBar__styles.default FooBar__styles.active color-xju2f9n"
+              class: "color-xju2f9n",
+              "data-style-src": "js/FooBar.react.js:3; js/FooBar.react.js:3"
```
necolas, line 954: If you have lots of styles within the same `create` call, and then use them all on the same element (e.g., different conditionals), it produces duplicates at the moment.

necolas: Maybe we'd want the line of the namespace key rather than the `create` call? Some `creates` are pretty large with many keys. If not, add some dev overhead to the runtime to dedupe?

packages/babel-plugin/__tests__/stylex-transform-stylex-attrs-test.js
necolas: I've changed this so that haste modules only display the file name, commonjs displays a shortened path.

packages/babel-plugin/__tests__/stylex-transform-stylex-attrs-test.js
necolas: Latest iteration displays the line number of the individual style objects

packages/babel-plugin/src/utils/add-sourcemap-data.js
```diff
@@ -0,0 +1,114 @@
...
+ * debug string.
+ */
+export function addSourceMapData(
+  obj: CompiledNamespaces,
+  babelPath: NodePath<t.CallExpression>,
+  state: StateManager,
+): CompiledNamespaces {
+  const result: MutableCompiledNamespaces = {};
+  for (const [key, value] of Object.entries(obj)) {
+    // $FlowIgnore (this repo's flow_modules types for babel are incomplete)
```
mellyeliu, line 67: we should open a issue to track this, flow has been frustrating...

mellyeliu (approved):

PR #890, opened 2025-02-13
Update API configuration docs

necolas:

Adds docs for the postcss-plugin and updates the docs for the babel-plugin to include `debug`. Also remove deprecated features from these docs.

## Pre-flight checklist

### DIFF
--- a/apps/docs/docs/api/configuration/babel-plugin.mdx
+++ b/apps/docs/docs/api/configuration/babel-plugin.mdx
@@ -10,6 +10,38 @@ sidebar_position: 1
 
 ## Configuration options
 
+### `aliases`
+
+```ts
+aliases: {[key: string]: string | Array<string>} // Default: undefined
+```
+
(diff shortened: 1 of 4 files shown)


apps/docs/docs/api/configuration/babel-plugin.mdx
```diff
@@ -10,106 +10,112 @@ sidebar_position: 1
...
+### `dev`
 
 ```ts
-runtimeInjection: boolean // Default: the value of `dev`
+dev: boolean // Default: false
 ```
 
-Should StyleX generate code to inject styles at runtime?
-This may be useful during development but should be disabled in production.
+When `true`, StyleX will insert debug class names to identify the source of the styles.
```
nmn, line 51: `dev` also insert function calls to inject the styles at runtime making it possible to use StyleX without setting up CSS file extraction.

Additionally, it makes hot reloading more usable.

nmn (approved):
Looks good! Thanks for adding this.

PR #897, opened 2025-02-20
feat: decouple debug class names and data-* attributes

mellyeliu:
## Context

As discussed, let's decouple debug class names and `data-stylex-src` attributes by creating a new config option `debugClassNamePrefix` to create more optionality on using `data-stylex-src` without debug classnames

The only thing worth considering is we now have chaining that looks like this: debugClassNamePrefix falls back to debug, which falls back to dev, which feels a bit opaque and potentially confusing.

## Testing
Tests pass?

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/src/utils/state-manager.js
+++ b/packages/babel-plugin/src/utils/state-manager.js
@@ -70,6 +70,7 @@ const CheckModuleResolution: Check<ModuleResolution> = z.unionOf3(
 export type StyleXOptions = $ReadOnly<{
   ...RuntimeOptions,
   aliases?: ?$ReadOnly<{ [string]: string | $ReadOnlyArray<string> }>,
+  enableDebugClassNames?: boolean,
   genConditionalClasses: boolean,
   importSources: $ReadOnlyArray<
     string | $ReadOnly<{ from: string, as: string }>,
@@ -162,6 +163,14 @@ export default class StateManager {
       false,
       'options.debug',
     );
+
+    const enableDebugClassNames: StyleXStateOptions['enableDebugClassNames'] =
+      z.logAndDefault(
+        z.boolean(),
+        options.enableDebugClassNames ?? true,
+        true,
+        'options.enableDebugClassNames',
+      );
 
     const test: StyleXStateOptions['test'] = z.logAndDefault(
       z.boolean(),
@@ -282,6 +291,7 @@ export default class StateManager {
       debug,
       definedStylexCSSVariables: {},
       dev,
+      enableDebugClassNames,
       genConditionalClasses,
       importSources,
       rewriteAliases:
--- a/packages/shared/src/convert-to-className.js
+++ b/packages/shared/src/convert-to-className.js
@@ -29,7 +29,11 @@ export function convertStyleToClassName(
   atRules: $ReadOnlyArray<string>,
   options: StyleXOptions = defaultOptions,
 ): StyleRule {
-  const { classNamePrefix = 'x', debug = false } = options;
+  const {
+    classNamePrefix = 'x',
+    debug = false,
+    enableDebugClassNames = true,
+  } = options;
   const [key, rawValue] = objEntry;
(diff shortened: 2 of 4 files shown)


necolas (changes requested):
This flag is only needed because we have so many difficulties rolling things out internally. So I think this flag should only disable debug class names if it is active and `debug` is true, rather than needing to be turned on/off separately. Let's call the flag something like `enableDebugClassNames` and have it set to `true` by default

mellyeliu:
makes sense! refactored so dev class names are enabled when `debug` and `enableDebugClassNames` are true, and off otherwise. this should unblock us from syncing https://github.com/facebook/stylex/pull/887. we'll have to set the internal config value of `enableDebugClassNames` to false when syncing

packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -163,6 +164,14 @@ export default class StateManager {
...
     );
 
+    const enableDebugClassNames: StyleXStateOptions['enableDebugClassNames'] =
+      z.logAndDefault(
+        z.boolean(),
+        options.enableDebugClassNames ?? true,
+        true,
+        'options.enableDebugClassNames',
+      );
+
```
nmn, line 174: This option should default to the value of `debug`.

Also, should we call this: `debugClassPrefix` maybe?

```suggestion
    const enableDebugClassNames: StyleXStateOptions['enableDebugClassNames'] =
      z.logAndDefault(
        z.boolean(),
        options.enableDebugClassNames ?? debug,
        true,
        'options.enableDebugClassNames',
      );

```

packages/shared/src/convert-to-className.js
```diff
@@ -29,7 +29,11 @@ export function convertStyleToClassName(
   atRules: $ReadOnlyArray<string>,
   options: StyleXOptions = defaultOptions,
 ): StyleRule {
-  const { classNamePrefix = 'x', debug = false } = options;
+  const {
+    classNamePrefix = 'x',
+    debug = false,
+    enableDebugClassNames = true,
+  } = options;
```
nmn, line 36:
```suggestion
  const {
    classNamePrefix = 'x',
    enableDebugClassNames = false,
  } = options;
```

packages/shared/src/convert-to-className.js
```diff
@@ -58,9 +62,10 @@ export function convertStyleToClassName(
 
   // NOTE: '<>' is used to keep existing hashes stable.
   // This should be removed in a future version.
-  const className = debug
-    ? `${key}-${classNamePrefix}${createHash('<>' + stringToHash)}`
-    : classNamePrefix + createHash('<>' + stringToHash);
+  const className =
+    debug && enableDebugClassNames
+      ? `${key}-${classNamePrefix}${createHash('<>' + stringToHash)}`
+      : classNamePrefix + createHash('<>' + stringToHash);
```
nmn, line 68:
```suggestion
  const className =
    enableDebugClassNames
      ? `${key}-${classNamePrefix}${createHash('<>' + stringToHash)}`
      : classNamePrefix + createHash('<>' + stringToHash);
```

packages/shared/src/utils/default-options.js
```diff
@@ -12,6 +12,7 @@ import type { StyleXOptions } from '../common-types';
 export const defaultOptions: StyleXOptions = {
   dev: false,
   debug: false,
+  enableDebugClassNames: true,
```
nmn, line 15:
```suggestion
  enableDebugClassNames: false,
```

nmn (changes requested):
Let's not require `debug` to be able to turn on `enableDebugClassnames`

nmn (approved):
I see the comment by @necolas and I'm  ok enough with this approach as well, so approved!

packages/babel-plugin/src/utils/state-manager.js
mellyeliu: previously had this approach re: Nicolas' comment above

PR #898, opened 2025-02-22
feat: create media query transformer to ensure last query wins

mellyeliu:
create media query transformer to ensure “last query wins” logic

## Context  
Now that we have a usable media query parser in https://github.com/facebook/stylex/pull/743, we can finally implement last media query wins logic. 

## Implementation
High level approach: traverse through styles, search for media queries, do two passes at each node - one to build up a list of negations for each query, and another to apply the negation. 

1. traverse nested styles with dfs
   - dfs to find all nested media queries
   - extract and flatten all media queries per into list for easier transformation

2. iterate media queries in reverse order
   - loop through the extracted media queries from last to first 

3. accumulate negations
   -backwards, for single pass
-  build up a list of negations from previous queries during reverse iteration
   - apply these negations to prevent earlier media queries from overriding more specific ones defined later

4. construct updated media queries
   - forwards order, because we need to reassign the object keys in the same order we see them to avoid breaking js object keys order
   - for each media query: use `MediaQuery.parser.parseToEnd` to convert query string into object representation
 - append all accumulated negations to queries
   - stringify media query using`MediaQuery.toString()`
   - append current queries to negations

5. return updated styles object with negated media queries

## Testing

Use `json.stringify` to ensure object key order is maintained

Scenarios like:
- basic usage: multiple widths 
- basic usage: complex styles object 
- basic usage: lots and lots of widths 
- nested queries
- handles comma-separated (or) media queries
- handles and media queries 
- keywords
- combination of keywords and > rules

Missing: 
- Nested not cases blocked on parsing issues

## Pre-flight checklist

### DIFF
--- a/packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-transform-test.js
+++ b/packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-transform-test.js
@@ -1,0 +1,339 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
+
+import { lastMediaQueryWinsTransform } from '../media-query-transform.js';
+
+const stylex = {
+  create: (styles) => styles,
+};
+
+describe('Media Query Transformer', () => {
+  test('basic usage: multiple widths', () => {
+    const originalStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (max-width: 1440px)': '1 / 4',
+          '@media (max-width: 1024px)': '1 / 3',
+          '@media (max-width: 768px)': '1 / -1',
+        },
+      },
+    };
+
+    const expectedStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (max-width: 1440px) and (not (max-width: 1024px)) and (not (max-width: 768px))':
+            '1 / 4',
+          '@media (max-width: 1024px) and (not (max-width: 768px))': '1 / 3',
+          '@media (max-width: 768px)': '1 / -1',
+        },
+      },
+    };
+
+    const result = lastMediaQueryWinsTransform(originalStyles);
+    expect(JSON.stringify(result)).toBe(JSON.stringify(expectedStyles));
+  });
+
+  test('basic usage: nested query', () => {
+    const originalStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (max-width: 1440px)': {
+            '@media (max-width: 1024px)': '1 / 3',
+            '@media (max-width: 768px)': '1 / -1',
+          },
+          '@media (max-width: 1024px)': '1 / 3',
+          '@media (max-width: 768px)': '1 / -1',
+        },
+      },
+    };
+
+    const expectedStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (max-width: 1440px) and (not (max-width: 1024px)) and (not (max-width: 768px))':
+            {
+              '@media (max-width: 1024px) and (not (max-width: 768px))':
+                '1 / 3',
+              '@media (max-width: 768px)': '1 / -1',
+            },
+          '@media (max-width: 1024px) and (not (max-width: 768px))': '1 / 3',
+          '@media (max-width: 768px)': '1 / -1',
+        },
+      },
+    };
+
+    const result = lastMediaQueryWinsTransform(originalStyles);
+    expect(JSON.stringify(result)).toBe(JSON.stringify(expectedStyles));
+  });
+
+  test('basic usage: nested query', () => {
+    const originalStyles = {
+      table: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (max-width: 1440px)': {
+            '@media (max-width: 1024px)': '1 / 3',
+            '@media (max-width: 768px)': '1 / -1',
+          },
+          '@media (max-width: 1024px)': '1 / 3',
+          '@media (max-width: 768px)': '1 / -1',
+        },
+        padding: '10px',
+      },
+    };
+
+    const expectedStyles = {
+      table: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (max-width: 1440px) and (not (max-width: 1024px)) and (not (max-width: 768px))':
+            {
+              '@media (max-width: 1024px) and (not (max-width: 768px))':
+                '1 / 3',
+              '@media (max-width: 768px)': '1 / -1',
+            },
+          '@media (max-width: 1024px) and (not (max-width: 768px))': '1 / 3',
+          '@media (max-width: 768px)': '1 / -1',
+        },
+        padding: '10px',
+      },
+    };
+
+    const result = lastMediaQueryWinsTransform(originalStyles);
+
+    expect(JSON.stringify(result)).toBe(JSON.stringify(expectedStyles));
+  });
+
+  test('basic usage: complex object', () => {
+    const originalStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (max-width: 1440px)': '1 / 4',
+          '@media (max-width: 1024px)': '1 / 3',
+          '@media (max-width: 768px)': '1 / -1',
+        },
+        grid: {
+          default: '1 / 2',
+          '@media (max-width: 1440px)': '1 / 4',
+        },
+        gridRow: {
+          default: '1 / 2',
+          padding: '10px',
+        },
+      },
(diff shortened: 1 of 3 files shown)


packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-transform-test.js
```diff
@@ -0,0 +1,217 @@
...
+        '@media (max-width: 1440px)': '1 / 4',
+        '@media (max-width: 1024px)': '1 / 3',
+        '@media (max-width: 768px)': '1 / -1',
+      },
+    };
+
+    const expectedStyles = {
+      gridColumn: {
+        default: '1 / 2',
+        '@media (max-width: 1440px) and (not (max-width: 1024px)) and (not (max-width: 768px))':
```
necolas, line 30: We should be concerned at how bloated generated code gets with this approach

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-transform-test.js
```diff
@@ -0,0 +1,217 @@
...
+          '1 / 4',
+        '@media (max-width: 1024px) and (not (max-width: 768px))': '1 / 3',
+        '@media (max-width: 768px)': '1 / -1',
+      },
+      grid: {
+        default: '1 / 2',
+        '@media (max-width: 1440px)': '1 / 4',
```
necolas, line 69: This is an example of where I think this approach breaks down. Let's say some conditional applies this MQ to the same element as the MQs above - non of the negations apply and you go back to having non deterministic rendering results

necolas (commented):
What does "last wins" mean exactly in this context? FWIW, the "last wins" approach to property merging in StyleX is only true for properties with the same name. The way we handle short/long form merging (inc within a single style object) is really problematic and something we're trying to undo. So if last-wins for MQ is a local concern within a style, I'm concerned that will also be problematic. Anything that relies on order of keys in authored code is going to have those problems

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-transform-test.js
nmn: This is a bit of a work in progress where we will be simplifying the media queries as well.

mellyeliu: the doc here provides a bit more context - https://docs.google.com/document/d/1KlMnWz__zVBNcH_HQmd64sPEFf6AdaXeG46rL_eU5ec/edit?tab=t.uwqdsvafq1l7#heading=h.m5zrpmurwti 

the plan is to eventually rewrite the media query parser to do the simplification step for us, or to pass this return to another fn to do a second round of parsing

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-transform-test.js
nmn: I think there's some misunderstanding of how the media queries work and what this PR is fixing.

All that this PR does, is that if you have multiple "conditions" in a single value (`{default: A, ...}`) then you no longer have to manually ensure that the conditions don't overlap and we do that for you automatically.

This removes the main source of non-determinism which is that we cannot guarantee that the media queries in the generated CSS will be in the same order as defined in the source.

When talking about media queries *within* multiple keys, such as a shorthand and a longhand, there is no non-determinism there. Based on the config options, either the "the last applied property" (`application-order`) will win, OR, the longhand property will win (`property-specificity`). 

Whichever "value" wins, this PR is only about making the "conditions" within the "value" deterministic and predictable.

nmn: To further clarify, the generated CSS order is:
```
<shorthands>
<Media Queries around shorthands>
<Pseudo Classes on shorthands>
<Media Queries and Pseudo Classes around shorthands>

<longhands>
<Media Queries around longhands>
<Pseudo Classes on longhands>
<Media Queries and Pseudo Classes around longhands>
```

i.e., the longhand property *always* has higher specificity regardless of the media query applied on the shorthand and this makes the styles deterministic.

mellyeliu: Hmm, maybe my example in the test cases is not be the most illustrative in how it's used in stylex. 

The object containing `default` and MQs is effectively a single value assigned to a style property, not a nested style declaration.

Here's an example of how this would resolve in stylex
```tsx
  foo: {
    paddingTop: {
      default: 50,
      '@media (max-width: 1440px)': 10px,
      '@media (max-width: 768px)': 15px,
    }
  },
  foo2: {
    padding: {
      default: 500,
      '@media (max-width: 1280px)': 20px, 
      '@media (max-width: 480px)': '40px',
    }
  }
});
```

Which would be used in an element like:
```tsx
// application-order 
stylex.props(styles.foo, styles.foo2) 
// is equivalent to stylex.props(styles.foo2)
// because padding overrides padding-top in application order


// property-specificity 
stylex.props(styles.foo, styles.foo2) 
// which would have padding-top from foo
// and padding from foo2 in the other directions


stylex.props(styles.foo2, styles.foo) 
// for both config options, element has padding-top from foo
// and padding from foo2 in the other directions
```

necolas: Ah yes that's right. Thanks

mellyeliu:
updated implementation to ensure object key ordering is preserved

mellyeliu:
ignore legacy media query syntax
ping @nmn pls review 👻

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-transform-test.js
```diff
@@ -0,0 +1,339 @@
...
+          '@media (max-width: 500px)': '60%',
+        },
+      },
+    };
+
+    const result = lastMediaQueryWinsTransform(originalStyles);
+    expect(JSON.stringify(result)).toBe(JSON.stringify(expectedStyles));
+  });
+
+  test.skip('handles and media queries', () => {
```
nmn, line 244: Why is this one skipped? An issue with the parser?

packages/style-value-parser/src/at-queries-from-tokens/media-query-transform.js
```diff
@@ -0,0 +1,111 @@
...
+
+    for (let i = mediaKeys.length - 1; i > 0; i--) {
+      // Skip last iteration
+      const mediaQuery = MediaQuery.parser.parseToEnd(mediaKeys[i]);
+      negations.push(mediaQuery);
+      accumulatedNegations.push([...negations]); // Clone array before pushing
+    }
+    accumulatedNegations.reverse();
+    accumulatedNegations.push([]);
+    console.log('accumulatedNegations', accumulatedNegations);
```
nmn, line 85: remove console.log

packages/style-value-parser/src/at-queries-from-tokens/media-query-transform.js
```diff
@@ -0,0 +1,111 @@
...
+
+    for (let i = 0; i < mediaKeys.length; i++) {
+      const currentKey = mediaKeys[i];
+      const currentValue = result[currentKey];
+
+      const baseMediaQuery = MediaQuery.parser.parseToEnd(currentKey);
+      const reversedNegations = [...accumulatedNegations[i]].reverse();
+      console.log('start');
+      console.dir(baseMediaQuery, { depth: null });
+      console.dir(reversedNegations, { depth: null });
```
nmn, line 95: remove

packages/style-value-parser/src/at-queries-from-tokens/media-query-transform.js
```diff
@@ -0,0 +1,111 @@
...
+
+      const combinedQuery = combineMediaQueryWithNegations(
+        baseMediaQuery,
+        reversedNegations,
+      );
+      console.dir(combinedQuery, { depth: null });
+
+      const newMediaKey = combinedQuery.toString();
+
+      delete result[currentKey];
```
nmn, line 105: Refactor this function a bit to improve readability and to not need to rely on key deletion.

nmn (approved):
Please remove the console.logs at least before merging.

Further code improvements can be made later.

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-transform-test.js
mellyeliu: yeppp see edge cases in https://github.com/facebook/stylex/pull/943

packages/style-value-parser/src/at-queries-from-tokens/media-query-transform.js
mellyeliu: i'll revisit this in the next PR when integrating with rules

PR #901, opened 2025-02-25
Simplify and add to StateManager unit tests

necolas:

Adding some missing tests and simplified how the StateManager tests are written (no logic, loops) before we add more feature flags.


necolas:
This PR is a dependency of https://github.com/facebook/stylex/pull/902

mellyeliu (approved):

PR #902, opened 2025-02-26
Add enableDevClassNames and enableDebugDataProp flags

necolas:

More flags for rolling out newer debugging features internally.

### DIFF
--- a/packages/shared/__tests__/convert-to-className-test.js
+++ b/packages/shared/__tests__/convert-to-className-test.js
@@ -31,7 +31,22 @@ describe('convert-to-className test', () => {
     const className = result[1];
     expect(className.startsWith('margin-x')).toBe(true);
   });
-  test('prefixes classname with prefer only when options.debug is false', () => {
+  test('prefixes classname with prefix only when options.enableDebugClassNames is false', () => {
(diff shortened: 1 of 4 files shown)


mellyeliu (approved):

packages/shared/__tests__/convert-to-className-test.js
```diff
@@ -31,7 +31,22 @@ describe('convert-to-className test', () => {
     const className = result[1];
     expect(className.startsWith('margin-x')).toBe(true);
   });
-  test('prefixes classname with prefer only when options.debug is false', () => {
+  test('prefixes classname with prefix only when options.enableDebugClassNames is false', () => {
```
mellyeliu, line 34: can we also add a test for when `debug` is true and `enableDebugClassNames` false?

necolas: Oops that was meant to be this one. I forgot to change the debug value after copy pasting

mellyeliu (approved):

necolas:
Merged to main in b1f1f7011e0c43bcde5a9659a4f109a3b5e7fa7d

PR #908, opened 2025-02-27
Improve monorepo task workflows

### LINKED ISSUE #878: GitHub Workflow for Typos has false positives
### Describe the issue

The typos workflow incorrectly produces errors about generated or snapshot object keys containing typos.

### Expected behavior

No incorrect CI failures caused by the typos workflow.

### Steps to reproduce

Put up a diff with test snapshots containing hashed object keys.

### Test case

https://github.com/facebook/stylex/pull/875

### Additional comments

_No response_

necolas:
* Faster task and CI builds by excluding 'app' workspace builds.
* Faster code formatting by excluding large fixtures.
* Automatic precommit checks for format and lint errors.
* `npm test` runs the same tests (locally) as the CI workflow.
* Minor reorganization of workflow jobs.

Fixes #878


necolas:
The "size" benchmark only fails because the base commit doesn't have the new benchmarks package. If you inspect the job you can see that the "patch" step ran successfully. After this lands, future PRs will collect benchmark data correctly

mellyeliu (approved):
Tested the precommit hooks locally and looks great, lint/prettier/flow process much simplified

PR #913, opened 2025-02-28
Rename packages in "app" workspace

necolas:
Rename app packages

See https://github.com/facebook/stylex/discussions/912


apps/docs/package.json
```diff
@@ -24,7 +24,6 @@
     "@fortawesome/react-fontawesome": "^0.2.2",
     "@mdx-js/react": "^1.6.22",
     "@stylexjs/stylex": "0.10.1",
-    "@vercel/analytics": "^1.1.1",
```
necolas, line 27: Fixes incompatible versions of React during docs install

apps/example-nextjs/.eslintrc.js
```diff
@@ -8,7 +8,6 @@
  */
 
 module.exports = {
-  extends: 'next/core-web-vitals',
```
necolas, line 11: Fixes monorepo linter crash over missing parser for this package

mellyeliu:
I was playing around with something like this yesterday, and I couldn't get the vercel build to deploy correctly on a fresh install + package-lock generation. I think the reason is that root level package.json has some issues and it causes a corrupted package-lock.json file, i fixed it yesterday by pasting the package-lock from main and rerunning `npm i` off of that instead but just tested and the build is still failing..

apps/docs/package.json
nmn: Yeah, We can bring this back once I manage to upgrade Docusaurus.

nmn (approved):
Everything looks good.

PR #922, opened 2025-03-02
Fix style compilation for string and number keys

### LINKED ISSUE #919: styles with numeric or kebab-case keys does not work anymore
### Describe the issue

styles with numeric or kebab-case keys does not work anymore since `"@stylexjs/postcss-plugin": "^0.11.0"`:

### Expected behavior

Styles should work.

### Steps to reproduce

* Clone https://github.com/hrasoa/stylex
* `pnpm dev`
* load the page, styles are not working and console output errors:
```
Style node path not found for key: 1
Style node path not found for key: color-red
Style node path not found for key: 1
Style node path not found for key: color-red
```

### Test case

_No response_

### Additional comments

_No response_

necolas:

Styles with string or number keys have different types in the Babel AST. This patch correctly generates sourcemap data for those key types. And it ensures that the compiled style is always preserved even if sourcemap data isn't generated.

Fix #919
Close #920

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-create-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-create-test.js
@@ -45,7 +45,13 @@ describe('@stylexjs/babel-plugin', () => {
           import stylex from 'stylex';
           export const styles = stylex.create({
             foo: {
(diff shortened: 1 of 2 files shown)


nmn (approved):
LGTM

PR #935, opened 2025-03-04
chore: Add more vercel config to the json file

nmn:

Move Vercel configuration to the `vercel.json` instead of the Vercel UI.


necolas (approved):
LGTM

PR #936, opened 2025-03-06
fix: mismatch in debug classnames when `enableDebugClassNames` is false

mellyeliu:
Added boolean check for `enableDebugClassNames` in `stylex-define-vars.js` and `evaluate-path.js` that was missed in https://github.com/facebook/stylex/pull/742

## Tests
Added tests in `stylex-define-vars-test` for debug mode, with `enableDebugClassNames` turned on and off to cover this change

Also added tests in `stylex-transform` for debug mode, with `enableDebugClassNames` turned on and off for https://github.com/facebook/stylex/pull/742
- stylex-transform-call-test
- stylex-transform-stylex-attrs-test
- stylex-transform-stylex-props-test

## Pre-flight checklist


packages/shared/__tests__/define-vars/stylex-define-vars-test.js
```diff
@@ -261,6 +261,82 @@ describe('stylex-define-vars test', () => {
...
+      __themeName__: classNamePrefix + createHash(themeName),
+      bgColor: `var(--${classNamePrefix + createHash(`${themeName}.bgColor`)})`,
+      bgColorDisabled: `var(--${
+        classNamePrefix + createHash(`${themeName}.bgColorDisabled`)
+      })`,
+      cornerRadius: `var(--${
+        classNamePrefix + createHash(`${themeName}.cornerRadius`)
+      })`,
+      fgColor: `var(--${classNamePrefix + createHash(`${themeName}.fgColor`)})`,
+    });
```
necolas, line 307: Do you think we could use an inline snapshot here instead of reproducing implementation details?

packages/babel-plugin/__tests__/stylex-transform-call-test.js
```diff
@@ -2057,6 +2057,46 @@ describe('@stylexjs/babel-plugin', () => {
...
+              marginRight: 12,
+            },
+            content: {
+              gridArea: 'content',
+            },
+            root: {
+              display: 'grid',
+              gridTemplateRows: '100%',
+              gridTemplateAreas: '"content"',
+            },
```
nmn, line 2076: Why are the examples `gridTemplateAreas`?

packages/babel-plugin/__tests__/stylex-transform-call-test.js
```diff
@@ -2057,6 +2057,46 @@ describe('@stylexjs/babel-plugin', () => {
...
+      ).toMatchInlineSnapshot(`
+        "import _inject from "@stylexjs/stylex/lib/stylex-inject";
+        var _inject2 = _inject;
+        import * as stylex from '@stylexjs/stylex';
+        _inject2(".x1wsuqlk{margin-right:12px}", 4000);
+        _inject2(".x1fdo2jl{grid-area:content}", 1000);
+        _inject2(".xrvj5dj{display:grid}", 3000);
+        _inject2(".x7k18q3{grid-template-rows:100%}", 3000);
+        _inject2(".x5gp9wm{grid-template-areas:\\"content\\"}", 2000);
+        const styles = {};
```
nmn, line 2096: This doesn't look right.

`styles` should not be an empty object.

packages/babel-plugin/src/utils/evaluate-path.js
```diff
@@ -173,9 +173,11 @@ function evaluateThemeRef(
         : utils.genFileBasedIdentifier({ fileName, exportName, key });
 
     const debug = state.traversalState.options.debug;
+    const enableDebugClassNames =
+      state.traversalState.options.enableDebugClassNames;
 
     const varName =
-      debug === true
+      debug && enableDebugClassNames
```
nmn, line 180: I again think that `enableDebugClassNames` should default to the value of `debug` so we don't have to be careful about reading both values in all places that use it...

nmn (commented):
I see an issue with the first test, but I don't see anything wrong with the code changes...

packages/babel-plugin/__tests__/stylex-transform-call-test.js
mellyeliu: copied over a test that doesn't have much signal, updated

packages/babel-plugin/src/utils/evaluate-path.js
```diff
@@ -173,9 +173,11 @@ function evaluateThemeRef(
         : utils.genFileBasedIdentifier({ fileName, exportName, key });
 
     const debug = state.traversalState.options.debug;
+    const enableDebugClassNames =
+      state.traversalState.options.enableDebugClassNames;
```
nmn, line 177:
```suggestion
    const {debug, enableDebugClassNames}  = state.traversalState.options;
```

packages/shared/__tests__/define-vars/stylex-define-vars-test.js
nmn: This test was written this way to verify that the _correct_ hash was generated.
In the past there was a bug where two different parts of the codebase ended up generating distinct hashes.

nmn (approved):

packages/babel-plugin/__tests__/stylex-transform-call-test.js
mellyeliu: it's because this is adding tests to the edge cases section with invalid styles, these tests are low signal, will just remove

PR #939, opened 2025-03-06
Move 'docs' to 'packages'

necolas:

Moved 'docs' into packages


mellyeliu (approved):
looks clean!

package.json
```diff
@@ -11,7 +11,7 @@
     "prettier": "prettier --write \"**/*.js\" \"**/*.ts\"",
     "prettier:report": "prettier --check \"**/*.js\"",
     "lint": "npm run lint:report -- --fix",
-    "lint:report": "eslint packages apps/docs apps/example-esbuild apps/example-rollup apps/example-webpack",
+    "lint:report": "eslint",
     "test:packages": "npm run test --workspaces --if-present",
     "test": "npm run build && npm run flow && npm run test:packages && npm run prettier:report && npm run lint:report"
   },
```
nmn, line 17: The `workspaces` field needs to be updated to ensure that `docs` isn't built before its dependencies. This is a limitation of how NPM monorepos work.

Adding `packages/babel-plugin` before `packages/*` should resolve this.

nmn (changes requested):
One small change needed for this to be reliable.

package.json
necolas: Good spot. Resolved

PR #940, opened 2025-03-07
fix: Escape special characters from debug var names

### LINKED ISSUE #937: [babel-plugin] in debug, styles values with dot are unsupported
### Describe the issue

Image

### Expected behavior

It should work.

### Steps to reproduce

create styles like:

```ts
  [1.5]: {
    paddingLeft: '1px',
  }
})
```
enable `debug` in the babel config.

### Test case

_No response_

### Additional comments

_No response_

nmn:

Fixes #937 

In `debug` mode we add prefixes to the generate variable names to make them more readable. The prefixes use the object keys that you use in your `stylex.defineVars` calls.

This PR escapes special characters (etc) to ensure that the generated variable names stay valid. Even if you use special characters in your `stylex.defineVars` calls.

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-define-vars-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-define-vars-test.js
@@ -317,6 +317,41 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
 
+    test('transforms variables object with numeric keys and add stylex.inject in debug mode', () => {
+      expect(
+        transform(
+          `
+          import stylex from 'stylex';
+          export const buttonTheme = stylex.defineVars({
+            '10': {
+              default: 'blue',
+              '@media (prefers-color-scheme: dark)': 'lightblue',
+              '@media print': 'white',
+            },
+            '1.5 pixels': {
+              default: 'grey',
+              '@media (prefers-color-scheme: dark)': 'rgba(0, 0, 0, 0.8)',
+            },
+            'corner#radius': 10,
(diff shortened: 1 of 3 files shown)


packages/babel-plugin/__tests__/stylex-transform-define-vars-test.js
```diff
@@ -317,6 +317,41 @@ describe('@stylexjs/babel-plugin', () => {
...
+        transform(
+          `
+          import stylex from 'stylex';
+          export const buttonTheme = stylex.defineVars({
+            '10': {
+              default: 'blue',
+              '@media (prefers-color-scheme: dark)': 'lightblue',
+              '@media print': 'white',
+            },
+            '1.5 pixels': {
```
necolas, line 331: I think we'd be better off not supporting stupid key names

nmn: To be fair, these are not supposed to be the CSS variables themselves. These are just JS references to CSS variables whose names are auto-generated. As this PR shows, we can easily escape the names for debug names.

We allow arbitrary strings in other StyleX APIs like `stylex.create`. Adding an arbitrary limitation on what you can use in the object references while creating CSS variables doesn't seem to help and makes the API more inconsistent.

necolas: One is a key that doesn't get turned into any CSS code, the other is not.

necolas (approved):

PR #942, opened 2025-03-11
perf: Fix edge-case slowdown in compilation

nmn:

Identified and fixed a bad edge-case where using certain patterns of constants would take an unreasonable amount of time to compile.

Created a fixture with the problematic code pattern, and added a new jest to to verify that it compiles.

Before the fix, the test that transforms this one file took ~100s on an M1 Max. 

After the fix, it dropped to ~1s.

i.e. That is a 100x perf improvement for this particular pattern of code.

### TLDR;

Object evaluation wasn't being cached properly.


packages/babel-plugin/src/utils/evaluate-path.js
```diff
@@ -519,6 +519,7 @@ function _evaluate(path: NodePath<>, state: State): any {
     const elems: $ReadOnlyArray<NodePath<>> = arrPath.get('elements');
     for (const elem of elems) {
       const elemValue = evaluate(elem, state.traversalState, state.functions);
+      // const elemValue = evaluateCached(elem, state);
```
necolas, line 522: Remove

packages/babel-plugin/src/utils/evaluate-path.js
```diff
@@ -551,13 +552,19 @@ function _evaluate(path: NodePath<>, state: State): any {
...
             reason: deoptReason,
             value,
-          } = evaluate(keyPath, state.traversalState, state.functions);
+          } = evaluate(
+            keyPath,
+            state.traversalState,
+            state.functions,
+            state.seen,
+          );
+          // evaluateCached(keyPath, state);
```
necolas, line 567: Remove

packages/babel-plugin/src/utils/evaluate-path.js
```diff
@@ -788,6 +800,7 @@ function _evaluate(path: NodePath<>, state: State): any {
           state.traversalState,
           state.functions,
         );
+        // const parsedObj = evaluateCached(object, state);
```
necolas, line 803: Remove

packages/babel-plugin/__tests__/stylex-perf-create-theme-test.js
```diff
@@ -0,0 +1,48 @@
...
+      [stylexPlugin, { ...defaultOpts, ...opts }],
+    ],
+  });
+  return { code: result.code, styles: result.metadata.stylex };
+}
+
+const themeFile = path.resolve(__dirname, '../__fixtures__/colorThemes.js');
+
+describe('create theme', () => {
+  it('should transform complex theme file', () => {
```
necolas, line 43:
```suggestion
  test('transform complex theme file', () => {
```

packages/babel-plugin/__tests__/stylex-perf-create-theme-test.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+'use strict';
+
+jest.autoMockOff();
+
+const { transformFileSync } = require('@babel/core');
+const stylexPlugin = require('../src/index');
+const path = require('path');
+// const rootDir = path.resolve(__dirname, './__fixtures__');
```
necolas, line 17: Remove

packages/babel-plugin/__fixtures__/colorThemes.js
```diff
@@ -0,0 +1,760 @@
...
+  // coolGrayAlpha05: 'rgba(17, 27, 33, 0.05)',
+  // coolGrayAlpha10: 'rgba(17, 27, 33, 0.1)',
+  // coolGrayAlpha20: 'rgba(17, 27, 33, 0.2)',
+  // coolGrayAlpha30: 'rgba(17, 27, 33, 0.3)',
+  // coolGrayAlpha40: 'rgba(17, 27, 33, 0.4)',
+  // coolGrayAlpha50: 'rgba(17, 27, 33, 0.5)',
+  // coolGrayAlpha60: 'rgba(17, 27, 33, 0.6)',
+  // coolGrayAlpha70: 'rgba(17, 27, 33, 0.7)',
+  // coolGrayAlpha80: 'rgba(17, 27, 33, 0.8)',
+  // coolGrayAlpha90: 'rgba(17, 27, 33, 0.9)',
```
necolas, line 54: Remove

packages/babel-plugin/__fixtures__/colorThemes.js
```diff
@@ -0,0 +1,760 @@
...
+ * @flow strict
+ */
+
+import type { Theme } from '@stylexjs/stylex';
+
+import { colors } from './colors.stylex';
+
+import * as stylex from '@stylexjs/stylex';
+
+// Base WDS color palette; not meant to be used directly. Instead use semantic colors.
```
necolas, line 16: Remove internal references

packages/babel-plugin/__fixtures__/colors.stylex.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+import type { VarGroup } from '@stylexjs/stylex';
+
+import * as stylex from '@stylexjs/stylex';
+
+opaque type SemanticColorsTheme: 'theme' = 'theme';
+
+const SemanticColors = {
+  '--accent': null as null | string,
```
necolas, line 17: Weird pattern to have all these values as `null`?

nmn: This is a workaround for a performance pitfall in browsers, but not needed for this test.

packages/babel-plugin/__fixtures__/colorThemes.js
```diff
@@ -0,0 +1,760 @@
...
+  linesDivider: baseColorPalette.blackAlpha10,
+  linesOutlineDefault: baseColorPalette.neutralGray400,
+  linesOutlineDeemphasized: baseColorPalette.blackAlpha20,
+
+  persistentAlwaysBranded: baseColorPalette.green500,
+  persistentAlwaysBlack: baseColorPalette.neutralGray1000,
+  persistentAlwaysWhite: baseColorPalette.whiteOpaque,
+  persistentActivityIndicator: baseColorPalette.green400,
+
+  productSystemsBubbleSurfaceIncoming: baseColorPalette.whiteOpaque, // surfaceElevatedDefault
```
mellyeliu, line 278: can probably remove these comments as well

packages/babel-plugin/__tests__/stylex-perf-create-theme-test.js
```diff
@@ -0,0 +1,48 @@
...
+      [stylexPlugin, { ...defaultOpts, ...opts }],
+    ],
+  });
+  return { code: result.code, styles: result.metadata.stylex };
+}
+
+const themeFile = path.resolve(__dirname, '../__fixtures__/colorThemes.js');
+
+describe('create theme', () => {
+  it('transform complex theme file', () => {
```
necolas, line 43:
```suggestion
  test('transform complex theme file', () => {
```

packages/babel-plugin/__tests__/stylex-perf-create-theme-test.js
```diff
@@ -0,0 +1,48 @@
...
+const classNamePrefix = 'x';
+const defaultOpts = {
+  stylexSheetName: '<>',
+  unstable_moduleResolution: { type: 'haste' },
+  classNamePrefix,
+};
+
+function transform(file, opts = defaultOpts) {
+  const result = transformFileSync(file, {
+    filename: opts.filename || '/stylex/packages/TestTheme.stylex.js',
```
necolas, line 27: Having a fake default filename here is a bit confusing, esp since we're passing in a "theme" in the test but the fallback is "Theme.stylex.js" which is a file suffix typically used for files containing defineVars exports

packages/babel-plugin/__tests__/stylex-perf-create-theme-test.js
```diff
@@ -0,0 +1,48 @@
+/**
```
necolas, line 1: Can you explain what this test is for? I thought we already had unit tests for theme creation. This is a unit test, so doesn't benchmark anything. If we want a perf benchmark, we should add to the benchmarks instead.

packages/babel-plugin/src/utils/evaluate-path.js
```diff
@@ -555,13 +555,19 @@ function _evaluate(path: NodePath<>, state: State): any {
...
             confident,
             deopt: resultDeopt,
             reason: deoptReason,
             value,
-          } = evaluate(keyPath, state.traversalState, state.functions);
+          } = evaluate(
+            keyPath,
+            state.traversalState,
+            state.functions,
+            state.seen,
```
mellyeliu, line 568: is the edge case that we weren't properly checking the value of seen here? can we update the summary to include more details

packages/babel-plugin/__tests__/stylex-perf-create-theme-test.js
nmn: I added it as a test as I was trying to debug the root cause of the slowdown. It's not quite a benchmark, it's just a verification that the transformation isn't taking an unreasonably long amount of time.

I'm open to suggestions on how exactly to define this. It could be a benchmark, but I'm not sure what that would look like.

necolas: No worries, I'll have a go at converting this into a benchmark next week

mellyeliu:
rebase

mellyeliu (approved):

packages/babel-plugin/__tests__/stylex-perf-create-theme-test.js
```diff
@@ -0,0 +1,69 @@
...
+
+    const start = performance.now();
+    const result = transform(themeFile);
+    const end = performance.now();
+    const timeTaken = end - start;
+    expect(result.code).toMatchSnapshot();
+    expect(result.styles).toMatchSnapshot();
+    console.log('timeTaken', timeTaken);
+
+    expect(timeTaken).toBeLessThan(simpleTimeTaken * 20);
```
mellyeliu, line 67: this number feels a bit arbitrary to me, but let's have the benchmark conversion as follow up to unblock internal

necolas:
PR for the createTheme benchmark - https://github.com/facebook/stylex/pull/974

PR #943, opened 2025-03-12
feat: improve media query parser implementation

mellyeliu:

Addes a new `normalize` function in the parser to flatten nested ands and remove unnecessary negations (negations % 2) 

To flatten ands:
  - When the function encounters an `and` rule, it loops over each of its child rules
  - If a child rule is also an `and` rule, its own child rules are merged (or "spread") into the parent `and` rule
  - Results in one `and` rule containing all the conditions on the same level

Remove nested nots:
  - When a `not` rule is encountered, the function recursively normalizes its inner rule and counts consecutive `not` wrappers.  
    - If the count is **even**, the negations cancel out, and the inner rule is returned as if no `not` was applied
    - If the count is **odd**, the result is a single `not` wrapping the innermost rule
  - This cancellation avoids unnecessary nesting and keeps the final query concise

Also fixes:
- Previous `toString` implementation errors
- Tests for `toString` fx in media query parsing. We want to ensure that if we parse a media query and convert it back to a string that we should get the same output.
- Added tests for flattening ands

Note: There's a small number of media query edge cases that are failing with the current parser implementation that we will have to investigate. Double negations cannot be fully tested until then

## Pre-flight checklist

### DIFF
--- a/packages/style-value-parser/src/at-queries-from-tokens/media-query.js
+++ b/packages/style-value-parser/src/at-queries-from-tokens/media-query.js
@@ -61,6 +61,12 @@ const _mediaKeywordParser: TokenParser<'screen' | 'print' | 'all'> =
     '=== "screen" | "print" | "all"',
   );
 
+function _isAndOrRule(
+  rule: MediaQueryRule,
+): rule is MediaAndRules | MediaOrRules {
+  return rule.type === 'and' || rule.type === 'or';
+}
+
 const mediaKeywordParser: TokenParser<MediaKeyword> = TokenParser.sequence(
   TokenParser.string('not').optional,
   _mediaKeywordParser,
@@ -281,9 +287,9 @@ export class MediaQuery {
     this.queries = queries;
   }
   toString(): string {
-    return `@media ${this.#toString(this.queries)}`;
+    return `@media ${this.#toString(this.queries, true)}`;
   }
-  #toString(queries: MediaQueryRule): string {
+  #toString(queries: MediaQueryRule, isTopLevel: boolean = false): string {
     switch (queries.type) {
       case 'media-keyword':
         return queries.not ? `not ${queries.key}` : queries.key;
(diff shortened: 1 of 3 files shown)


packages/style-value-parser/src/at-queries-from-tokens/media-query.js
```diff
@@ -299,15 +305,55 @@ export class MediaQuery {
...
+            flattened.push(norm);
+          }
+        }
+        return { type: 'and', rules: flattened };
+      }
+      case 'or':
+        return {
+          type: 'or',
+          rules: rule.rules.map((r) => MediaQuery.normalize(r)),
+        };
```
nmn, line 339: You can flatten Or rules too. `or` that contains `or`.

packages/style-value-parser/src/at-queries-from-tokens/media-query.js
```diff
@@ -299,15 +305,55 @@ export class MediaQuery {
...
+        let count = 1;
+        let inner = MediaQuery.normalize(rule.rule);
+        while (inner.type === 'not') {
+          count++;
+          inner = MediaQuery.normalize(inner.rule);
+        }
+        if (inner.type === 'pair' || inner.type === 'word-rule') {
+          return count % 2 === 0 ? inner : { type: 'not', rule: inner };
+        }
+        return inner;
```
nmn, line 350: Collapsing a not of a not is a good start. Some additional ways to simplify:

- `not a <= b` can become `a > b` etc.
- `not a || not b` can become `not (a && b)`
- etc.

packages/style-value-parser/src/at-queries-from-tokens/media-query.js
```diff
@@ -281,9 +287,9 @@ export class MediaQuery {
     this.queries = queries;
   }
   toString(): string {
-    return `@media ${this.#toString(this.queries)}`;
+    return `@media ${this.#toString(this.queries, true)}`;
   }
-  #toString(queries: MediaQueryRule): string {
+  #toString(queries: MediaQueryRule, isTopLevel: boolean = false): string {
     switch (queries.type) {
```
nmn, line 293: I suggest calling `normalize` in this function itself so you're guaranteed to always get the most simplified string representations of whatever rule has been parsed.

nmn (approved):
There are many opportunities for further improvements, but this PR is a solid improvement and a step in the right direction.

packages/style-value-parser/src/at-queries-from-tokens/media-query.js
mellyeliu: yeppp all those on my todo

mellyeliu: + a merge intervals sort of thing across the ranges

packages/style-value-parser/src/at-queries-from-tokens/media-query.js
mellyeliu: added to constructor, then will update https://github.com/facebook/stylex/pull/898 to instantiate new media query in return of `combineMediaQueryWithNegations`. or combine that as an andNot in the parser itself. we shall see..

PR #944, opened 2025-03-13
feat: create new `defineConsts` api

mellyeliu:
## Context

Adds support for new API `stylex.defineConsts` to inject constants into the css transform pipeline and replace `var(--token)` usage with raw values when processing rules

```js
// colors.stylex.js
import stylex from 'stylex'

export const colors = stylex.defineConsts({
  red: '#ff0000',
  green: '#00ff00',
})

// component.js
import stylex from 'stylex'
import { colors } from './colors.stylex'

  box: {
    color: colors.red,
    backgroundColor: colors.green,
  },
})
```

### shared 
- accepts a JS object of constants  
- returns an object mapping unique hashed keys to `{ constKey, constVal }`  
- includes the hash as `key` to match css variable format

### defineConsts babel transform

1. Walk ast and detect call to `stylex.defineConsts`
4. call `stylexDefineConsts` to convert obj to js output and metadata
5. replace the call in the AST with a plain JS object (primarily for debugging capabilities)
   `const colors = { red: '#ff0000' }`
6. register styles into `state.registerStyles` using the hash and `{ constKey, constVal }` to be used in rules processing

### `processStylexRules`

1. extract `constKey` and `constVal` from any rules that came from `defineConsts`  
   store them in dict from `var(--hash)` to value
2. for each css rule in `rules`, replace:
   - `var(--hash)` to raw value
   - `var(--hash, fallback)` to raw value (to do in follow up with TokenParser)
   this is done for both `ltr` and `rtl` if present 

## Testing
Added transform and shared utils tests

Tested E2E in examples 
- General usecases: number, string values etc
- Nested styles
```
export const COLOR = stylex.defineConsts({
  red: 'rgb(225, 135, 135)',
  blue: 'rgb(0, 0, 255)',
  green: 'rgb(134, 150, 134)',
});

export const BUTTON_COLOR = stylex.defineConsts({
  red: COLOR.red,
});

```
image

### DIFF
--- a/packages/babel-plugin/__tests__/stylex-transform-define-consts-test.js
+++ b/packages/babel-plugin/__tests__/stylex-transform-define-consts-test.js
@@ -1,0 +1,322 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ *
+ */
+
+'use strict';
+
+jest.autoMockOff();
+
+const { transformSync } = require('@babel/core');
+const stylexPlugin = require('../src/index');
+import { messages } from '@stylexjs/shared';
+
+const classNamePrefix = 'x';
+const defaultOpts = {
+  unstable_moduleResolution: { type: 'haste' },
+  classNamePrefix,
+  debug: false,
+};
+
+function transform(source, opts = {}) {
+  return transformSync(source, {
+    filename: opts.filename || '/stylex/packages/TestTheme.stylex.js',
+    parserOpts: {
+      flow: 'all',
+    },
+    babelrc: false,
+    plugins: [
+      [
+        stylexPlugin,
+        {
+          runtimeInjection: true,
+          unstable_moduleResolution: { type: 'haste' },
+          ...defaultOpts,
+          ...opts,
+        },
+      ],
+    ],
+  });
+}
+
+describe('@stylexjs/babel-plugin', () => {
+  describe('[transform] stylex.defineConsts()', () => {
+    test('transforms constants object', () => {
+      const { code, metadata } = transform(`
+        import stylex from 'stylex';
+        export const breakpoints = stylex.defineConsts({
+          sm: '(min-width: 768px)',
+          md: '(min-width: 1024px)',
+          lg: '(min-width: 1280px)',
+        });
+      `);
+
+      expect(code).toMatchInlineSnapshot(`
+        "import stylex from 'stylex';
+        export const breakpoints = {
+          sm: "(min-width: 768px)",
+          md: "(min-width: 1024px)",
+          lg: "(min-width: 1280px)",
+          __constName__: "TestTheme.stylex.js//breakpoints",
+          __constHash__: "1050fec2"
+        };"
+      `);
+
+      expect(metadata.stylex).toMatchInlineSnapshot(`
+        [
+          [
+            "x1izlsax",
+            {
+              "constKey": "x1izlsax",
+              "constVal": "(min-width: 768px)",
+              "ltr": "",
+              "rtl": null,
+            },
+            0,
+          ],
+          [
+            "xe5hjsi",
+            {
+              "constKey": "xe5hjsi",
+              "constVal": "(min-width: 1024px)",
+              "ltr": "",
+              "rtl": null,
+            },
+            0,
+          ],
+          [
+            "xmbwnbr",
+            {
+              "constKey": "xmbwnbr",
+              "constVal": "(min-width: 1280px)",
+              "ltr": "",
+              "rtl": null,
+            },
+            0,
+          ],
+        ]
+      `);
+    });
+
+    test('throws error for constant keys that start with --', () => {
+      expect(() =>
+        transform(`
+          import stylex from 'stylex';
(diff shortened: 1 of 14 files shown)


BenjaminLucier:
@mellyeliu Is this PR setting up to enable. this RFC https://github.com/facebook/stylex/discussions/683 ?

I left a comment on the RFC asking something similar as well https://github.com/facebook/stylex/discussions/683#discussioncomment-12589447

mellyeliu:
@BenjaminLucier yep! we're implementing the base API first to allow for shareable consts, then will expand to media queries once we make some improvements to media query parsing

mellyeliu:
@nmn requesting review

packages/babel-plugin/__tests__/stylex-transform-define-consts-test.js
```diff
@@ -0,0 +1,354 @@
...
+      `);
+    });
+
+    test('transforms literal constants object', () => {
+      const { code, metadata } = transform(`
+        import stylex from 'stylex';
+        export const spacing = stylex.defineConsts({
+          '--small': '8px',
+          '--medium': '16px',
+          '--large': '24px',
```
nmn, line 109: This should throw an error. Disallow usage of keys that start with `--` to be used as consts.

Using `--` lets you take control of the actual strings and this could be dangerous when it comes to consts as need to guarantee that they are replaced at compile time.

packages/babel-plugin/__tests__/stylex-transform-define-consts-test.js
```diff
@@ -0,0 +1,354 @@
...
+              "constVal": "1200px",
+              "ltr": "",
+              "rtl": null,
+            },
+            0,
+          ],
+          [
+            "x4spo47",
+            {
+              "constKey": "font-size*large",
```
nmn, line 300: the const key should be escaped to only use valid characters for CSS.

nmn: The usage of such a key will probably be escaped when it is referenced within a `stylex.create`.

nmn (changes requested):
Found some issues so far. Will continue to review further tomorrow.

packages/babel-plugin/__tests__/stylex-transform-define-consts-test.js
mellyeliu: I'm not building the key the same way as in `defineVars` (ie. key is hashed regardless of if it starts with `--` but agree it's better for clarity

mellyeliu:
@nmn requesting rereview

packages/babel-plugin/src/index.js
```diff
@@ -335,10 +346,17 @@ function processStylexRules(
     return '';
   }
 
+  const constsMap: Map<string, string | number> = new Map();
+  for (const [keyhash, ruleObj] of rules) {
+    if (ruleObj?.constKey && ruleObj?.constVal) {
+      constsMap.set(`var(--${keyhash})`, ruleObj.constVal);
+    }
+  }
+
```
nmn, line 355: Do a filter to separate the rules that defineConsts from the rest of the rules. This way `sortedRules` should not contain any constants.

Also, You'd need to recursively check the values of the constants to replace any constant that references another constant.

packages/shared/__tests__/stylex-define-consts-test.js
```diff
@@ -0,0 +1,189 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import styleXDefineConsts from '../src/stylex-define-consts';
+import createHash from '../src/hash';
+import * as messages from '../src/messages';
+
+const themeName = 'TestTheme.stylex.js//buttonTheme';
```
nmn, line 14: rename this constant.

packages/shared/__tests__/stylex-define-consts-test.js
```diff
@@ -0,0 +1,189 @@
...
+    expect(jsOutput).toEqual({
+      sm: '(min-width: 768px)',
+      md: '(min-width: 1024px)',
+    });
+
+    expect(injectableStyles).toMatchObject({
+      [classNamePrefix + createHash(`${themeName}.sm`)]: {
+        constKey: 'sm',
+        constVal: '(min-width: 768px)',
+        key: classNamePrefix + createHash(`${themeName}.sm`),
```
nmn, line 37: Use `constKey` for the generated varName.
```suggestion
        constKey: classNamePrefix + createHash(`${themeName}.sm`),
        constVal: '(min-width: 768px)',
```

nmn (approved):
Make the fixes please!

mellyeliu:
TODOs:
- Add support for use within `stylex.firstThatWorks`
- Testing inlining the generated CSS to a `stylex.create` call similar to what we have in `transform-import-export-test.js`
- Hot reloading isn't triggered upon constant change. I tried hashing the const rules and injecting it into the JS object, but did not the updates until rebuilding the app or checking out to a different branch entirely

packages/shared/__tests__/stylex-define-consts-test.js
mellyeliu: will do in follow up for both defineVars, defineConsts, and I think similar naming is used in evaluate path etc..

PR #950, opened 2025-03-17
Rename 'apps' to 'examples'

necolas:

Renames the 'apps' dir to 'examples'. Going forward we can include more examples of app/bundler integrations, as well as non-app examples like a demo component library that can be imported into app examples.


mellyeliu (approved):
LGTM!

component library + improved examples was an intern project idea I was thinking about for the upcoming summer internships, might share a project plan proposal

PR #953, opened 2025-03-18
Reorganize monorepo scripts

necolas:

Minor reorganization of the repo's scripts and tools. Private package names aren't scoped to `@stylexjs` and the release tooling is moved up to the root package.


package.json
```diff
@@ -11,7 +11,8 @@
     "prettier": "prettier --write \"**/*.js\" \"**/*.ts\"",
     "prettier:report": "prettier --check \"**/*.js\"",
     "lint": "npm run lint:report -- --fix",
-    "lint:report": "eslint",
+    "lint:report": "eslint packages tools",
```
necolas, line 14: eslint wasn't linting

package.json
```diff
@@ -38,7 +39,7 @@
     "prettier": "^3.5.0",
     "prettier-plugin-hermes-parser": "^0.26.0",
     "rimraf": "^5.0.10",
-    "typescript": "^5.3.3"
```
necolas, line 41: typescript doesn't seem to be used by any monorepo tooling

necolas:
ping @mellyeliu

packages/scripts/README.md
```diff
@@ -1,37 +1,3 @@
-# @stylexjs/scripts
+# scripts
```
mellyeliu, line 1: what's the distinction we're making here between `scripts` and `tools` here? `tools` specifically as broader developer tooling unrelated to code logic?

mellyeliu (approved):

packages/scripts/README.md
necolas: At the moment, everything left in `scripts` is installed as a `bin` in the dependent packages - so they're basically CLI tools used within some of the packages. The root tools are limited to managing monorepo-wide devx tasks like publishing to npm, precommit checks, etc

PR #957, opened 2025-03-20
Improve monorepo eslint config

necolas:
1. Fix the copyright header lint rule. Extra blank comment lines are now avoided. All pragmas are preserved.
2. Get eslint to process `*.mjs` files.
3. Update the React and Flow versions specified in the config.
4. Lint the example directories.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,12 +7,14 @@
  *
  */
 
+const path = require('path');
+
 module.exports = {
   settings: {
     react: {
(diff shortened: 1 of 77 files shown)


nmn:
> Fix the copyright header lint rule. Extra blank comment lines are now avoided. All pragmas are preserved.

Hallelujah! I struggled with this SO much! It's a trivial thing, but it has always bugged me!

nmn (approved):

PR #954, opened 2025-03-19
Delete esbuild and webpack plugins

necolas:
## What changed / motivation ?

Removes the remaining deprecated plugins - `esbuild-plugin` and `webpack-plugin`.

PR #960, opened 2025-03-22
chore: remove leftover esbuild and webpack examples and references

mellyeliu:

Just removing some legacy examples and references that rely on https://github.com/facebook/stylex/pull/954

## Pre-flight checklist


necolas (changes requested):
We still want these examples, but they'll eventually need to be updated to use the supported approaches.

necolas (approved):

hauxir:
> We still want these examples, but they'll eventually need to be updated to use the supported approaches.

What are the supported approaches?

PR #961, opened 2025-03-22
chore: improve documentation for lint rules

mellyeliu:
## Context

From discussions with ESLint oncall StyleX lint rules often provide lots of confusion, particularly around logical -> standard properties. We've previously added autofixes, but updating rules documentation here and internally to reduce confusion

## Pre-flight checklist

### DIFF
--- a/packages/eslint-plugin/README.md
+++ b/packages/eslint-plugin/README.md
@@ -1,8 +1,7 @@
 # @stylexjs/eslint-plugin
 
-The ESLint rule is a standalone ESLint plugin that mostly maintains an `allowlist` for valid styles and their valid values.
-
-This was originally created from Flow types which is why it's currently not very powerful.
+The ESLint rule is a standalone ESLint plugin that mostly maintains an
+`allowlist` for valid styles and their valid values.
 
 ## Installation
 
@@ -12,18 +11,21 @@ npm install --save-dev @stylexjs/eslint-plugin
 
 ## Enable Flow Types
 
-If you need to import this package in an environment where you need the Flow type definitions
-shipped with this package to work, please add the following options to your `.flowconfig` file:
+If you need to import this package in an environment where you need the Flow
+type definitions shipped with this package to work, please add the following
+options to your `.flowconfig` file:
 
 ```
 module.system.node.resolve_dirname=flow_modules
 module.system.node.resolve_dirname=node_modules
 ```
+
 This will fix the missing types for `eslint` and `estree`.
 
 ## Enabling the plugin and rules
 
-Once you've installed the npm package you can enable the plugin and rules by opening your ESLint configuration file and adding the plugin and rules.
+Once you've installed the npm package you can enable the plugin and rules by
+opening your ESLint configuration file and adding the plugin and rules.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/README.md
```diff
@@ -1,8 +1,10 @@
 # @stylexjs/eslint-plugin
 
-The ESLint rule is a standalone ESLint plugin that mostly maintains an `allowlist` for valid styles and their valid values.
+The ESLint rule is a standalone ESLint plugin that mostly maintains an
+`allowlist` for valid styles and their valid values.
 
-This was originally created from Flow types which is why it's currently not very powerful.
+This was originally created from Flow types which is why it's currently not very
```
necolas, line 6: Personally, I don't think context like this is helpful or worth including.

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
 
 ### @stylexjs/valid-styles
 
-StyleX requires styles that are statically analyzable. This rule will detect invalid styles that stylex cannot handle.
+StyleX requires styles that are statically analyzable. This rule will detect
+invalid styles that stylex cannot handle.
```
necolas, line 48:
```suggestion
invalid styles that StyleX cannot handle.
```

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
...
+
+#### Disallowed CSS properties and suggested fixes
+
+This guide lists common CSS properties that are **not allowed** in StyleX, along
```
necolas, line 52: Which "guide"?

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
...
+with their **suggested replacements**. The goal is to encourage explicit,
+modular, and logical CSS, with an explicit preference for longhand values.
```
necolas, line 54: I'm not sure what explicit, modular, logical CSS is in this context and how that's different from StyleX by default

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
...
+
+##### Disallowed: Logical styles for `margin` and `padding`. Note: this is autofixable.
```
necolas, line 56: This reads like it disallows logical direction CSS. But it's also confusing because it mentions `margin` which is a shorthand not a logical property

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
...
+
+## Logical styles for `margin` and `padding`
+
+To support better internationalization and writing-mode agnosticism, avoid
+directional or alias properties for margins and paddings. Use CSS logical
```
necolas, line 61: Disallows "physical" directions and non-standard "logical" direction property names.

Although I don't think we should be in a position where we have to mention the non-standard ones in OSS because support for that should be off by default. The long rule message if they're used should only be to nudge people in the right direction if they make that mistake. You shouldn't be able to disable warnings for those properties

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
...
+- `animationDirection`
+- `animationFillMode`
+- `animationPlayState`
 
 ### @stylexjs/sort-keys
 
 This rule helps to sort the StyleX property keys according to
 [property priorities](https://github.com/facebook/stylex/blob/main/packages/shared/src/utils/property-priorities.js).
+
+# Enforce standard logical CSS properties in StyleX
```
necolas, line 118: This title looks wrong?

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
...
+- `padding: '8px 16px 8px 16px'`
+
+#### Config options
+
+This rule has a few custom config options that can be set.
+
+```js
+{
+  allowImportant: false,                       // Whether `!important` is allowed
+  preferInline: false                          // Whether inline variants are preferred over directional
```
necolas, line 142: What does this mean?

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
...
+- `animationPlayState`
 
 ### @stylexjs/sort-keys
 
 This rule helps to sort the StyleX property keys according to
 [property priorities](https://github.com/facebook/stylex/blob/main/packages/shared/src/utils/property-priorities.js).
+
+# Enforce standard logical CSS properties in StyleX
+
+### Rule: `stylex-valid-shorthands`
```
necolas, line 120: Shouldn't all the stuff like `font` be under this section?

packages/eslint-plugin/README.md
```diff
@@ -39,9 +44,133 @@ Once you've installed the npm package you can enable the plugin and rules by ope
...
+
+- Exporting `stylex.defineVars()` in a file **not** ending in `.stylex.jsx` or
+  `.stylex.tsx`
+- Using the `.stylex.jsx` / `.stylex.tsx` extension without exporting
+  `stylex.defineVars()`
+- Mixing `stylex.defineVars()` with other exports in the same file
+
+#### Instead...
+
+- Use `.stylex.jsx` or `.stylex.tsx` for files that only export
```
necolas, line 161: Why jsx/tsx instead of js/mjs/ts? There shouldn't be any jsx in those files afaik

mellyeliu: some of this is generated by copilot and i am paying for it...

packages/eslint-plugin/README.md
mellyeliu: `preferInline` expands shorthands to standard logical equivalents when set to true, otherwise uses directional expansion (default).

ie. 
```
//prefer-inline: true
margin: 10 15 20 25, 

marginTop: 10px;
marginRight: 15px;
marginBottom: 20px;
marginLeft: 25px;

//prefer-inline: false
margin: 10 15 20 25, 

marginTop: 10px;
marginInlineEnd: 15px;
marginBottom: 20px;
marginInlineStart 25px;

```

packages/eslint-plugin/README.md
mellyeliu: yea this is just reformatted, but agree and may as well remove

packages/eslint-plugin/README.md
mellyeliu: meant to add this under valid-shorthands as rationale for expanding

packages/eslint-plugin/README.md
mellyeliu: ya, the categorization is also a bit messy because properties like `font` etc are also disallowed by `valid-styles`, but it probably fits better here

packages/eslint-plugin/README.md
mellyeliu: hmmm ya it's more of an internal issue, I included it here because of the apparent confusion it causes (even with the error message in place) and thought more places to surface couldn't hurt, but maybe it's more useful in internal lint documentation

necolas (approved):

PR #964, opened 2025-03-26
chore: disallow spreading in stylex.create

mellyeliu:
## Context

Explicitly disallow spreads within `stylex.create()` calls in the docs for planned js runtime merging optimizations. Update docs to reflect this.

## Testing

Added test with object spread in `stylex.create()` to ensure throw.

## Pre-flight checklist

### DIFF
--- a/packages/docs/docs/learn/04-styling-ui/01-defining-styles.mdx
+++ b/packages/docs/docs/learn/04-styling-ui/01-defining-styles.mdx
@@ -31,6 +31,7 @@ The following are **not** allowed:
 - Function calls (except StyleX functions)
 - Values imported from other modules (except for CSS Variables created using
   StyleX from a `.stylex.js` file.)
+- Object spreads (e.g., `{...style}`)
 
 ## Creating styles
 
--- a/packages/babel-plugin/__tests__/stylex-validation-declarations-test.js
+++ b/packages/babel-plugin/__tests__/stylex-validation-declarations-test.js
@@ -127,3 +127,21 @@ describe.skip('@stylexjs/babel-plugin', () => {
     });
   });
 });
+
+describe('stylex.create validation', () => {
+  test('throws on object spread in stylex.create', () => {
+    expect(() =>
+      transform(
+        `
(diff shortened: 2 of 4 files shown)


packages/docs/docs/learn/04-styling-ui/01-defining-styles.mdx
```diff
@@ -31,6 +31,7 @@ The following are **not** allowed:
 - Function calls (except StyleX functions)
 - Values imported from other modules (except for CSS Variables created using
   StyleX from a `.stylex.js` file.)
+- Top level object spreads
```
necolas, line 34: All object spreads are disallowed

packages/babel-plugin/src/utils/add-sourcemap-data.js
```diff
@@ -74,10 +74,12 @@ export function addSourceMapData(
...
       // $FlowIgnore
       .find((prop) => {
+        const k = prop.node.key;
         return (
-          prop.node.key.name === key ||
           // string and number properties (normalized to string)
-          String(prop.node.key.value) === key
+          (t.isIdentifier(k) && k.name === key) ||
+          (t.isStringLiteral(k) && k.value === key) ||
+          (t.isNumericLiteral(k) && String(k.value) === key)
```
necolas, line 82: This is the main change of the diff. Does this fix a bug that should have test coverage?

necolas:
> npx jest stylex-transform-create

You should run the package's test script instead, as this might not always perform the same task.

```
npm run test -w @stylexjs/babel-plugin
```

packages/babel-plugin/src/utils/add-sourcemap-data.js
mellyeliu: lemme separate this out into a different PR, but the main bug I found was when trying to spread, which we aren't allowing anyway, so I didn't add a test for it. it's more of a defensive type check + preventing unintended behaviour

necolas (approved):
Nice

PR #965, opened 2025-03-26
fix: add type safety to sourcemap construction

mellyeliu:
## Context

Adds type-safe key matching in addSourceMapData using explicit Babel typeguards (`t.isIdentifier`, `t.isStringLiteral`, `t.isNumericLiteral`). This improves type safety and prevents runtime errors  in edge cases like object spreads (disallowed now, but theoretical) or computed keys, as the AST shape can vary and .name/value may be undefined

## Testing
```
npm run test -w @stylexjs/babel-plugin
```

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/src/utils/add-sourcemap-data.js
+++ b/packages/babel-plugin/src/utils/add-sourcemap-data.js
@@ -74,10 +74,12 @@ export function addSourceMapData(
       .get('arguments.0.properties')
       // $FlowIgnore
       .find((prop) => {
+        const k = prop.node.key;
         return (
-          prop.node.key.name === key ||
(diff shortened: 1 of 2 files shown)


necolas (approved):
Would be good to have a transform to catch potential regressions here

packages/babel-plugin/src/utils/add-sourcemap-data.js
```diff
@@ -74,10 +74,12 @@ export function addSourceMapData(
       .get('arguments.0.properties')
```
nmn, line 74: Using this will remove the need for the `$Flowignore`
```
.get('arguments')[0].get('properties')
```

PR #966, opened 2025-03-27
feat: add new eslint rule for flagging legacy media query/pseudo class syntax

vincentriemer:
# Summary:

This diff adds a new eslint rule for flagging usage of stylex's legacy media query/pseudo class syntax that looks like this:

```js
  root: {
    width: '100%',
   '@media (min-width: 600px)': {
      width: '50%',
    }
  }
});
```

The lint messages direct developers to instead use this syntax:

```js
  root: {
    width: {
      default: '100%',
      '@media (min-width: 600px)': '50%'
    }
  }
});
```

The new lint rule is notably not auto-fixable in this diff but that is something I'm going to be working on next.

# Test Plan:

I introduced new eslint "RuleTester" tests that are passing.

### DIFF
--- a/packages/eslint-plugin/__tests__/stylex-no-legacy-media-queries-test.js
+++ b/packages/eslint-plugin/__tests__/stylex-no-legacy-media-queries-test.js
@@ -1,0 +1,88 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+jest.disableAutomock();
+
+const { RuleTester: ESLintTester } = require('eslint');
+const rule = require('../src/stylex-no-legacy-media-queries');
+
+const eslintTester = new ESLintTester({
+  parser: require.resolve('hermes-eslint'),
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/__tests__/stylex-no-legacy-media-queries-test.js
```diff
@@ -0,0 +1,85 @@
...
+const eslintTester = new ESLintTester({
+  parser: require.resolve('hermes-eslint'),
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+});
+
+eslintTester.run('stylex-no-legacy-media-queries', rule.default, {
+  valid: [
```
mellyeliu, line 24: can you add a test that checks nested media queries as well?

ie. something like

```
        main: {
          '::placeholder': {
            color: '#999',
          },
          width: {
            default: '100%',
            '@media (min-width: 600px)': {
                     default: '50%',
                     '@media screen': '40%'
              },
          }
        },
        ':dummy': {
          color: '#999',
        }
      });
```

mellyeliu (approved):
Nice! Let's also add documentation in `packages/eslint-plugin/README.md` about the rule

PR #967, opened 2025-03-30
Improve performance of example-nextjs by utilizing babel-loader

km-tr:

This PR improves the performance of the `example-nextjs` project by implementing
`babel-loader`. Utilizing `babel-loader` for JavaScript/TypeScript transpilation
leads to faster build times and an improved development experience compared to
the previous setup.

## Additional Context

**Technical Details:**

- Using `babel.config.js` previously forced all compilation through Babel,
  slowing down the build process.
- This change configures `babel-loader` within the webpack configuration
  (`next.config.js`). This ensures only files requiring StyleX compilation are
  processed by Babel, while the rest leverage Next.js's faster SWC compiler.
- Additionally, the PostCSS compilation target has been narrowed down to only
  `.tsx` and `.stylex.ts` files, further optimizing the build pipeline.

**Benchmark Results:**

_Command: `npm run -w ./examples/example-nextjs example:build`_

**Before:**

```
Time (mean ± σ):     14.754 s ±  0.343 s    [User: 24.845 s, System: 2.581 s]
Range (min … max):   14.385 s … 15.173 s    5 runs
```

**After:**

```
Time (mean ± σ):     11.423 s ±  0.395 s    [User: 19.161 s, System: 2.293 s]
Range (min … max):   10.918 s … 11.841 s    5 runs
```

**Internal Codebase Improvement:** Furthermore, we observed a significant build
time improvement in our internal company codebase, reducing from 397.48s to
271.48s after applying these changes.

**Other Changes:**

- The `example:build` script in `example-nextjs` was failing. Added `styled-jsx`
  to `examples/example-nextjs/package.json` as a dependency to resolve this
  issue.
- Due to a large number of pre-existing lint errors, linting during the Next.js
  build process has been temporarily disabled
  (`eslint: { ignoreDuringBuilds: true }` in `next.config.js`) to allow the
  build to complete successfully. Addressing these lint errors is outside the
  scope of this performance-focused PR.

**Acknowledgements:** Special thanks to my mentor @javascripter for their
valuable advice and guidance on implementing these performance improvements.

## Pre-flight checklist


javascripter:
For context:

Using`babel.config.js` forces Next.js to compile a lot of files that don't need to be transformed with StyleX. Using `babel-loader` allows you to selectively compile files with StyleX Babel Plugin while delegating most files to Next.js default config.

For example, if we print the path of processed files in StyleX Babel Plugin, we see lots of internal files that don't need to be processed by StyleX Babel Plugin in the current main branch.
```
   Disabled SWC as replacement for Babel because of custom Babel configuration "babel.config.js" https://nextjs.org/docs/messages/swc-disabled
 ✓ Ready in 1378ms
   Using external babel configuration from /Users/tsubasa/Projects/github.com/facebook/stylex/examples/example-nextjs/babel.config.js
StyleX: path /Users/tsubasa/Projects/github.com/facebook/stylex/examples/example-nextjs/app/layout.tsx
StyleX: path /Users/tsubasa/Projects/github.com/facebook/stylex/examples/example-nextjs/app/page.tsx
StyleX: path /Users/tsubasa/Projects/github.com/facebook/stylex/examples/example-nextjs/node_modules/next/dist/server/future/route-kind.js
StyleX: path /Users/tsubasa/Projects/github.com/facebook/stylex/examples/example-nextjs/node_modules/next/dist/server/app-render/entry-base.js
StyleX: path /Users/tsubasa/Projects/github.com/facebook/stylex/examples/example-nextjs/node_modules/next/dist/client/components/error-boundary.js
StyleX: path /Users/tsubasa/Projects/github.com/facebook/stylex/examples/example-nextjs/node_modules/next/dist/client/components/not-found-error.js
StyleX: path /Users/tsubasa/Projects/github.com/facebook/stylex/examples/example-nextjs/node_modules/next/dist/server/future/route-modules/app-page/module.compiled.js
 ○ Compiling / ...
```

Next.js itself [uses](https://github.com/vercel/next.js/blob/3261044fcb3b0fb96c021f4e0ba0b7a7c25b5b77/packages/next/src/build/get-babel-loader-config.ts#L40) babel-loader to run React Compiler with babel as well and this PR is also is compatible with their [`experimental.reactCompiler` option](https://nextjs.org/docs/app/api-reference/config/next-config-js/reactCompiler).

examples/example-nextjs/next.config.js
```diff
@@ -9,6 +9,45 @@
...
 
+const path = require('path');
+
 module.exports = {
   transpilePackages: ['@stylexjs/open-props'],
+  eslint: { ignoreDuringBuilds: true },
+  webpack: (config, { dev, isServer }) => {
+    // Process only files that require StyleX compilation using babel-loader
+    config.module.rules.push({
+      test: /\.(tsx|stylex\.ts|jsx|stylex\.js)$/,
```
javascripter, line 20: @km-tr 
This might be a bit too restrictive for general use as `.ts` and `.js` files may technically contain StyleX imports.

Might be better to just test for  `.js`, `.jsx`, `.ts`, `.tsx` in this example and let users fine-tune config according to their setup.

examples/example-nextjs/postcss.config.js
```diff
@@ -42,14 +42,39 @@ const openPropsIncludePaths = getPackageIncludePaths('@stylexjs/open-props', [
...
 
+const dev = process.env.NODE_ENV !== 'production';
+
 module.exports = {
   plugins: {
     '@stylexjs/postcss-plugin': {
       include: [
-        'app/**/*.{js,jsx,ts,tsx}',
-        'components/**/*.{js,jsx,ts,tsx}',
+        'app/**/*.{tsx,stylex.ts}',
```
javascripter, line 51: Same

examples/example-nextjs/next.config.js
km-tr: Thank you. I've made it more generic.

examples/example-nextjs/package.json
```diff
@@ -10,19 +10,20 @@
...
   },
   "dependencies": {
-    "@stylexjs/stylex": "0.11.1",
     "@stylexjs/open-props": "0.11.1",
+    "@stylexjs/stylex": "0.11.1",
     "next": "14.2.21",
     "react": "^18.3.0",
-    "react-dom": "^18.3.0"
+    "react-dom": "^18.3.0",
+    "styled-jsx": "^5.1.1"
```
necolas, line 18: Why do we need this dependency?

km-tr: This is to fix the build error in example-nextjs.

The root cause is that multiple versions of React are used in this repository. Specifically, React v17.0.2 is defined in docs, causing version 17 to be installed in the root’s node_modules. Meanwhile, examples/example-nextjs specifies React v18 directly in its own package.json, causing version 18 to be installed in its local node_modules.

Although example-nextjs typically uses React v18, the React invoked via styled-jsx during the build process was incorrectly referencing React v17. To address this, a temporary adjustment was made so that React called through styled-jsx would also reference examples/example-nextjs/node_modules.

necolas: Strange, but makes sense

BMCwebdev:
I know this PR is closed, but hoping I could still ask @javascripter and @km-tr a question in here. I love the work done here, and I have implemented it, but I run into an issue. Trying to get Jest to run without a `babel.config.js` I can't get Jest to compile the StyleX code. I also have a component library that is built with uncompiled StyleX code. So in consuming apps, this library is a packages in node_modules. This setup you have above works great for building and dev, but again, I cannot get Jest to work this way. I've tried creating a jest-transformer and jest.config to try and use the same setup, but still doesn't work

```
  The Jest Problem:
  The exact same StyleX configuration works in Next.js webpack but fails in
  Jest. Since the code is identical, this must be a Jest-specific module 
  resolution or processing issue.

  Key Clue:
  The error shows Jest referencing:
  /src/components/Theme/dynamicThemeUtils.jsx:14:42

  But we confirmed this path doesn't exist in the published package. This
  suggests Jest is either:

  1. Following source maps to non-existent source files
  2. Processing internal imports within the bundled dist files differently than
   webpack
  3. Using different module resolution with pnpm's symlink structure
  ```

**Comment by javascripter** [comment]
@BMCwebdev 
Have you configured jest with babel-jest to transform files with babel? I haven't verified the below but in my experience you need to follow these steps:

1. Update [transformIgnorePatterns](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring) to make sure your component library gets transformed via babel-jest.
2. Create babel.config.js and verify your tests pass. This confirms your babel config and `transformIgnorePatterns` are configured properly and is working as expected.
3. Either rename babel.config.js to custom-babel.config.js or inline your babel config there:

https://www.npmjs.com/package/babel-jest
```
// jest.config.js
"transform": {
  "\\.[jt]sx?$": [
    "babel-jest",
    {
      "extends": "./custom-babel.config.js" // set your custom babel config path
       // or inline your babel config here
    }
  ]
},
```

BMCwebdev:
Thanks @javascripter!

PR #974, opened 2025-04-07
Add benchmark for babel-plugin transform of createTheme

necolas:
This benchmark will help monitor improvements and regressions to the createTheme transform. We can use this benchmark infra to test the performance of other transform tests before we do any future refactoring.

After creating this benchmark, I locally reverted the change in #942 to validate the impact of that fix (higher is better)

| **Results** | **Base** | **Patch** | **Ratio** |  |
| :--- | ---: | ---: | ---: | ---: |
|  |  |  |  |
| **babel-plugin: createTheme** |  |  |  |  |
| &middot; basic theme | 449 | 453 | 1.01 |  |
| &middot; complex theme | 0.02 | 65.55 | 3277.50 |  |

Ref #942


packages/benchmarks/perf/fixtures/themes.js
```diff
@@ -9,9 +9,8 @@
```
mellyeliu, line 9: maybe we rename these `basic-themes.js` and `complex-themes.js` to be more clear about the distinction

.github/workflows/benchmarks.yml
```diff
@@ -3,6 +3,55 @@ name: benchmarks
...
+      run: |
+        git checkout -f ${{ github.event.pull_request.base.sha }}
+        npm install --loglevel error
+        if npm run perf -w benchmarks -- -o ${{ env.BASE_JSON }}; then
+          echo "Ran successfully on base branch"
+        else
+          echo "{}" > ${{ env.BASE_JSON }}  # Empty JSON as default
+          echo "Benchmark script not found on base branch, using default values"
+        fi
+    - name: 'Benchmark patch'
```
mellyeliu, line 29: should we clean workspace beforehand to purge untracked files/node-modules? 
```
- name: 'Clean workspace'
  run: git clean -fdx
```

mellyeliu (approved):

PR #977, opened 2025-04-08
[babel-plugin] Refactor stylex-transform-create-test.js

necolas:
Improve test coverage and make navigating the suite easier.

* Rename tests and organize into groups.
  * Consolidate redundant tests.
  * Move a 'props' tests to props test file.
  * Move tests for legacy API to bottom of file. 
  * Return `code` and `metadata` from `transform`; future iteration will remove use of `runtimeInjection` option from tests.
* Use OSS default imports and options.
  * Imports are updated to be `import * as stylex from '@stylexjs/stylex'` (except legacy tests).
  * Tests run using cjs modules by default.
* Add missing tests for dynamic styles.
  * Tests for use with pseudo-class, pseudo-element, and queries.

image

image


nmn (approved):
Thanks! Long overdue.

mellyeliu (approved):
good timing :D

Issue #978, opened 2025-04-08
Pseudo elements not working within dynamic styles


mellyeliu:
### Describe the issue

When defining a pseudo elements within dynamic styles (or with inline css vars), value does not resolve correctly.

### Expected behavior

Pseudo elements should work correctly within dynamic styles calls

### Test case

```tsx
const styles = stylex.create({
  computedStyles: (computedValue: ComputedValues) => ({
    fontSize: computedValue.fontSize,
    lineHeight: computedValue.lineHeight,
    marginBottom: computedValue.baselineTrim, // value is set properly
    '::after': {
      content: "''",
      display: 'table',
      marginBottom: computedValue.baselineTrim, // value does not resolve
    },
  }),
});
```

necolas:
This is a snapshot of the generated output for that test case:

```js
          expect(code).toMatchInlineSnapshot(`
            "import _inject from "@stylexjs/stylex/lib/stylex-inject";
            var _inject2 = _inject;
            import * as stylex from '@stylexjs/stylex';
            _inject2(".x6zurak{font-size:var(--fontSize)}", 3000);
            _inject2(".x18bv5gf{line-height:var(--lineHeight)}", 3000);
            _inject2(".x138musq{margin-bottom:var(--marginBottom)}", 4000);
            _inject2(".x100rkj9::after{content:''}", 8000);
            _inject2(".x1azrkns::after{display:table}", 8000);
            _inject2(".xi57mhz::after{margin-bottom:var(--s7ux03)}", 9000);
            _inject2("@property --fontSize { syntax: \\"*\\"; inherits: false;}", 0);
            _inject2("@property --lineHeight { syntax: \\"*\\"; inherits: false;}", 0);
            _inject2("@property --marginBottom { syntax: \\"*\\"; inherits: false;}", 0);
            _inject2("@property --s7ux03 { syntax: \\"*\\"; inherits: false;}", 0);
            export const styles = {
              computedStyles: computedValue => [{
                kGuDYH: "x6zurak",
                kLWn49: "x18bv5gf",
                k1K539: "x138musq",
                k5JduY: "x100rkj9",
                kc3QkE: "x1azrkns",
                k5jvz2: "xi57mhz",
                $$css: true
              }, {
                "--fontSize": (val => typeof val === "number" ? val + "px" : val != null ? val : undefined)(computedValue.fontSize),
                "--lineHeight": computedValue.lineHeight != null ? computedValue.lineHeight : undefined,
[...]

PR #980, opened 2025-04-09
fix: fix pseudo elements in dynamic styles

mellyeliu:

Pseudo elements not working in dynamic styles because we're setting `inherits: false`, see [978](https://github.com/facebook/stylex/issues/978)

## Linked PR/Issues

Fixes [978](https://github.com/facebook/stylex/issues/978)

## Testing

Ran on tests added in https://github.com/facebook/stylex/pull/977
Before:
image

After
image

Value is now set correctly

## Pre-flight checklist

### DIFF
--- a/packages/babel-plugin/__tests__/transform-create-test.js
+++ b/packages/babel-plugin/__tests__/transform-create-test.js
@@ -1185,7 +1185,7 @@ describe('@stylexjs/babel-plugin', () => {
                 color: {
                   ':hover': hover,
                   ':active': active,
-                  ':focus': focys,
+                  ':focus': focus,
                   ':nth-child(2n)': 'purple',
                 },
               }),
@@ -1210,7 +1210,7 @@ describe('@stylexjs/babel-plugin', () => {
               }, {
                 "--1113oo7": hover != null ? hover : undefined,
                 "--hxnnmm": active != null ? active : undefined,
-                "--8tbbve": focys != null ? focys : undefined
+                "--8tbbve": focus != null ? focus : undefined
               }]
             };"
           `);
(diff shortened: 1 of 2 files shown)


packages/babel-plugin/__tests__/stylex-transform-stylex-props-test.js
```diff
@@ -313,6 +313,64 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
 
+    test('Last property wins, even if shorthand', () => {
```
nitishmehrotra870, line 316: I dont see a test for pseudo elements specifically, are you planning to add it?

mellyeliu: this is out of date now let me push a rebase, but those are covered in the tests in `stylex-transform-create-test.js`

packages/babel-plugin/__tests__/stylex-transform-create-test.js
```diff
@@ -1346,6 +1346,52 @@ describe('@stylexjs/babel-plugin', () => {
...
 
+      test('"::before" and ":hover" side by side', () => {
+        const { code } = transform(`
+          import * as stylex from '@stylexjs/stylex';
+          export const styles = stylex.create({
+            foo: (a, b, c, d) => ({
+              color: a, // should get inherits: false
+              ':hover': {
+                backgroundColor: b, // should get inherits: false
+              },
```
necolas, line 1357: This is not the valid syntax for pseudo-classes. I don't think we need this extra test though - it looks redundant given we already have a test for pseudo-classes.

necolas (changes requested):
Please could you also fix a typo I spotted on L1188 - `':focus': focys,` - the variable name should be `focus`.

necolas (approved):

packages/babel-plugin/__tests__/stylex-transform-create-test.js
mellyeliu: ah that's right. thanks

mellyeliu:
synced changes and tested internally

PR #982, opened 2025-04-09
[babel-plugin] Rename unit test files

necolas:
Remove 'stylex' prefix from test names


packages/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
```diff
@@ -10,7 +10,7 @@
 jest.autoMockOff();
 
 const { transformSync } = require('@babel/core');
-const stylexPlugin = require('../src/index');
+const stylexPlugin = require('../../src/index');
```
mellyeliu, line 13: can be in a separate PR, but maybe we should set up a module alias so this isn't as prone to breaking

mellyeliu (approved):

PR #985, opened 2025-04-10
[babel-plugin] Refactor defineVars unit tests

necolas:
* Use OSS defaults in tests.
* Reorganize tests.
* Simplify fixtures to focus on subject of test.
* Snapshot metadata.

image

### DIFF
--- a/packages/babel-plugin/__tests__/transform-define-vars-test.js
+++ b/packages/babel-plugin/__tests__/transform-define-vars-test.js
@@ -12,19 +12,16 @@ jest.autoMockOff();
 const { transformSync } = require('@babel/core');
 const stylexPlugin = require('../src/index');
 
-const classNamePrefix = 'x';
 const defaultOpts = {
-  unstable_moduleResolution: { type: 'haste' },
-  classNamePrefix,
   debug: false,
+  unstable_moduleResolution: {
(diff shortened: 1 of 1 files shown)


mellyeliu (approved):

packages/babel-plugin/__tests__/transform-define-vars-test.js
```diff
@@ -12,797 +12,602 @@ jest.autoMockOff();
...
+                "rtl": null,
+              },
+              0,
+            ],
+          ],
+        }
+      `);
     });
 
-    test('transforms variables object with named import', () => {
```
mellyeliu, line 163: is there a reason we're dropping testing named imports? i think there's extra handling in the transform to scan named imports we probably do want to test

necolas: there's another test file just for imports - those tests don't belong here. we've been duplicating so many tests (without full coverage) just to check if they also work when the imports are aliased

mellyeliu: makes sense! are you referring to `transform-import-test`? we'll need to add tests for each transform separately to make sure it's still transpiled with aliases

PR #987, opened 2025-04-10
[babel-plugin] Snapshot metadata in transform-create-test

necolas:
* Add `metadata` snapshots to transform-create-test instead of using `runtimeInjection`.
* Delete separate `metadata-test` file.


nmn (approved):
makes sense.

PR #991, opened 2025-04-11
[babel-plugin] Refactor transform import test

necolas:
* Rename test to transform-import-export-test
* Add fixture factories to help:
  a) Assert identical metadata across tests
  b) Test all exports from StyleX

I don't usually like having "programs" in tests, but in this case it's important to be able to use the same fixtures across many tests that have different ways of importing and calling stylex exports.


packages/babel-plugin/__tests__/transform-import-export-test.js
```diff
@@ -0,0 +1,527 @@
...
+/**
+ * Fixture factory
+ *
+ * This is used to create a consistent fixture across all the tests for
+ * different ways of importing StyleX. The fixture uses all of the
+ * StyleX exports to make sure they are transformed as expected.
+ */
+
+const defaultImportSource = {
+  imprt: '* as stylex',
```
nmn, line 47: typo?

nmn: I think this is intentional. I think adding an `_` is preferrable than dropping the `o`.

nmn (approved):

packages/babel-plugin/__tests__/transform-import-export-test.js
```diff
@@ -0,0 +1,527 @@
...
+'use strict';
+
+jest.autoMockOff();
+
+const { transformSync } = require('@babel/core');
+const stylexPlugin = require('../src/index');
+
+function transform(source, opts = {}) {
+  const { code, metadata } = transformSync(source, {
+    filename: opts.filename || '/stylex/packages/vars.stylex.js',
```
necolas, line 17: This line surfaces a difference between the `importSources` tests and all the others. If you remove the fallback filename, only the `importSources` tests fail:

```
  ● @stylexjs/babel-plugin › [transform] stylex imports › importSources (string)

    /stylex/packages/fixture.js: No filename found for generating theme name.
```

And if you use the same fallback filename for every file, the `importSources` tests pass but new CSS is added to the metadata.

Either we have a bug in the compiler, or there's some esoteric detail I don't understand.

necolas:
@nmn Any idea why there is an difference between metadata and thrown errors depending on whether `importSources` is used?

necolas:
Previous iteration had a bug in the test fixture factory that prevented `defineVars` from being compiled. Fixed now

PR #993, opened 2025-04-11
Remove the 'stylex.attrs' API

necolas:
This API was a copy-paste of the `props` API, with additional runtime overhead. Given that the majority of StyleX users - including Meta - don't use this API, it is removed from core. The user-space 'attrs' wrapper around props is documented for non-React users.


packages/docs/docs/api/javascript/props.mdx
```diff
@@ -55,3 +48,33 @@ const styles = stylex.create({
   )}
 />;
 ```
+
+### Not using React?
+
+:::tip For Solid, Svelte, Qwik, Vue
+
+For frameworks that expect `class` instead of `className`, and expect `style` to be a string instead of an object, you can
+wrap the `props` API and perform the relevant transform. Then use this function in place of `stylex.props`.
```
mellyeliu, line 57:
```suggestion
For frameworks that expect `class` instead of `className`, and require `style` to be a string instead of an object,
wrap the `props` API with the function below and use it in place of `stylex.props`.
```

mellyeliu (approved):

nmn:
Could we move `attrs` to a separate file to fix the runtime overhead but keep it in core?

Moving it out of core would mean that we would not be able to optimize `attrs()` calls like we do for `props()`.

necolas:
Maybe you could use something like this instead and add it to stylextras.

```jsx
// Either add the wrapper manually or have a babel plugin do it
<div {...attrs(stylex.props(styles))} />
```

```js
export function attrs({ className, 'data-style-src': dataStyleSrc, style }) {
  const result = {};
  // Convert className to class
  if (className != null && className !== '') {
    result.class = className;
  }
  // Convert style object to string
  if (style != null && Object.keys(style).length > 0) {
    result.style = Object.keys(style)
      .map((key) => `${key}:${style[key]};`)
      .join('');
  }
  if (dataStyleSrc != null && dataStyleSrc !== '') {
    result['data-style-src'] = dataStyleSrc;
  }
  return result;
}
```

I don't think it makes sense for us to maintain a duplicate API here though. We want APIs like this, `inline()`, etc., to something people can build on top of StyleX in user-space.

nmn:
Ok. Hold off on merging this till next week then. I'll create a Babel plugin for this in Stylextras.

necolas:
I'd like to merge this now. Can you use the history to find the source code when you have time to create something in Stylextras?

hyperknot:
This introduced a performance regression, when following the online docs. The bundle size is now much bigger, as well as none of the static compilations are in place now. Basically it kind of castrated the whole performance core of StyleX, when using non-React frameworks.

Please advise us how to add it back such that the performance does not suffer. Adding a runtime function is not a solution.

I've made a Vite example for React and Solid here, the Solid one is affected:
https://github.com/hyperknot/bug_repro/tree/main/stylex_solid_attr_performance

Before: JS size 26 kB
After: JS size 37 kB

Here is the diff for my JS bundle:

```diff
diff --git a/index-SmZ8gE2Y.js b/index-CDa1jOWi.js
index 0073ee9..4f78774 100644
--- a/index-SmZ8gE2Y.js
+++ b/index-CDa1jOWi.js
@@ -34,10 +34,16 @@
     const fetchOpts = getFetchOpts(link);
     fetch(link.href, fetchOpts);
   }
 })();
 const IS_DEV = false;
+const equalFn = (a, b) => a === b;
+const $PROXY = Symbol("solid-proxy");
+const SUPPORTS_PROXY = typeof Proxy === "function";
+const signalOptions = {
+  equals: equalFn
+};
 let runEffects = runQueue;
 const STALE = 1;
 const PENDING = 2;
 const UNOWNED = {
   owned: null,
@@ -70,10 +76,19 @@ function createRoot(fn, detachedOwner) {
 }
 function createRenderEffect(fn, value, options) {
   const c = createComputation(fn, value, false, STALE);
   updateComputation(c);
 }
+function createMemo(fn, value, options) {
+  options = options ? Object.assign({}, signalOptions, options) : signalOptions;
+  const c = createComputation(fn, value, true, 0);
+  c.observers = null;
+  c.observerSlots = null;
+  c.comparator = options.equals || void 0;
+  updateComputation(c);
+  return readSignal.bind(c);
+}
 function untrack(fn) {
   if (Listener === null) return fn();
   const listener = Listener;
   Listener = null;
   try {
@@ -81,10 +96,39 @@ function untrack(fn) {
     return fn();
   } finally {
     Listener = listener;
   }
 }
+function readSignal() {
+  if (this.sources && this.state) {
+    if (this.state === STALE) updateComputation(this);
+    else {
+      const updates = Updates;
+      Updates = null;
+      runUpdates(() => lookUpstream(this), false);
+      Updates = updates;
+    }
+  }
+  if (Listener) {
+    const sSlot = this.observers ? this.observers.length : 0;
+    if (!Listener.sources) {
+      Listener.sources = [this];
+      Listener.sourceSlots = [sSlot];
+    } else {
+      Listener.sources.push(this);
+      Listener.sourceSlots.push(sSlot);
+    }
+    if (!this.observers) {
+      this.observers = [Listener];
+      this.observerSlots = [Listener.sources.length - 1];
+    } else {
+      this.observers.push(Listener);
+      this.observerSlots.push(Listener.sources.length - 1);
+    }
+  }
+  return this.value;
+}
 function writeSignal(node, value, isComp) {
   let current = node.value;
   if (!node.comparator || !node.comparator(current, value)) {
     node.value = value;
     if (node.observers && node.observers.length) {
@@ -276,12 +320,112 @@ function castError(err) {
 }
 function handleError(err, owner = Owner) {
   const error = castError(err);
   throw error;
 }
-function createComponent(Comp, props) {
-  return untrack(() => Comp(props || {}));
+function createComponent(Comp, props2) {
+  return untrack(() => Comp(props2 || {}));
+}
+function trueFn() {
+  return true;
+}
+const propTraps = {
+  get(_, property, receiver) {
+    if (property === $PROXY) return receiver;
+    return _.get(property);
+  },
+  has(_, property) {
+    if (property === $PROXY) return true;
+    return _.has(property);
+  },
+  set: trueFn,
+  deleteProperty: trueFn,
+  getOwnPropertyDescriptor(_, property) {
+    return {
+      configurable: true,
+      enumerable: true,
+      get() {
+        return _.get(property);
+      },
+      set: trueFn,
+      deleteProperty: trueFn
+    };
+  },
+  ownKeys(_) {
+    return _.keys();
+  }
+};
+function resolveSource(s) {
+  return !(s = typeof s === "function" ? s() : s) ? {} : s;
+}
+function resolveSources() {
+  for (let i = 0, length = this.length; i < length; ++i) {
+    const v = this[i]();
+    if (v !== void 0) return v;
+  }
+}
+function mergeProps(...sources) {
+  let proxy = false;
+  for (let i = 0; i < sources.length; i++) {
+    const s = sources[i];
+    proxy = proxy || !!s && $PROXY in s;
+    sources[i] = typeof s === "function" ? (proxy = true, createMemo(s)) : s;
+  }
+  if (SUPPORTS_PROXY && proxy) {
+    return new Proxy({
+      get(property) {
+        for (let i = sources.length - 1; i >= 0; i--) {
+          const v = resolveSource(sources[i])[property];
+          if (v !== void 0) return v;
+        }
+      },
+      has(property) {
+        for (let i = sources.length - 1; i >= 0; i--) {
+          if (property in resolveSource(sources[i])) return true;
+        }
+        return false;
+      },
+      keys() {
+        const keys = [];
+        for (let i = 0; i < sources.length; i++) keys.push(...Object.keys(resolveSource(sources[i])));
+        return [...new Set(keys)];
+      }
+    }, propTraps);
+  }
+  const sourcesMap = {};
+  const defined = /* @__PURE__ */ Object.create(null);
+  for (let i = sources.length - 1; i >= 0; i--) {
+    const source = sources[i];
+    if (!source) continue;
+    const sourceKeys = Object.getOwnPropertyNames(source);
+    for (let i2 = sourceKeys.length - 1; i2 >= 0; i2--) {
+      const key = sourceKeys[i2];
+      if (key === "__proto__" || key === "constructor") continue;
+      const desc = Object.getOwnPropertyDescriptor(source, key);
+      if (!defined[key]) {
+        defined[key] = desc.get ? {
+          enumerable: true,
+          configurable: true,
+          get: resolveSources.bind(sourcesMap[key] = [desc.get.bind(source)])
+        } : desc.value !== void 0 ? desc : void 0;
+      } else {
+        const sources2 = sourcesMap[key];
+        if (sources2) {
+          if (desc.get) sources2.push(desc.get.bind(source));
+          else if (desc.value !== void 0) sources2.push(() => desc.value);
+        }
+      }
+    }
+  }
+  const target = {};
+  const definedKeys = Object.keys(defined);
+  for (let i = definedKeys.length - 1; i >= 0; i--) {
+    const key = definedKeys[i], desc = defined[key];
+    if (desc && desc.get) Object.defineProperty(target, key, desc);
+    else target[key] = desc ? desc.value : void 0;
+  }
+  return target;
 }
 const booleans = ["allowfullscreen", "async", "autofocus", "autoplay", "checked", "controls", "default", "disabled", "formnovalidate", "hidden", "indeterminate", "inert", "ismap", "loop", "multiple", "muted", "nomodule", "novalidate", "open", "playsinline", "readonly", "required", "reversed", "seamless", "selected"];
 const Properties = /* @__PURE__ */ new Set(["className", "value", "readOnly", "noValidate", "formNoValidate", "isMap", "noModule", "playsInline", ...booleans]);
 const ChildProperties = /* @__PURE__ */ new Set(["innerHTML", "textContent", "innerText", "children"]);
 const Aliases = /* @__PURE__ */ Object.assign(/* @__PURE__ */ Object.create(null), {
@@ -461,48 +605,48 @@ function style(node, value, prev) {
       prev[s] = v;
     }
   }
   return prev;
 }
-function spread(node, props = {}, isSVG, skipChildren) {
+function spread(node, props2 = {}, isSVG, skipChildren) {
   const prevProps = {};
-  createRenderEffect(() => typeof props.ref === "function" && use(props.ref, node));
-  createRenderEffect(() => assign(node, props, isSVG, true, prevProps, true));
+  createRenderEffect(() => typeof props2.ref === "function" && use(props2.ref, node));
+  createRenderEffect(() => assign(node, props2, isSVG, true, prevProps, true));
   return prevProps;
 }
 function use(fn, element, arg) {
   return untrack(() => fn(element, arg));
 }
 function insert(parent, accessor, marker, initial) {
   if (marker !== void 0 && !initial) initial = [];
   if (typeof accessor !== "function") return insertExpression(parent, accessor, initial, marker);
   createRenderEffect((current) => insertExpression(parent, accessor(), current, marker), initial);
 }
-function assign(node, props, isSVG, skipChildren, prevProps = {}, skipRef = false) {
-  props || (props = {});
+function assign(node, props2, isSVG, skipChildren, prevProps = {}, skipRef = false) {
+  props2 || (props2 = {});
   for (const prop in prevProps) {
-    if (!(prop in props)) {
+    if (!(prop in props2)) {
       if (prop === "children") continue;
-      prevProps[prop] = assignProp(node, prop, null, prevProps[prop], isSVG, skipRef, props);
+      prevProps[prop] = assignProp(node, prop, null, prevProps[prop], isSVG, skipRef, props2);
     }
   }
-  for (const prop in props) {
+  for (const prop in props2) {
     if (prop === "children") {
       continue;
     }
-    const value = props[prop];
-    prevProps[prop] = assignProp(node, prop, value, prevProps[prop], isSVG, skipRef, props);
+    const value = props2[prop];
+    prevProps[prop] = assignProp(node, prop, value, prevProps[prop], isSVG, skipRef, props2);
   }
 }
 function toPropertyName(name) {
   return name.toLowerCase().replace(/-([a-z])/g, (_, w) => w.toUpperCase());
 }
 function toggleClassKey(node, key, value) {
   const classNames = key.trim().split(/\s+/);
   for (let i = 0, nameLen = classNames.length; i < nameLen; i++) node.classList.toggle(classNames[i], value);
 }
-function assignProp(node, prop, value, prev, isSVG, skipRef, props) {
+function assignProp(node, prop, value, prev, isSVG, skipRef, props2) {
   let isCE, isProp, isChildProp, propAlias, forceProp;
   if (prop === "style") return style(node, value, prev);
   if (prop === "classList") return classList(node, value, prev);
   if (value === prev) return prev;
   if (prop === "ref") {
@@ -528,11 +672,11 @@ function assignProp(node, prop, value, prev, isSVG, skipRef, props) {
     }
   } else if (prop.slice(0, 5) === "attr:") {
     setAttribute(node, prop.slice(5), value);
   } else if (prop.slice(0, 5) === "bool:") {
     setBoolAttribute(node, prop.slice(5), value);
-  } else if ((forceProp = prop.slice(0, 5) === "prop:") || (isChildProp = ChildProperties.has(prop)) || ((propAlias = getPropAlias(prop, node.tagName)) || (isProp = Properties.has(prop))) || (isCE = node.nodeName.includes("-") || "is" in props)) {
+  } else if ((forceProp = prop.slice(0, 5) === "prop:") || (isChildProp = ChildProperties.has(prop)) || ((propAlias = getPropAlias(prop, node.tagName)) || (isProp = Properties.has(prop))) || (isCE = node.nodeName.includes("-") || "is" in props2)) {
     if (forceProp) {
       prop = prop.slice(5);
       isProp = true;
     }
     if (prop === "class" || prop === "className") className(node, value);
@@ -691,76 +835,299 @@ function cleanChildren(parent, current, marker, replacement) {
       } else inserted = true;
     }
   } else parent.insertBefore(node, marker);
   return [node];
 }
+var styleq = {};
+var hasRequiredStyleq;
+function requireStyleq() {
+  if (hasRequiredStyleq) return styleq;
+  hasRequiredStyleq = 1;
+  Object.defineProperty(styleq, "__esModule", {
+    value: true
+  });
+  styleq.styleq = void 0;
+  var cache = /* @__PURE__ */ new WeakMap();
+  var compiledKey = "$$css";
+  function createStyleq(options) {
+    var disableCache;
+    var disableMix;
+    var transform;
+    if (options != null) {
+      disableCache = options.disableCache === true;
+      disableMix = options.disableMix === true;
+      transform = options.transform;
+    }
+    return function styleq2() {
+      var definedProperties = [];
+      var className2 = "";
+      var inlineStyle = null;
+      var debugString = "";
+      var nextCache = disableCache ? null : cache;
+      var styles2 = new Array(arguments.length);
+      for (var i = 0; i < arguments.length; i++) {
+        styles2[i] = arguments[i];
+      }
+      while (styles2.length > 0) {
+        var possibleStyle = styles2.pop();
+        if (possibleStyle == null || possibleStyle === false) {
+          continue;
+        }
+        if (Array.isArray(possibleStyle)) {
+          for (var _i = 0; _i < possibleStyle.length; _i++) {
+            styles2.push(possibleStyle[_i]);
+          }
+          continue;
+        }
+        var style2 = transform != null ? transform(possibleStyle) : possibleStyle;
+        if (style2.$$css != null) {
+          var classNameChunk = "";
+          if (nextCache != null && nextCache.has(style2)) {
+            var cacheEntry = nextCache.get(style2);
+            if (cacheEntry != null) {
+              classNameChunk = cacheEntry[0];
+              debugString = cacheEntry[2];
+              definedProperties.push.apply(definedProperties, cacheEntry[1]);
+              nextCache = cacheEntry[3];
+            }
+          } else {
+            var definedPropertiesChunk = [];
+            for (var prop in style2) {
+              var value = style2[prop];
+              if (prop === compiledKey) {
+                var compiledKeyValue = style2[prop];
+                if (compiledKeyValue !== true) {
+                  debugString = debugString ? compiledKeyValue + "; " + debugString : compiledKeyValue;
+                }
+                continue;
+              }
+              if (typeof value === "string" || value === null) {
+                if (!definedProperties.includes(prop)) {
+                  definedProperties.push(prop);
+                  if (nextCache != null) {
+                    definedPropertiesChunk.push(prop);
+                  }
+                  if (typeof value === "string") {
+                    classNameChunk += classNameChunk ? " " + value : value;
+                  }
+                }
+              } else {
+                console.error("styleq: ".concat(prop, " typeof ").concat(String(value), ' is not "string" or "null".'));
+              }
+            }
+            if (nextCache != null) {
+              var weakMap = /* @__PURE__ */ new WeakMap();
+              nextCache.set(style2, [classNameChunk, definedPropertiesChunk, debugString, weakMap]);
+              nextCache = weakMap;
+            }
+          }
+          if (classNameChunk) {
+            className2 = className2 ? classNameChunk + " " + className2 : classNameChunk;
+          }
+        } else {
+          if (disableMix) {
+            if (inlineStyle == null) {
+              inlineStyle = {};
+            }
+            inlineStyle = Object.assign({}, style2, inlineStyle);
+          } else {
+            var subStyle = null;
+            for (var _prop in style2) {
+              var _value = style2[_prop];
+              if (_value !== void 0) {
+                if (!definedProperties.includes(_prop)) {
+                  if (_value != null) {
+                    if (inlineStyle == null) {
+                      inlineStyle = {};
+                    }
+                    if (subStyle == null) {
+                      subStyle = {};
+                    }
+                    subStyle[_prop] = _value;
+                  }
+                  definedProperties.push(_prop);
+                  nextCache = null;
+                }
+              }
+            }
+            if (subStyle != null) {
+              inlineStyle = Object.assign(subStyle, inlineStyle);
+            }
+          }
+        }
+      }
+      var styleProps = [className2, inlineStyle, debugString];
+      return styleProps;
+    };
+  }
+  var styleq$1 = styleq.styleq = createStyleq();
+  styleq$1.factory = createStyleq;
+  return styleq;
+}
+var styleqExports = /* @__PURE__ */ requireStyleq();
+function props() {
+  for (var _len = arguments.length, styles2 = new Array(_len), _key = 0; _key < _len; _key++) {
+    styles2[_key] = arguments[_key];
+  }
+  const [className2, style2, dataStyleSrc] = styleqExports.styleq(styles2);
+  const result = {};
+  if (className2 != null && className2 !== "") {
+    result.className = className2;
+  }
+  if (style2 != null && Object.keys(style2).length > 0) {
+    result.style = style2;
+  }
+  if (dataStyleSrc != null && dataStyleSrc !== "") {
+    result["data-style-src"] = dataStyleSrc;
+  }
+  return result;
+}
+function sx(...styles2) {
+  return attrs(props(...styles2));
+}
+function attrs({
+  className: className2,
+  "data-style-src": dataStyleSrc,
+  style: style2
+}) {
+  const result = {};
+  if (className2 != null && className2 !== "") {
+    result.class = className2;
+  }
+  if (style2 != null && Object.keys(style2).length > 0) {
+    result.style = Object.keys(style2).map((key) => `${key}:${style2[key]};`).join("");
+  }
+  if (dataStyleSrc != null && dataStyleSrc !== "") {
+    result["data-style-src"] = dataStyleSrc;
+  }
+  return result;
+}
 var _tmpl$$3 = /* @__PURE__ */ template(`<div>component from ../alias`);
+const styles$3 = {
+  base: {
+    kC7eKd: "xi9vgnp",
+    ku1ltF: null,
+    kHypHr: null,
+    kWkggS: null,
+    kKwaWg: null,
+    kl9DO0: null,
+    k1YJky: null,
+    kDwdnV: null,
+    kAiEkY: null,
+    kz484i: null,
+    kgSjnq: null,
+    kMwMTN: "xkkx6qt",
+    $$css: true
+  }
+};
 function AliasComponent() {
   return (() => {
     var _el$ = _tmpl$$3();
-    spread(_el$, {
-      className: "xi9vgnp x19xla2n"
-    }, false);
+    spread(_el$, mergeProps(() => sx(styles$3.base)), false);
     return _el$;
   })();
 }
 const aliasConst = "alias OK";
 var _tmpl$$2 = /* @__PURE__ */ template(`<div>component from ../workspace`);
+const styles$2 = {
+  base: {
+    kC7eKd: "xi9vgnp",
+    ku1ltF: null,
+    kHypHr: null,
+    kWkggS: null,
+    kKwaWg: null,
+    kl9DO0: null,
+    k1YJky: null,
+    kDwdnV: null,
+    kAiEkY: null,
+    kz484i: null,
+    kgSjnq: null,
+    kMwMTN: "xymhn1y",
+    $$css: true
+  }
+};
 function WorkspaceComponent() {
   return (() => {
     var _el$ = _tmpl$$2();
-    spread(_el$, {
-      className: "xi9vgnp xymhn1y"
-    }, false);
+    spread(_el$, mergeProps(() => sx(styles$2.base)), false);
     return _el$;
   })();
 }
 const workspaceConst = "workspace OK";
 var _tmpl$$1 = /* @__PURE__ */ template(`<div>component from ../outside`);
+const styles$1 = {
+  base: {
+    kC7eKd: "xi9vgnp",
+    ku1ltF: null,
+    kHypHr: null,
+    kWkggS: null,
+    kKwaWg: null,
+    kl9DO0: null,
+    k1YJky: null,
+    kDwdnV: null,
+    kAiEkY: null,
+    kz484i: null,
+    kgSjnq: null,
+    kMwMTN: "xv98f2o",
+    $$css: true
+  }
+};
 function OutsideComponent() {
   return (() => {
     var _el$ = _tmpl$$1();
-    spread(_el$, {
-      className: "xi9vgnp xgrr9y0"
-    }, false);
+    spread(_el$, mergeProps(() => sx(styles$1.base)), false);
     return _el$;
   })();
 }
 const outsideConst = "outside OK";
 var _tmpl$ = /* @__PURE__ */ template(`<div><div>stylex local value - red if works</div><div>stylex var from ./src/ - green if works</div><div>stylex var from ../ - orange if works</div><div>JS var from ../outside: </div><div>stylex var from ../outside - purple if works</div><div>JS var from @alias: </div><div>stylex var from @alias - blue if works</div><div>JS var from @repo/workspace: </div><div>stylex var from @repo/workspace - magenta if works`);
+const styles = {
+  local: {
+    kMwMTN: "x1e2nbdu",
+    $$css: true
+  },
+  srcVar: {
+    kMwMTN: "x1lm2guq",
+    $$css: true
+  },
+  projVar: {
+    kMwMTN: "x1sst8ud",
+    $$css: true
+  },
+  outsideVar: {
+    kMwMTN: "xv98f2o",
+    $$css: true
+  },
+  aliasVar: {
+    kMwMTN: "xkkx6qt",
+    $$css: true
+  },
+  workspaceVar: {
+    kMwMTN: "xymhn1y",
+    $$css: true
+  }
+};
 function App() {
   return (() => {
     var _el$ = _tmpl$(), _el$2 = _el$.firstChild, _el$3 = _el$2.nextSibling, _el$4 = _el$3.nextSibling, _el$5 = _el$4.nextSibling;
     _el$5.firstChild;
     var _el$7 = _el$5.nextSibling, _el$8 = _el$7.nextSibling;
     _el$8.firstChild;
     var _el$0 = _el$8.nextSibling, _el$1 = _el$0.nextSibling;
     _el$1.firstChild;
     var _el$11 = _el$1.nextSibling;
-    spread(_el$2, {
-      className: "x1e2nbdu"
-    }, false);
-    spread(_el$3, {
-      className: "x1lm2guq"
-    }, false);
-    spread(_el$4, {
-      className: "x1sst8ud"
-    }, false);
+    spread(_el$2, mergeProps(() => sx(styles.local)), false);
+    spread(_el$3, mergeProps(() => sx(styles.srcVar)), false);
+    spread(_el$4, mergeProps(() => sx(styles.projVar)), false);
     insert(_el$5, outsideConst, null);
-    spread(_el$7, {
-      className: "xgrr9y0"
-    }, false);
+    spread(_el$7, mergeProps(() => sx(styles.outsideVar)), false);
     insert(_el$, createComponent(OutsideComponent, {}), _el$8);
     insert(_el$8, aliasConst, null);
-    spread(_el$0, {
-      className: "x19xla2n"
-    }, false);
+    spread(_el$0, mergeProps(() => sx(styles.aliasVar)), false);
     insert(_el$, createComponent(AliasComponent, {}), _el$1);
     insert(_el$1, workspaceConst, null);
-    spread(_el$11, {
-      className: "xymhn1y"
-    }, false);
+    spread(_el$11, mergeProps(() => sx(styles.workspaceVar)), false);
     insert(_el$, createComponent(WorkspaceComponent, {}), null);
     return _el$;
   })();
 }
 const root = document.getElementById("root");
```

hyperknot:
I'm using this file for the conversion:

https://github.com/hyperknot/bug_repro/blob/main/stylex_solid_attr_performance/alias/styleUtils.ts

Replacing sx with the long form `attrs(stylex.props(...styles))` does help one level, but it still doesn't get statically compiled into classnames. I mean this is the whole promise of StyleX, that you get compiled performance in static cases. Now it's gone for attr, or at least I don't know how to get it back.

necolas:
If I modify one of the StyleX compiler tests, it gets compiled as expected:
 
image

We always extract to static CSS, even when the runtime is used to handle merging of classes across component boundaries.

hyperknot:
I don't understand the `import attrs from 'attrs'` line, where is it from?

This is the file I'm using, basically identical to the on in docs + TS types added. This one is not recognised by babel.
https://github.com/hyperknot/bug_repro/blob/main/stylex_solid_attr_performance/alias/styleUtils.ts

necolas:
> I don't understand the import attrs from 'attrs' line, where is it from?

It's just an example of importing the wrapper function described in the docs. You always have to have the stylex.props function in the file being transformed by babel.

hyperknot:
@necolas thank you for your answer. I've created a minimal repro case here:
https://github.com/hyperknot/bug_repro/tree/main/stylex_solid_attr

Can you have a look at it?

This is what the output is:
image

I added the 0.12's old .attrs function and it seems that it wasn't compiled out either. So it might have been broken the whole time.

Is there like a magic string matching in the babel plugin, which is looking for `stylex.props` especially?

Question 2: What is the point of the 
```
  // Convert style object to a style string
  if (style != null && Object.keys(style).length > 0) {
    result.style = Object.keys(style)
      .map((key) => `${key}:${style[key as keyof typeof style]};`)
      .join('')
  }
  ```
  
I don't understand in which case it is used. I thought stylex only works with compiled styles from the output of `stylex.create`, so I don't see how the style object can be passed?

**Comment by hyperknot** [comment]
I've literally spent days ironing out the Vite support for React and Solid, and this is the last step. If you could help mefix this, I'd make a PR to include an official example for Vite + React and Vite Solid. It's only these attrs which are blocking it.

**Comment by necolas** [comment]
> So it might have been broken the whole time.

Nothing changed in terms of functionality in 0.13, but non-React users have to use `props` now and create their own runtime helper.

https://stylexjs.com/docs/api/javascript/props/

> Is there like a magic string matching in the babel plugin, which is looking for stylex.props especially?

The plugin looks for StyleX imports, based on the `importSources` configuration, and then processes all the functions in the API.

The `props` function is the only one that also has a runtime, because StyleX let's you efficiently merge classes and inline styles at runtime even though they've been compiled. This feature is what makes StyleX special. But if StyleX can resolve all the styles being passed to `props` at build time, it will add one extra optimization and transform them to the final class string. It can't do that if you wrap that function and import it from another file, for example.

> I thought stylex only works with compiled styles from the output of stylex.create, so I don't see how the style object can be passed?

It supports dynamic styles too 

https://stylexjs.com/docs/learn/styling-ui/defining-styles/#dynamic-styles

**Comment by hyperknot** [comment]
Thank you for your answer. It's exactly that one extra step of optimisation I'd like to do.

> So it might have been broken the whole time.

By this, I was referring to the fact, even under 0.12, the attrs was not optimised, as it resulted in mergeProps. I understand that the test show something different, but in practice, it generated mergeProps.

I definitely want to avoid mergeProps, as that's one of the biggest promise of StyleX is that for static styling there is no runtime overhead.

So what I'm trying to achieve is to find a solution for non-React frameworks, which doesn't result in mergeProps for static styles.

Can you tell me how can I do this? Are there any possible solutions as of 0.14 which can avoid generating mergeProps for static styles?

**Comment by necolas** [comment]
I don't know where mergeProps is coming from but it's not StyleX. You can see that the StyleX code is transformed straight into classes when using the pattern mentioned in the docs.

**Comment by hyperknot** [comment]
Thank you, that was great help!

This is what the StyleX babel step creates under 0.12. It's running first.

```
const styles = {
  local: {
    kMwMTN: "x1e2nbdu",
    kGuDYH: "x1i1m3gp",
    kybGjl: "x1bvjpef",
    k1TLXF: null,
    kMnn75: null,
    kmVMDM: null,
    kNySMw: null,
    $$css: true
  },
  srcVar: {
    kMwMTN: "x1lm2guq",
    $$css: true
  },
  projVar: {
    kMwMTN: "x1sst8ud",
    $$css: true
  },
  dynamic: fontSize => [{
    kGuDYH: "x6zurak",
    kMwMTN: "xju2f9n",
    kybGjl: null,
    k1TLXF: null,
    kMnn75: null,
    kmVMDM: null,
    kNySMw: null,
    $$css: true
  }, {
    "--fontSize": (val => typeof val === "number" ? val + "px" : val != null ? val : undefined)(fontSize)
  }]
};
function App() {
  return <div>
      <div {...{
      className: "x1e2nbdu x1i1m3gp x1bvjpef"
    }}>props local value - red if works</div>
      <div {...{
      className: "x1lm2guq"
    }}>props var from ./src/ - green if works</div>
      <div {...{
      className: "x1sst8ud"
    }}>props var from ../ - orange if works</div>
      <div {...stylex.props(styles.local, styles.dynamic('14px'))}>
        props dynamic - 14px blue underlined if works
      </div>

      <div {...stylex.attrs(styles.local)}>
        native attr (only works on 0.12) local value - red if works
      </div>
      <div {...stylex.attrs(styles.srcVar)}>
        native attr (only works on 0.12) var from ./src/ - green if works
      </div>
      <div {...stylex.attrs(styles.projVar)}>
        native attr (only works on 0.12) var from ../ - orange if works
      </div>
      <div {...stylex.attrs(styles.local, styles.dynamic('14px'))}>
        native attr (only works on 0.12) dynamic - 14px blue underlined if works
      </div>
    </div>;
}
export default App;
```

As you can see, the `...stylex.attrs` is not transformed.

Maybe I'm doing something wrong on the Babel setup?

This is my Babel plugin:

```
function stylexPlugin(): Plugin {
  const filter = createFilter('**/*.{js,ts,jsx,tsx}', 'node_modules/**')

  return {
    name: 'stylex-js-transform',
    enforce: 'pre',
    async transform(code: string, id: string) {
      if (!filter(id)) return null

      // skip files without StyleX usage
      if (!code.includes('stylex')) {
        return null
      }

      const result = await babel.transformAsync(code, {
        ...stylexBabelConfig,
        filename: id,
        sourceMaps: true,
      })
      console.log(id)
      console.log(result?.code)

      return result?.code ? { code: result.code, map: result.map } : null
    },
  }
}
```

hyperknot:
OK, I figured out, I didn't downgrade the babel plugin, just the main stylex package! When downgraded all the packages to 0.12 it works correctly.


Now, please help me figure out how to avoid this regression in 0.13+. I really is a regression, this really is the 1st statement on the homepage of StyleX, it is the reason why many of use have invested countless hours into StyleX.

The biggest claim, the very first statement of the homepage is not true anymore for non-React users!

image

necolas:
There's no regression to how `props` is compiled. For non-React use, please refer to this https://stylexjs.com/docs/api/javascript/props/#not-using-react

hyperknot:
There is a huge performance regression in the whole library for non-React use.

I'll open a separate issue about it.

hyperknot:
I've opened this issue. https://github.com/facebook/stylex/issues/1112

PR #976, opened 2025-04-08
Replace $Rest in stylex

SamChou19815:
## What changed / motivation ?

The Flow team is looking into killing $Rest now that Omit exists for a long time. This PR replaces these in stylex.

## Linked PR/Issues

Fixes # (issue)

## Additional Context

Screenshots, Tests, Anything Else

## Pre-flight checklist

- [x] I have read the contributing guidelines
      [Contribution Guidelines](https://github.com/facebook/stylex/blob/main/.github/CONTRIBUTING.md)
- [x] Performed a self-review of my code

PR #994, opened 2025-04-11
Replace $Diff in StyleXTypes

SamChou19815:

The Flow team is looking into killing `$Diff` now that Omit exists for a long time. This PR replaces the final `$Diff` in `StyleXTypes` with `Omit`. This was previously replaced from `$Rest` in #976 because some internal code is relying on weird behavior of `$Diff`, but they have been taken care of now so we can move away from it.

## Pre-flight checklist


nmn (approved):

mellyeliu:
closing for https://github.com/facebook/stylex/pull/995

PR #995, opened 2025-04-11
Replace $Diff in StyleXTypes

necolas:
The Flow team is looking into killing `$Diff` now that Omit exists for a long time. This PR replaces the final `$Diff` in `StyleXTypes` with `Omit`. This was previously replaced from `$Rest` in #976 because some internal code is relying on weird behavior of `$Diff`, but they have been taken care of now so we can move away from it.

Close #994


necolas:
Fixes code formatting issue in https://github.com/facebook/stylex/pull/994

mellyeliu (approved):

PR #997, opened 2025-04-15
style-value-parser/feat: add parser support for more types of media queries

mellyeliu:
Lots of edge cases to fix:
- ~~support `only` keyword~~
- ~~calc parsing broken with certain parentheses~~
     - ~~implement `toString` in `Calc` parser class~~
- ~~duplicate not parsing broken~~
- ~~strange nested parentheses edge cases~~
- ~~refactor some width -> max-width logic~~
- ...to add as I test

Testing:
- so many tests! will refactor in a follow up PR
- added a number of invalid media query assertions
- i'm almost certain i'm missing some edge cases, but I think this covers 95+% of use cases and most of what we'll see practically. will continue to iterate

refs: [mdn media query docs](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries) 
[lightning css playground](https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22%40custom-media%20--modern%20(color)%2C%20(hover)%3B%5Cn%5Cn%5Cn%40custom-media%20--%20(color)%2C%20(hover)%3B%5Cn%5Cn.foo%20%7B%5Cn%20%20background%3A%20yellow%3B%5Cn%5Cn%20%20-webkit-border-radius%3A%202px%3B%5Cn%20%20-moz-border-radius%3A%202px%3B%5Cn%20%20border-radius%3A%202px%3B%5Cn%5Cn%20%20-webkit-transition%3A%20background%20200ms%3B%5Cn%20%20-moz-transition%3A%20background%20200ms%3B%5Cn%20%20transition%3A%20background%20200ms%3B%5Cn%5Cn%20%20%26.bar%20%7B%5Cn%20%20%20%20color%3A%20green%3B%5Cn%20%20%7D%5Cn%7D%5Cn%5Cn%5Cn%5Cn%40media%20(min-width%3A%20500px)%20or%20(max-width%3A%20600px)%20and%20not%20(max-width%3A%2000px)%5Cn%20%7B%5Cn%20%20.a%20%7B%5Cn%20%20%20%20color%3A%20green%3B%5Cn%20%20%7D%5Cn%7D%5Cn%5Cn%40media%20print%20and%20((max-width%3A%20768px)%20or%20(orientation%3A%20portrait))%20%7B%5Cn%20%20.a%20%7B%5Cn%20%20%20%20color%3A%20green%3B%5Cn%20%20%7D%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D) to double check syntax

### DIFF
--- a/packages/style-value-parser/src/at-queries/__tests__/media-query-test.js
+++ b/packages/style-value-parser/src/at-queries/__tests__/media-query-test.js
@@ -1035,17 +1035,32 @@ describe('Test CSS Type: @media queries', () => {
     expect(MediaQuery.parser.parseToEnd(query).toString()).toEqual(query);
   });
 
-
-  test.skip('@media (min-width: calc(300px + 5em))', () => {
-    expect(
-      MediaQuery.parser.parseToEnd('@media (min-width: calc(300px + 5em))'),
-    ).toMatchInlineSnapshot();
+  test('@media (min-width: calc(300px + 5em))', () => {
+    const query = '@media (min-width: calc(300px + 5em))';
+    expect(MediaQuery.parser.parseToEnd(query)).toMatchInlineSnapshot(`
+      MediaQuery {
+        "queries": {
+          "key": "min-width",
+          "type": "pair",
+          "value": "calc(300px + 5em)",
+        },
+      }
+    `);
+    expect(MediaQuery.parser.parseToEnd(query).toString()).toEqual(query);
   });
 
-  test.skip('@media (max-height: calc(100vh - 50px))', () => {
-    expect(
-      MediaQuery.parser.parseToEnd('@media (max-height: calc(100vh - 50px))'),
-    ).toMatchInlineSnapshot();
+  test('@media (max-height: calc(100vh - 50px))', () => {
+    const query = '@media (max-height: calc(100vh - 50px))';
+    expect(MediaQuery.parser.parseToEnd(query)).toMatchInlineSnapshot(`
+      MediaQuery {
+        "queries": {
+          "key": "max-height",
+          "type": "pair",
+          "value": "calc(100vh - 50px)",
+        },
+      }
+    `);
+    expect(MediaQuery.parser.parseToEnd(query).toString()).toEqual(query);
   });
 
   test('@media (aspect-ratio: 16 / 9)', () => {
--- a/packages/style-value-parser/src/at-queries/__tests__/media-query-test.js
+++ b/packages/style-value-parser/src/at-queries/__tests__/media-query-test.js
@@ -1891,8 +1891,7 @@ describe('Test CSS Type: @media queries', () => {
     expect(MediaQuery.parser.parseToEnd(query).toString()).toEqual(query);
   });
 
-  // nested not queries are broken
-  test.skip('@media not (not (not (min-width: 400px)))', () => {
+  test('@media not (not (not (min-width: 400px)))', () => {
     const query = '@media not (not (not (min-width: 400px)))';
     const mediaQuery = MediaQuery.parser.parseToEnd(query);
     expect(mediaQuery).toMatchInlineSnapshot(`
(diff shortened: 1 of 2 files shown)


packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-test.js
```diff
@@ -1874,7 +1892,7 @@ describe('Test CSS Type: @media queries', () => {
   });
 
   // nested not queries are broken
```
necolas, line 1894:
```suggestion
```

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-test.js
```diff
@@ -1028,22 +1035,32 @@ describe('Test CSS Type: @media queries', () => {
...
-    expect(
-      MediaQuery.parser.parseToEnd('@media (width: calc(100% - 20px))'),
-    ).toMatchInlineSnapshot();
-  });
-
-  test.skip('@media (min-width: calc(300px + 5em))', () => {
-    expect(
-      MediaQuery.parser.parseToEnd('@media (min-width: calc(300px + 5em))'),
-    ).toMatchInlineSnapshot();
+  test('@media (min-width: calc(300px + 5em))', () => {
```
necolas, line 1038: Do we have any tests for when this MQ is malformed with no spaces around the operator within `calc`?

mellyeliu:
yeah, the original intention was to build it reasonably close to the mdn spec so we have a parser approximating standard media query support, and we can limit what we want to include from there. for instance, I'm not sure if we actually want to support media query with calc (not sure how much use cases there are) but maybe we can parameterize it for flexibility moving forward

nmn:
The parser is meant to help with many different things, so one of the things it might do is improve the ESLint plugin. There having a fully spec-compliant parser would mean the ability to offer auto-fixes.

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-test.js
```diff
@@ -18,6 +18,7 @@ describe('Test CSS Type: @media queries', () => {
         "queries": {
           "key": "screen",
           "not": false,
+          "only": false,
```
nmn, line 21: I don't think it's valid to use `only` on any random query, and I think it can only exist on the `@media` level of the query.

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-test.js
```diff
@@ -1028,22 +1035,32 @@ describe('Test CSS Type: @media queries', () => {
...
+    expect(MediaQuery.parser.parseToEnd(query)).toMatchInlineSnapshot(`
+      MediaQuery {
+        "queries": {
+          "key": "min-width",
+          "type": "pair",
+          "value": "calc(300px + 5em)",
+        },
+      }
+    `);
+    expect(MediaQuery.parser.parseToEnd(query).toString()).toEqual(query);
```
nmn, line 1049: I also think it's better to use snapshot tests instead of `toEqual(query)` as a pattern for these tests.

The parser, by design, will remove extra whitespace etc. It's job is to produce equivalent strings, not necessarily the same strings as the input. Instead of picking and choosing between the two patterns, it's better to use the safer pattern everywhere.

packages/style-value-parser/src/at-queries-from-tokens/media-query.js
```diff
@@ -277,35 +295,103 @@ const mediaOrRulesParser: TokenParser<MediaOrRules | MediaQueryRule> =
...
+          ) {
+            const dim = queries.value as Length;
+            valueStr = `${dim.value}${dim.unit}`;
+          } else {
+            valueStr = candidate;
+          }
+        } else {
+          valueStr = '';
+        }
         return `(${queries.key}: ${valueStr})`;
```
nmn, line 390: This long chain can be simplified with early returns.

```suggestion
        if (Array.isArray(queries.value)) {
          return `(${queries.key}: ${queries.value[0]} / ${queries.value[2]})`;
        }
        if (typeof queries.value === 'string') {
          return `(${queries.key}: ${queries.value})`;
        } 
        if (
          typeof queries.value === 'object' &&
          queries.value !== null &&
          typeof queries.value.toString === 'function'
        ) {
          const candidate = queries.value.toString();
          if (
            candidate === '[object Object]' &&
            typeof queries.value !== 'string' &&
            !Array.isArray(queries.value)
          ) {
            const dim = queries.value as Length;
            return `(${queries.key}: ${dim.value}${dim.unit})`;
          }
          return `(${queries.key}: ${candidate})`;
        } 
        return `(${queries.key}: )`;
```

This reveals a problem with the last `else` case where `(${queries.key}: )` doesn't seem like a valid thing to return.

packages/style-value-parser/src/at-queries-from-tokens/media-query.js
```diff
@@ -277,35 +295,103 @@ const mediaOrRulesParser: TokenParser<MediaOrRules | MediaQueryRule> =
...
+        } else if (typeof queries.value === 'string') {
+          valueStr = queries.value;
+        } else if (
+          typeof queries.value === 'object' &&
+          queries.value !== null &&
+          typeof queries.value.toString === 'function'
+        ) {
+          const candidate = queries.value.toString();
+          if (
+            candidate === '[object Object]' &&
```
nmn, line 378: `'[object Object]'` is also a bit sketchy.

packages/style-value-parser/src/at-queries-from-tokens/media-query.js
```diff
@@ -277,35 +295,103 @@ const mediaOrRulesParser: TokenParser<MediaOrRules | MediaQueryRule> =
...
+    () =>
+      mediaOrRulesParser.surroundedBy(
+        TokenParser.tokens.OpenParen,
+        TokenParser.tokens.CloseParen,
+      ),
+    () =>
+      mediaAndRulesParser.surroundedBy(
+        TokenParser.tokens.OpenParen,
+        TokenParser.tokens.CloseParen,
+      ),
```
nmn, line 334: Pretty sure that you also need to account for optional whitespace here.

nmn (changes requested):
Some issues need fixing.

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-test.js
mellyeliu: that's how it's currently implemented, `only` is only allowed at the top level of a query, not within params or within a nested clause. i'll add some tests to cover it

packages/style-value-parser/src/at-queries-from-tokens/__tests__/media-query-test.js
mellyeliu: makes sense, I agree. I'm going to do a larger refactor to organize these tests in the next PR, and will rewrite how we test toString as well

packages/style-value-parser/src/at-queries-from-tokens/media-query.js
mellyeliu: added a throw in that case, let's support this set of values first

nmn (approved):
I don't see any issues for now. We can iterate more if any issues arise.

PR #998, opened 2025-04-15
[babel-plugin] Refactor transform-polyfills-test

necolas:
Move tests for all CSS polyfills into 'transform-polyfills'. Snapshot the metadata of each transform.

The logical property/value tests are preserved in the `legacy` test directory for now. Almost all these tests are no longer needed, as there is no polyfill transform applied.

### DIFF
--- a/packages/babel-plugin/__tests__/legacy/transform-logical-properties-test.js
+++ b/packages/babel-plugin/__tests__/legacy/transform-logical-properties-test.js
@@ -10,7 +10,7 @@
 jest.autoMockOff();
 
 const { transformSync } = require('@babel/core');
-const stylexPlugin = require('../src/index');
+const stylexPlugin = require('../../src/index');
 
 function transform(source, opts = {}) {
   return transformSync(source, {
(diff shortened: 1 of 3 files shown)


packages/babel-plugin/__tests__/legacy/transform-logical-properties-test.js
```diff
@@ -632,170 +616,6 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
 
-    /**
-     * Non-standard properties
```
necolas, line 636: These got moved to the 'polyfills' test file. But the goal is to be able to delete these once there is appropriate compiler validation.

packages/babel-plugin/__tests__/legacy/transform-logical-values-test.js
```diff
@@ -128,74 +128,6 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
 
-    /**
-     * Non-standard values
```
necolas, line 132: These got moved to the 'polyfills' test file. But the goal is to be able to delete these once there is appropriate compiler validation.

mellyeliu (approved):

PR #1000, opened 2025-04-15
[babel-plugin] Move implementation detail tests

necolas:
Moves files that test implementation details rather than the behavior of the babel-plugin. These unit tests are now co-located with their code.


packages/babel-plugin/src/utils/evaluation-errors.js
```diff
@@ -33,8 +33,8 @@ Could not determine the property being accessed.
 `;
 
 export const IMPORT_PATH_RESOLUTION_ERROR = `Could not resolve the path to the imported file.
```
mellyeliu, line 35: can we use this error message in `packages/babel-plugin/__tests__/validation-create-test.js`?

mellyeliu (approved):

PR #1005, opened 2025-04-17
[wip] feat: style injection for `defineConsts` hot reloading

mellyeliu:
add style injection for dev mode hot reloading 

very wip... debugging.....

### DIFF
--- a/packages/babel-plugin/src/utils/state-manager.js
+++ b/packages/babel-plugin/src/utils/state-manager.js
@@ -643,8 +643,8 @@ export default class StateManager {
           | {
               constKey: string,
               constVal: string | number,
-              rtl?: string | null,
               ltr: string,
+              rtl?: string | null,
             }
         ),
         number,
@@ -684,17 +684,48 @@ export default class StateManager {
 
       this.injectImportInserted = injectName;
     }
-    for (const [_key, { ltr, rtl }, priority] of styles) {
+
+    for (const [_key, styleObj, priority] of styles) {
+      // build metadataProps as before…
+      const metadataProps: Array<t.ObjectProperty> = [];
+      if ((styleObj as any).constKey != null) {
+        metadataProps.push(
+          t.objectProperty(
+            t.identifier('constKey'),
+            t.stringLiteral((styleObj as any).constKey),
+          ),
+        );
+      }
+      if ((styleObj as any).constVal != null) {
+        const val =
+          typeof (styleObj as any).constVal === 'number'
+            ? t.numericLiteral((styleObj as any).constVal)
+            : t.stringLiteral((styleObj as any).constVal);
+        metadataProps.push(
+          t.objectProperty(t.identifier('constVal'), val),
+        );
+      }
+
+      const args: Array<t.Expression> = [
+        // 1) ltr
+        t.stringLiteral(styleObj.ltr),
+        // 2) priority
+        t.numericLiteral(priority),
+        // 3) rtl OR explicit null
+        styleObj.rtl != null
+          ? t.stringLiteral(styleObj.rtl)
+          : t.nullLiteral(),
+        // 4) metadata object, only if we collected props
+        ...(metadataProps.length > 0
+          ? [t.objectExpression(metadataProps)]
+          : []),
+      ];
+
       statementPath.insertBefore(
-        t.expressionStatement(
-          t.callExpression(injectName, [
-            t.stringLiteral(ltr),
-            t.numericLiteral(priority),
-            ...(rtl != null ? [t.stringLiteral(rtl)] : []),
-          ]),
-        ),
+        t.expressionStatement(t.callExpression(injectName, args)),
       );
     }
+
   }
 
   markComposedNamespace(
--- a/packages/stylex/src/StyleXSheet.js
+++ b/packages/stylex/src/StyleXSheet.js
@@ -91,12 +91,21 @@ export class StyleXSheet {
     this.injected = false;
     this.ruleForPriority = new Map();
     this.rules = [];
+    this.constantsCache = new Map();
 
     this.rootTheme = opts.rootTheme;
     this.rootDarkTheme = opts.rootDarkTheme;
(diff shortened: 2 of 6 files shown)


packages/babel-plugin/src/utils/state-manager.js
```diff
@@ -684,17 +684,48 @@ export default class StateManager {
...
+      }
+      if ((styleObj as any).constVal != null) {
+        const val =
+          typeof (styleObj as any).constVal === 'number'
+            ? t.numericLiteral((styleObj as any).constVal)
+            : t.stringLiteral((styleObj as any).constVal);
+        metadataProps.push(
+          t.objectProperty(t.identifier('constVal'), val),
+        );
+      }
```
nmn, line 707: There should never be a `constKey` without a `constVal` right?

```suggestion
      const constKey: ?string = (styleObj as any).constKey
      const constVal: ?number | string = (styleObj as any).constVal
      if (constKey != null && constVal != null) {
        metadataProps.push(
          t.objectProperty(
            t.identifier('constKey'),
            t.stringLiteral((styleObj as any).constKey),
          ),
        );
        const val =
          typeof (styleObj as any).constVal === 'number'
            ? t.numericLiteral(constVal)
            : t.stringLiteral(constVal);
        metadataProps.push(
          t.objectProperty(t.identifier('constVal'), val),
        );
      }
```

nmn: But also, `constKey` and `constVal` is all one global key-value map and so it probably makes sense to use the `constKey` as an actual key.

packages/stylex/src/StyleXSheet.js
```diff
@@ -91,13 +91,22 @@ export class StyleXSheet {
...
     this.rules = [];
+    this.constantsCache = new Map();
 
     this.rootTheme = opts.rootTheme;
     this.rootDarkTheme = opts.rootDarkTheme;
     this.supportsVariables =
       opts.supportsVariables ?? doesSupportCSSVariables();
   }
 
+  constantsCache: Map<
```
nmn, line 102: nit: It's not a `cache`. It's just typical state. So maybe call this `constsMap`?

packages/stylex/src/StyleXSheet.js
```diff
@@ -91,13 +91,22 @@ export class StyleXSheet {
...
+    string,
+    {|
+      value: string | number,
+      rules: Array<number | string>,
+    |},
```
nmn, line 107: Object types are exact by default.

```suggestion
    {
      value: string | number,
      rules: Array<number | string>,
    },
```

nmn: Also, the `rules` should either be an array of `number` (indexes) or `string`, but not both.

`string` *might* make a bit more sense as the indices are not stable.

packages/stylex/src/StyleXSheet.js
```diff
@@ -280,54 +289,123 @@ export class StyleXSheet {
...
-    // Inject the stylesheet if it hasn't already been
-    if (this.injected === false) {
+  insert(
+    rawLTRRule: string,
+    priority: number,
+    rawRTLRule: ?string,
+    metadata?: ?{|
+      constKey?: string,
+      constVal?: string | number,
+    |},
```
nmn, line 299:
```suggestion
    metadata?: ?$ReadOnly<{
      constKey?: string,
      constVal?: string | number,
    }>,
```

packages/stylex/src/StyleXSheet.js
```diff
@@ -280,54 +289,123 @@ export class StyleXSheet {
...
 
-    if (rawRTLRule != null) {
+    // 2) defineConsts update: patch or register constant values
+    if (
+      rawLTRRule === '' &&
+      metadata?.constKey != null &&
+      metadata.constVal != null
+    ) {
+      const key: string = (metadata.constKey: any);
+      const newVal: string = String((metadata.constVal: any));
```
nmn, line 313: I thought we had moved to `as` syntax? This should be:

```suggestion
      const key: string = metadata.constKey as any;
      const newVal: string = String(metadata.constVal as any);
```

However, if you're not seeing a syntax error with your existing code we might have issue with the `.flowconfig`.

packages/stylex/src/StyleXSheet.js
```diff
@@ -280,54 +289,123 @@ export class StyleXSheet {
...
+      console.log(`[defineConsts] invoked with key=${key}, constVal=${newVal}`);
+      console.log('[defineConsts] cache before:', this.constantsCache);
+
+      let entry = this.constantsCache.get(key);
+      if (!entry) {
+        // first defineConsts for this key nothing recorded yet
+        console.log(`[defineConsts] first define for key=${key}, recording value only`);
+        entry = { value: newVal, rules: [] };
+        this.constantsCache.set(key, entry);
```
nmn, line 322: It would be slow, but you *might* want to search for the newly added `constKey` in *all* existing rules.

Since this primarily done for hot reloading, there is a possibility that someone would use a constant before defining it.

packages/stylex/src/StyleXSheet.js
```diff
@@ -280,54 +289,123 @@ export class StyleXSheet {
...
     if (this.isHeadless()) {
       return;
     }
-
     const tag = this.getTag();
     const sheet = tag.sheet;
-
     if (sheet != null) {
       try {
         sheet.insertRule(rule, Math.min(insertPos, sheet.cssRules.length));
```
nmn, line 404: When calculating `insertPos`, new rules are "spliced" in and the indices can then move around.

You need to either not rely on indices or make sure you update the indices when you splice things in.

nmn (commented):
I know it's a draft, but I'm doing this now before the work week.

mellyeliu:
will wait to rebase on https://github.com/facebook/stylex/pull/1017

mellyeliu:
closing, will revisit as we explore runtimeInjection alternatives/reassessments

PR #1008, opened 2025-04-18
[babel-plugin] Move "shared" modules into "babel-plugin"

necolas:
This gets rid of the "shared" package which was previously used to share logic with the "dev-runtime" package, which no longer exists. I tried to keep this change relatively small, so it's focused on moving the code first. Consolidating logic and reorganize the modules within the babel-plugin can be done progressively in the future.

All the non-duplicate unit tests from "shared" have already been moved to the "babel-plugin" tests, so the shared tests are deleted in this diff (except the "types" tests). The source code is moved into the plugin under a new "shared" folder. The few instances where dependencies on "shared" code existed are modified to avoid it.

The "cli" package was importing the hash function, instead this is duplicated in the CLI package.

The "eslint-plugin" package was importing the rule priorities, instead these are inlined and all pseudo-classes use the same priority for now.


packages/babel-plugin/src/shared/README.md
```diff
@@ -1,6 +1,6 @@
-# @stylexjs/shared
+# shared
 
-This package contains most of the core JavaScript logic for stylex.
+This legacy package contains most of the core JavaScript logic for stylex.
```
mellyeliu, line 3: now that i'm reading these a lot of the documentation should be updated, I'll look into that next week

packages/eslint-plugin/src/utils/getPropertyPriorityAndType.js
```diff
@@ -24,31 +18,27 @@ export default function getPropertyPriorityAndType(
   key: string,
 ): PriorityAndType {
   if (key.startsWith('@supports')) {
-    return { priority: AT_RULE_PRIORITIES['@supports'], type: 'atRule' };
+    return { priority: 30, type: 'atRule' };
```
mellyeliu, line 21: do you know what we were using priority for within the linter? a quick grep shows it's used in `sort-keys` but I don't immediately understand why...

we don't use `sort-keys` internally anyway, iirc the OSS lint rule is partially broken

packages/eslint-plugin/src/utils/getPropertyPriorityAndType.js
```diff
@@ -24,31 +18,27 @@ export default function getPropertyPriorityAndType(
...
   }
 
   if (key.startsWith('::')) {
-    return { priority: PSEUDO_ELEMENT_PRIORITY, type: 'pseudoElement' };
+    return { priority: 5000, type: 'pseudoElement' };
```
mellyeliu, line 25: maybe we declare these values on top as constants

mellyeliu (approved):
will be nice to have things colocated (does get disruptive running `npm i` in the babel plugin after every update in shared)

I think Naman had some opinions of maintaining a separate shared package (for developers to use independently?) cc: @nmn

packages/babel-plugin/src/shared/README.md
necolas: I think most of this will end up removed through the process of reorganizing the code now it's moved

packages/eslint-plugin/src/utils/getPropertyPriorityAndType.js
necolas: It was just to order the pseudo classes in code the same way they get ordered in the CSS

necolas:
> I think Naman had some opinions of maintaining a separate shared package (for developers to use independently?)

I don't think there's a strong enough case for this. What we've been open to enabling is people building plugins that operate ahead of stylex to convert code into code that conforms to the API the stylex plugin processes. But it's net-negative for us to continue maintaining a separate public package of APIs that isn't stylex, and isn't documented.

If we eventually find that it's easier to have some common code that is shared across packages, we can re-create a "shared" private package, but instead of it being built and published separately it would be inlined during the build of the public packages. That's what we do [in the React repo](https://github.com/facebook/react/blob/main/packages/shared/package.json) for example.

packages/eslint-plugin/src/utils/getPropertyPriorityAndType.js
nmn: @mellyeliu Your media query work should eventually make the sorting here mostly redundant and a purely stylistic thing.

nmn:
@necolas I don't have strong feelings right now since `dev-runtime` has been discontinued, but there were two other assumptions for the `shared` package:
1. The hope was that it made it easier to contribute to StyleX by having a package that is entirely independent of Babel.
2. With the new tools like Biome, Rolldown etc, becoming more common, it would be easier to build plugins for Babel-alternatives by re-using the `shared` package.

I don't think either of the two assumptions has panned out, so I can't really make a strong argument for keeping the separate package, and unification may have some simplification benefits.

PR #1009, opened 2025-04-19
fix: fix first that works bug where last value is a variable

mellyeliu:
`firstThatWorks(a, b, c)` turns into an array of values that eventually gets passed into some shared `stylex.create` utils. 
```
borderColor: stylex.firstThatWorks('blue', 'var(--accent)')
```
When a property value array started with a CSS variable (e.g. ["var(--x)", "sticky"]), variableFallbacks called composeVars(...) and mistakenly spread its string result into an array. 

Resulting in broken output like:  broken output like 
```
".borderColor-xnybe8l{border-color:v;border-color:a;border-color:r;border-color:(;border-color:-;border-color:-;border-color:a;border-color:c;border-color:c;border-color:e;border-color:n;border-color:t;border-color:);border-color:blue}"
```

This fix wraps the composeVars result in an array to keep the return value consistent.
```
.borderColor-xnybe8l{border-color:var(--accent);border-color:blue;}
```

we were never testing this, so this edge case was missed

### DIFF
--- a/packages/babel-plugin/__tests__/transform-stylex-create-test.js
+++ b/packages/babel-plugin/__tests__/transform-stylex-create-test.js
@@ -1172,8 +1172,8 @@ describe('@stylexjs/babel-plugin', () => {
       });
 
       describe('function value: stylex.firstThatWorks()', () => {
+        // Check various combinations of fallbacks
         test('args: value, value', () => {
-          // Checks various combinations of fallbacks
           const { code, metadata } = transform(`
             import * as stylex from '@stylexjs/stylex';
             export const styles = stylex.create({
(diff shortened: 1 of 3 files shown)


necolas:
We should stack this on https://github.com/facebook/stylex/pull/1008, which has the unit tests that exposed this bug. This PR should un-skip those tests and make sure the new snapshots look right.

packages/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -1302,12 +1352,32 @@ describe('@stylexjs/babel-plugin', () => {
...
+                  3000,
+                ],
+              ],
+            }
+          `);
         });
 
-        // TODO: Fix parser bug
-        test.skip('args: func, var, value, value', () => {
+        test('args: func, var, value, value', () => {
```
necolas, line 1380: Thinking about this one. The assumptions originally made here might not be correct. For example, there's no fallback if variables aren't supported.

necolas (approved):

packages/babel-plugin/__tests__/transform-stylex-create-test.js
nmn: We intentionally didn't implement a fallback for browsers where variables aren't supported as they're very widely supported in all browsers. Given that it would increase the CSS size a bit, do you think it's worth doing or doing conditionally based on a compiler config?

PR #1013, opened 2025-04-21
[babel-plugin] Rename 'genConditionalClasses' to 'enableInlinedConditionalMerge'

necolas:
Turns this feature on by default but keeps the feature flag in case we have problems internally.

* Rename 'genConditionalClasses' to 'enableInlinedConditionalMerge'
* Enable the feature flag by default.
* Remove option from plugin documentation


mellyeliu (approved):
can we call this something like `enableConditionalClassInlining`? or something to flag that it's for conditional styles

PR #1015, opened 2025-04-21
style-value-parser/feat: implement toString function in calc parser

mellyeliu:
This is a dependency to support media queries with `calc` and other uses

Two main changes:
- Added a new `group` node type to preserve user written parentheses in `calc()` expressions. We wrap all explicitly parenthesized expressions (like `(10 + 5)`) in a `{ type: 'group', expr }` node.
- Updated the parser to support nested groups and maintain correct order of operations

Implemented:
- toString()` method reconstructs the original `calc(...)` string exactly, including parentheses
   - We can work on simplication afterwards (ie. collapsing stuff like parentheses implied by order of operation)
- Added test for round trip accuracy: `parse -> toString` returns the original string for all valid inputs

Known further improvements:
- Parser does not flag `calc(a+ b )` as invalid (flags other instances). Will fix in follow up

### DIFF
--- a/packages/style-value-parser/src/css-types-from-tokens/__tests__/calc-test.js
+++ b/packages/style-value-parser/src/css-types-from-tokens/__tests__/calc-test.js
@@ -243,8 +243,36 @@ describe('Test CSS Type: calc()', () => {
   test('rejects invalid calc expressions', () => {
     expect(() => Calc.parser.parseToEnd('calc()')).toThrow();
     expect(() => Calc.parser.parseToEnd('calc(10 + )')).toThrow();
+    expect(() => Calc.parser.parseToEnd('calc(10 +5 )')).toThrow();
+    expect(() => Calc.parser.parseToEnd('calc(10+5 )')).toThrow();
     expect(() => Calc.parser.parseToEnd('calc(10 @ 5)')).toThrow();
     expect(() => Calc.parser.parseToEnd('calc(10px + 5em)')).not.toThrow();
     expect(() => Calc.parser.parseToEnd('notcalc(10 + 5)')).toThrow();
+  });
+
+  test('toString round‑trips calc expressions', () => {
+    const cases = [
+      'calc(10)',
+      'calc(3.14)',
+      'calc(-5)',
+      'calc(50%)',
+      'calc(20px)',
+      'calc(10 + 5)',
+      'calc(20px + 10%)',
+      'calc(10 - 5)',
+      'calc(10 * 5)',
+      'calc(10 / 2)',
+      'calc((10 + 5) * 2)',
+      'calc(100% - 20px * 2 + 10px)',
+      'calc((100% - (30px / 2)))',
+    ];
+    cases.forEach((str) => {
+      expect(Calc.parser.parse(str).toString()).toBe(str);
+    });
+  });
+
+  test.skip('rejects invalid calc spacing', () => {
+    // TODO: Parser does not enforce this edge case properly, need to fix
+    expect(() => Calc.parser.parseToEnd('calc(10+ 5 )')).toThrow();
   });
 });
--- a/packages/style-value-parser/src/css-types-from-tokens/__tests__/calc-test.js
+++ b/packages/style-value-parser/src/css-types-from-tokens/__tests__/calc-test.js
@@ -136,9 +136,12 @@ describe('Test CSS Type: calc()', () => {
       new Calc({
         type: '*',
         left: {
-          type: '+',
-          left: 10,
-          right: 5,
+          type: 'group',
+          expr: {
+            type: '+',
+            left: 10,
+            right: 5,
+          },
         },
         right: 2,
       }),
@@ -148,13 +151,16 @@ describe('Test CSS Type: calc()', () => {
         type: '-',
         left: new Percentage(100),
         right: {
-          type: '/',
-          left: {
-            type: NumberType.Integer,
-            value: 30,
-            unit: 'px',
+          type: 'group',
+          expr: {
+            type: '/',
+            left: {
+              type: NumberType.Integer,
+              value: 30,
+              unit: 'px',
+            },
+            right: 2,
           },
-          right: 2,
         },
       }),
     );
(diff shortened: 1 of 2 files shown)


packages/style-value-parser/src/css-types-from-tokens/calc.js
```diff
@@ -39,6 +39,8 @@ type Division = {
   right: CalcValue,
 };
 
+type Group = { type: 'group', expr: CalcValue };
```
mellyeliu, line 42: this denotes whether parens are wrapping this sub-section of the calc

packages/style-value-parser/src/css-types-from-tokens/calc.js
```diff
@@ -186,3 +189,53 @@ export class Calc {
...
+  ) {
+    const opNode: Addition | Subtraction | Multiplication | Division =
+      value as any;
+    return (
+      calcValueToString(opNode.left) +
+      ' ' +
+      opNode.type +
+      ' ' +
+      calcValueToString(opNode.right)
+    );
```
nmn, line 226:
```suggestion
    return [
      calcValueToString(opNode.left),
      opNode.type,
      calcValueToString(opNode.right)
    ].join(' ');
```

packages/style-value-parser/src/css-types-from-tokens/calc.js
```diff
@@ -186,3 +189,53 @@ export class Calc {
...
+  // binary operations
+  if (
+    value != null &&
+    typeof value === 'object' &&
+    'left' in value &&
+    'right' in value &&
+    typeof (value as any).type === 'string'
+  ) {
+    const opNode: Addition | Subtraction | Multiplication | Division =
+      value as any;
```
nmn, line 219: You shouldn't need this many `as any` in the code.

packages/style-value-parser/src/css-types-from-tokens/calc.js
```diff
@@ -186,3 +189,53 @@ export class Calc {
       .map(([_, value, _closeParen]) => new Calc(value));
   }
 }
+
+function calcValueToString(value: CalcValue | string): string {
```
nmn, line 193: Why does this need to accept `string` as a param type?

packages/style-value-parser/src/css-types-from-tokens/calc.js
```diff
@@ -186,3 +189,53 @@ export class Calc {
...
+  // raw identifiers/constants
+  if (typeof value === 'string') {
+    return value;
+  }
+
+  // explicit user‑written group
+  if (value != null && typeof value === 'object' && 'expr' in value) {
+    const group: Group = value as any;
+    return '(' + calcValueToString(group.expr) + ')';
+  }
```
nmn, line 208: AFAIK, this should work:
```suggestion
  if (value != null && typeof value === 'object' && value.type === 'group') {
    return '(' + calcValueToString(group.expr) + ')';
  }
```

packages/style-value-parser/src/css-types-from-tokens/calc.js
nmn, line 219: This should also work if you've already checked for all non-object types previously.

```suggestion
  if (
    value != null &&
    typeof value === 'object' &&
    ( value.type === '+' ||
      value.type === '-' ||
      value.type === '*' ||
      value.type === '/' )
  ) {
```

packages/style-value-parser/src/css-types-from-tokens/calc.js
```diff
@@ -186,3 +189,53 @@ export class Calc {
...
+  // dimensions or percentages
+  if (
+    value != null &&
+    typeof value === 'object' &&
+    'value' in value &&
+    'unit' in value
+  ) {
+    const d: { value: number, unit: string } = value as any;
+    return `${d.value}${d.unit}`;
+  }
```
nmn, line 238: It should be possible to check for this case before the other cases
with a `'unit' in value` or similar.

If there are still Flow issues, it's better to make a single safe helper function that checks the type rather than littering `any` types everywhere.

nmn (approved):
Code looks good. Flow types could be a lot better.

packages/style-value-parser/src/css-types-from-tokens/__tests__/calc-test.js
```diff
@@ -136,9 +136,12 @@ describe('Test CSS Type: calc()', () => {
       new Calc({
         type: '*',
         left: {
-          type: '+',
-          left: 10,
-          right: 5,
+          type: 'group',
```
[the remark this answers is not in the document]
mellyeliu: apparently they're called 'grouping expression' or 'parenthetical expression'

packages/style-value-parser/src/css-types-from-tokens/calc.js
mellyeliu: i tried using 'unit' in value first and also wrote a helper, but flow still couldn't refine the union without casting. it doesn’t recognize that checking type === '+' excludes other shapes like Percentage, so it throws even with guards. i think we'd have to restructure the entire `CalcValue` type to keep flow happy. happy to revisit if we want to clean this up more structurally

packages/style-value-parser/src/css-types-from-tokens/__tests__/calc-test.js
```diff
@@ -237,8 +243,36 @@ describe('Test CSS Type: calc()', () => {
...
   });
+
+  test('toString round‑trips calc expressions', () => {
+    const cases = [
+      'calc(10)',
+      'calc(3.14)',
+      'calc(-5)',
+      'calc(50%)',
+      'calc(20px)',
+      'calc(10 + 5)',
```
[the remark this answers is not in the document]
mellyeliu: yeah i want to add a series of follow up simplifications to:
- remove parens when unneeded by order of operations,
- simplify expressions

building out the basic validator first

mellyeliu:
The issue here is that I don't think flow can narrow in on the union because not every value has a `type` field - thus something like `value.type === '+'` doesn't filter out the other flow types. Would need a major refactor of the entire `CalcValue` type that I can explore once the calc parser is complete

PR #1017, opened 2025-04-23
[stylex] Rewrite runtime style injection

necolas:
Best to review this one commit at a time. This code for client-side style injection isn't used internally at Meta, which has its own forks.

* Add support for updating style sheets in iframes and shadow roots.
* Add support for hydrating from SSR-generated `<style>` tag.
* Add unit test coverage for style injection.
* Add unit test coverage for runtime stylex module.
* Add basic React app (using runtime injection) to example-rollup.

image


packages/scripts/flow-translator/rewrite-imports.js
```diff
@@ -37,18 +37,16 @@ async function rewriteImportsInFolder(
...
+  const dirents = await fs.readdir(inputDir, {
+    recursive: true,
+    withFileTypes: true,
+  });
   for (const dirent of dirents) {
-    // We only care about the `stylex` package for now which has
-    // a flat file tree.
-    if (dirent.isDirectory()) {
-      continue;
-    }
```
necolas, line 46: This script would silently fail to rewrite imports when not using a flat file tree. Should be fixed now.

packages/stylex/.babelrc.js
```diff
@@ -7,76 +7,22 @@
...
-            oldValue.lastIndexOf('/') + 1,
-          );
-        }
-      },
-      ExportDefaultDeclaration(path) {
-        path.remove();
-      },
-    },
-  };
-}
```
necolas, line 26: Not used by `build-haste` in this package

packages/stylex/.babelrc.js
```diff
@@ -7,76 +7,22 @@
...
-          if (source.endsWith('.js')) {
-            node.value = source.slice(0, -3) + '.mjs';
-            return;
-          }
-          node.value = source + '.mjs';
-        }
-      },
-    },
-  };
-}
```
necolas, line 47: Not needed when using rollup to build the packages

packages/stylex/.babelrc.js
```diff
@@ -7,76 +7,22 @@
...
-  presets,
-  plugins,
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        exclude: ['@babel/plugin-transform-typeof-symbol'],
+        targets: 'defaults',
+        // Convert files to cjs for jest testing
+        modules: BABEL_ENV === 'test' ? 'cjs' : false,
```
necolas, line 21: Rollup otherwise wants to consume esm modules whatever the output format.

packages/stylex/package.json
```diff
@@ -2,29 +2,19 @@
...
-    },
-    "./lib/StyleXTypes": {
-      "import": "./lib/StyleXTypes.js",
-      "require": "./lib/StyleXTypes.js",
-      "types": "./lib/StyleXTypes.d.ts"
-    },
-    "./lib/StyleXSheet": {
-      "import": "./lib/es/StyleXSheet.mjs",
-      "require": "./lib/StyleXSheet.js",
-      "types": "./lib/StyleXSheet.d.ts"
```
necolas, line 22: This file should not be part of the public API

packages/stylex/package.json
```diff
@@ -2,29 +2,19 @@
...
+      "require": "./lib/cjs/stylex.js",
+      "types": "./lib/cjs/stylex.d.ts"
     },
     "./lib/stylex-inject": {
```
necolas, line 14: I'll fix this export path in another diff

packages/stylex/package.json
```diff
@@ -34,13 +24,12 @@
   },
   "license": "MIT",
   "scripts": {
-    "prebuild": "gen-types -i src/ -o lib/",
-    "build:cjs": "cross-env BABEL_ENV=cjs babel src/ --out-dir lib/ --copy-files",
-    "build:esm": "cross-env BABEL_ENV=esm babel src/ --out-dir lib/es --out-file-extension .mjs",
-    "postbuild:esm": "rollup -c ./rollup.config.mjs",
+    "prebuild": "rimraf lib && gen-types -i src/ -o lib/cjs && cp -r lib/cjs lib/es",
+    "build:cjs": "cross-env BABEL_ENV=cjs rollup -c ./rollup.config.mjs",
+    "build:esm": "cross-env BABEL_ENV=esm rollup -c ./rollup.config.mjs",
```
necolas, line 29: Use rollup to build bundles

packages/stylex/package.json
```diff
@@ -34,13 +24,12 @@
...
     "build": "npm run build:cjs && npm run build:esm",
-    "build-haste": "rewrite-imports -i src/ -o lib/",
-    "test": "jest"
+    "build-haste": "rimraf lib && rewrite-imports -i src/ -o lib/",
```
necolas, line 31: During the internal sync, we run the esm AND cjs builds for some reason (as this package is a dependency of the babel-plugin build). Later, we run the haste build but don't seem to use the esm/cjs builds, so I'm assuming it's safe to clean out the directory before this build.

packages/stylex/package.json
```diff
@@ -34,13 +24,12 @@
   },
   "license": "MIT",
   "scripts": {
-    "prebuild": "gen-types -i src/ -o lib/",
-    "build:cjs": "cross-env BABEL_ENV=cjs babel src/ --out-dir lib/ --copy-files",
-    "build:esm": "cross-env BABEL_ENV=esm babel src/ --out-dir lib/es --out-file-extension .mjs",
-    "postbuild:esm": "rollup -c ./rollup.config.mjs",
+    "prebuild": "rimraf lib && gen-types -i src/ -o lib/cjs && cp -r lib/cjs lib/es",
```
necolas, line 27: This cleans out the build directory before every build, to prevent us potentially publishing to npm with artifacts from past versions that might be lying around on your local filesystem. It also copies the types across cjs/es dirs for when using flow.

packages/stylex/src/inject.js
```diff
@@ -0,0 +1,20 @@
...
+
+'use strict';
+
+import { createSheet } from './stylesheet/createSheet';
+import { addSpecificityLevel } from './stylesheet/utils';
+
+const sheet = createSheet();
+
+export default function inject(cssText: string, priority: number): void {
+  addSpecificityLevel(cssText, Math.floor(priority / 1000));
```
necolas, line 18: This `priority/1000` logic probably needs to be revisited for both inject and the babel plugin. Can be done as part of a refactor of how priorities are defined and used.

packages/stylex/src/types/VarTypes.js
```diff
@@ -7,7 +7,7 @@
  * @flow strict
  */
 
-import type { ValueWithDefault } from './util-types';
+import type { ValueWithDefault } from './utils';
```
necolas, line 10: Some of these paths will need to be updated in the internal sync script

packages/stylex/src/stylesheet/createOrderedCSSStyleSheet.js
```diff
@@ -0,0 +1,208 @@
...
+  /**
+   * Hydrate approximate record from any existing rules in the sheet.
+   */
+  if (sheet != null) {
+    let group = 0;
+    for (let i = 0; i < sheet.cssRules.length; i++) {
+      const cssRule = sheet.cssRules[i];
+      const cssText = cssRule.cssText;
+      // Create record of existing selectors and rules
+      if (cssText.indexOf('stylesheet-group') > -1) {
```
mellyeliu, line 48: should we use a regex pattern here instead with something like `startsWith`, to make this safer?

packages/stylex/src/inject.js
mellyeliu, line 18: i think we need to assign the return value of addSpecificityLevel and pass that into sheet.insert, it doesn’t mutate cssText in place

packages/stylex/src/stylesheet/createOrderedCSSStyleSheet.js
necolas: I don't think that's necessary as this string is controlled by the script and no other cssText is going to contain it. It's how React Native for Web has worked for nearly a decade.

mellyeliu (approved):

necolas:
I'll do a test sync of this PR before merging. I need to update the sync script and double-check this runtime injection code isn't used anywhere internally yet.

PR #1019, opened 2025-04-24
feat: Add new stylex.positionTry for @property-try declarations

### LINKED ISSUE #1003: [babel-plugin] Define API for `@position-try`

CSS now a @position-try at rule (with limited browser support today)

https://developer.mozilla.org/en-US/docs/Web/CSS/@position-try

Discussion: https://github.com/facebook/stylex/discussions/860

nmn:

Adds a new `stylex.positionTry` API that lets you create `@position-try` rules.
It works pretty much just like `stylex.keyframes`.

## Linked PR/Issues

Fixes #1003 

## Additional Context

New test file is included which shows the new feature working.

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-import-export-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-import-export-test.js
@@ -52,6 +52,7 @@ const defaultImportMap = {
   defineVars: 'stylex.defineVars',
   firstThatWorks: 'stylex.firstThatWorks',
   keyframes: 'stylex.keyframes',
+  positionTry: 'stylex.positionTry',
   props: 'stylex.props',
 };
 
@@ -74,6 +75,7 @@ function createStylesFixture({
     defineVars,
     firstThatWorks,
     keyframes,
+    positionTry,
     props,
   } = importMap;
 
@@ -99,6 +101,15 @@ function createStylesFixture({
 
   return `
     ${defineConstsAndVarsOutput}
+    const fallback1 = ${positionTry}({
+      anchorName: '--myAnchor',
+      positionArea: 'top left',
+    });
+    const fallback2 = ${positionTry}({
+      anchorName: '--otherAnchor',
+      top: 'anchor(bottom)',
+      insetInlineStart: 'anchor(start)',
+    });
     const styles = ${create}({
       root: {
         animationName: ${keyframes}({
@@ -109,6 +120,7 @@ function createStylesFixture({
             backgroundColor: 'orange'
           },
         }),
+        positionTryFallbacks: \`\${fallback1}, \${fallback2}\`,
         color: {
           default: 'red',
           [constants.mediaQuery]: 'blue'
@@ -210,9 +222,12 @@ describe('@stylexjs/babel-plugin', () => {
           bar: "var(--x1hi1hmf)",
           __themeName__: "xop34xu"
         };
+        const fallback1 = "--x5jppmd";
+        const fallback2 = "--x17pzx6";
         const styles = {
           root: {
             kKVMdj: "x1qar0u3",
+            k9M3vk: "x7cint9",
             kMwMTN: "x1e2nbdu x14693no",
             kVAEAm: "x15oojuh",
             $$css: true
@@ -229,6 +244,22 @@ describe('@stylexjs/babel-plugin', () => {
         {
           "stylex": [
             [
+              "--x5jppmd",
+              {
+                "ltr": "@position-try --x5jppmd {anchor-name:anchor-name;anchor-name:--myAnchor;position-area:position-area;position-area:top left;}",
+                "rtl": "@position-try --x5jppmd {anchor-name:--myAnchor;position-area:top left;}",
+              },
+              1,
+            ],
+            [
+              "--x17pzx6",
+              {
+                "ltr": "@position-try --x17pzx6 {anchor-name:anchor-name;anchor-name:--otherAnchor;inset-inline-start:inset-inline-start;inset-inline-start:anchor(start);top:top;top:anchor(bottom);}",
+                "rtl": "@position-try --x17pzx6 {anchor-name:--otherAnchor;inset-inline-start:anchor(start);top:anchor(bottom);}",
+              },
+              1,
+            ],
+            [
               "xjx6k13-B",
               {
                 "ltr": "@keyframes xjx6k13-B{from{background-color:yellow;}to{background-color:orange;}}",
@@ -240,6 +271,14 @@ describe('@stylexjs/babel-plugin', () => {
               "x1qar0u3",
               {
                 "ltr": ".x1qar0u3{animation-name:xjx6k13-B}",
+                "rtl": null,
+              },
+              3000,
+            ],
+            [
+              "x7cint9",
+              {
+                "ltr": ".x7cint9{position-try-fallbacks:--x5jppmd,--x17pzx6}",
                 "rtl": null,
               },
               3000,
@@ -292,6 +331,7 @@ describe('@stylexjs/babel-plugin', () => {
           defineVars: 'foo.defineVars',
           firstThatWorks: 'foo.firstThatWorks',
           keyframes: 'foo.keyframes',
+          positionTry: 'foo.positionTry',
           props: 'foo.props',
         },
       });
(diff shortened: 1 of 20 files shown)


packages/babel-plugin/__tests__/transform-stylex-position-try-test.js
```diff
@@ -0,0 +1,297 @@
...
+          left: '0',
+          width: SIZE,
+          height: SIZE
+        });
+        export const styles = stylex.create({
+          root: {
+            positionTryFallbacks: \`\${name}\`,
+          }
+        });
+      `);
```
necolas, line 113: This fixture isn't really testing "template literals used in position-try object". It's testing literals in `create`, which we already test elsewhere. I noticed there's also a keyframes test like this, so probably should update that one too.

packages/babel-plugin/__tests__/transform-stylex-position-try-test.js
```diff
@@ -0,0 +1,297 @@
...
+                "rtl": null,
+              },
+              3000,
+            ],
+          ],
+        }
+      `);
+    });
+
+    test('position-try object with RTL properties', () => {
```
necolas, line 201: Let's remove this one since we don't need RTL-specific tests going forward

packages/babel-plugin/__tests__/transform-stylex-position-try-test.js
```diff
@@ -0,0 +1,297 @@
...
+              1,
+            ],
+          ],
+        }
+      `);
+    });
+  });
+});
+
+describe('stylex.positionTry', () => {
```
necolas, line 236: Are these junk tests that Cursor generated? We can delete these.

packages/babel-plugin/src/shared/stylex-position-try.js
```diff
@@ -0,0 +1,96 @@
...
+    (value, key) => generateRtl([key, value]) ?? value,
+  );
+
+  const ltrString = constructPositionTryObj(ltrStyles);
+  const rtlString = constructPositionTryObj(rtlStyles);
+
+  const positionTryName = '--' + classNamePrefix + createHash(ltrString);
+
+  const ltr = `@position-try ${positionTryName} {${ltrString}}`;
+  const rtl =
```
necolas, line 55: Let's not add this RTL code to new APIs

packages/babel-plugin/src/shared/stylex-position-try.js
```diff
@@ -0,0 +1,96 @@
...
+// Similar to `stylex.keyframes` but for position-try-fallbacks
+// Takes an object of positioning properties and returns a string after hashing it
+// The generated string must be prefixed with -- for anchor positioning
+export default function styleXPositionTry(
+  styles: { +[string]: string | number },
+  options: StyleXOptions = defaultOptions,
+): [string, InjectableStyle] {
+  const { classNamePrefix = 'x' } = options;
+  const expandedObject = Pipe.create(styles)
+    .pipe((styles) => expandPositionShorthands(styles, options))
```
necolas, line 36: Why do shorthands need to be expanded here?

packages/babel-plugin/src/visitors/stylex-create-theme.js
```diff
@@ -94,11 +95,26 @@ export default function transformStyleXCreateTheme(
       return animationName;
     }
 
+    // eslint-disable-next-line no-inner-declarations
+    function positionTry<Obj: { +[k: string]: string | number }>(
+      animation: Obj,
+    ): string {
+      const [animationName, injectedStyle] = stylexPositionTry(
```
necolas, line 102: These var names look like they come from keyframes not positionTry

packages/babel-plugin/src/visitors/stylex-create.js
```diff
@@ -88,6 +89,20 @@ export default function transformStyleXCreate(
...
+    function positionTry<
+      Obj: {
+        +[k: string]: string | number,
+      },
+    >(fallbackStyles: Obj): string {
+      const [positionTryName, injectedStyle] = stylexPositionTry(
+        fallbackStyles,
+        state.options,
+      );
+      injectedKeyframes[positionTryName] = injectedStyle;
```
necolas, line 102: This var name doesn't make sense anymore

necolas (changes requested):
We should also add the relevant validation to prevent arbitrary properties being set in positionTry calls - https://drafts.csswg.org/css-anchor-position-1/#om-position-try

packages/babel-plugin/src/visitors/stylex-position-try.js
```diff
@@ -0,0 +1,127 @@
...
+import type { FunctionConfig } from '../utils/evaluate-path';
+
+import * as t from '@babel/types';
+import StateManager from '../utils/state-manager';
+import styleXPositionTry from '../shared/stylex-position-try';
+import { evaluate } from '../utils/evaluate-path';
+import { firstThatWorks as stylexFirstThatWorks } from '../shared';
+
+const messages = {
+  ILLEGAL_ARGUMENT_LENGTH:
```
mellyeliu, line 20: we should be consistent with either defining these in the `shared/messages.js` file or keep them colocated with the babel visitors. we can export template literals from the file to reuse other messages too, like the similar one we have for `stylex.create` calls

```
export const ILLEGAL_ARGUMENT_LENGTH =
  'stylex.create() should have 1 argument.';
```

packages/babel-plugin/src/shared/stylex-position-try.js
nmn: This is a badly named function. It does the appropriate thing for each config option and doesn’t expand shorthands in either of the non-legacy config options.

packages/babel-plugin/src/shared/stylex-position-try.js
nmn: AFAIK, the RTL code is still used at for Meta codebases. Other changes are needed before this code can be removed. I'll start working on removing that code as it should be safe, even at Meta.

necolas: It might be used for existing APIs, but we don't need it for this one. This API won't be used at Meta for a while, and I don't think we need to propagate the issues with RTL to new APIs in general.

necolas (approved):
LGTM! Thanks for the follow up

necolas (changes requested):
Since this is a new exported API, there are a few tests that need updating to account for this new API, specifically the import/export tests

packages/stylex/src/types/StyleXCSSTypes.js
```diff
@@ -1411,6 +1411,10 @@ export type CSSProperties = $ReadOnly<{
   perspective?: all | perspective,
   perspectiveOrigin?: all | perspectiveOrigin,
   pointerEvents?: all | pointerEvents,
+
+  positionAnchor?: all | string,
```
necolas, line 1415: These types already exist below

packages/stylex/src/stylex.js
```diff
@@ -184,6 +189,7 @@ type IStyleX = {
     ...v: $ReadOnlyArray<T>
   ) => $ReadOnlyArray<T>,
   keyframes: (keyframes: Keyframes) => string,
+  positionTry: (positionTry: PositionTry) => string,
```
necolas, line 192: You'll need to update the unit tests too

packages/babel-plugin/src/shared/messages.js
```diff
@@ -26,6 +26,24 @@ export const unboundCallValue = (fn: string): string =>
 export const DUPLICATE_CONDITIONAL =
   'The same pseudo selector or at-rule cannot be used more than once.';
 export const ESCAPED_STYLEX_VALUE = 'Escaping a create() value is not allowed.';
+export const ILLEGAL_ARGUMENT_LENGTH = 'create() should have 1 argument.';
```
necolas, line 29: This looks like a bad merge? The messages were refactored in #1030

necolas (changes requested):
Still needs to update the import/export tests here https://github.com/facebook/stylex/blob/main/packages/babel-plugin/__tests__/transform-import-export-test.js

There was a coverage regression because you need to update the runtime test suite after adding a new export. And it looks the rebase was resolved incorrectly which reintroduced deleted code and duplicated some types?

packages/stylex/src/types/StyleXCSSTypes.js
nmn: another bad merge I assume.

necolas (approved):
LGTM. Let's add docs in another patch

PR #1022, opened 2025-04-25
[style-value-parser] chore: replace character parser with token parser

mellyeliu:
We're moving from the character based parser to the token based parser. Let's delete this unused code

Also updates token naming like `core2` to `token-parser` and relevant (now redundant) file paths like `css-types-from-tokens` to `css-types`


mellyeliu:
separate commits for easy review, will squash before merge

necolas (approved):
Please do a **rebase and merge** rather than a squash merge. The separate diffs are much cleaner than the squashed diff, which makes it look like a lot of tests were rewritten.

PR #1027, opened 2025-04-29
[babel-plugin] Add unit tests for processStylexRules

necolas:

Adds some initial tests for the `stylexPlugin.processStylexRules` function that operates on extracted metadata. This takes the test coverage of `src/index.js` from 44% to 84%.


mellyeliu (approved):

mellyeliu:
we should add a test somewhere to assert that it throws for circular const refs, but I'm not sure if it best belongs here

Issue #1028, opened 2025-04-29
Incorrect error messages across most APIs


mellyeliu:
### Describe the issue

Many error messages are incorrectly hardcoded to refer to `stylex.create()`, even when thrown from other APIs like `stylex.keyframes()`, `stylex.defineVars()`, or `stylex.defineConsts()`.

eg.
```
SyntaxError: unknown file: The return value of stylex.defineVars() must be bound to a named export.
      1 |
      2 | import * as stylex from '@stylexjs/stylex';
    > 3 | const constants = stylex.defineConsts({
        |                   ^
      4 |   YELLOW: 'yellow'
      5 | });
```

## Expected behaviour

Error messages should reference a specific API:
The return value of stylex.defineConsts() must be bound to a named export.

## Fix

```
ILLEGAL_ARGUMENT_LENGTH =
  'stylex.create() should have 1 argument.';
UNBOUND_STYLEX_CALL_VALUE =
  'stylex.create() calls must be bound to a bare variable.';
```

should be converted to 

```
export const illegalArgumentLength = (fn: string, argLength: number) =>
  `${fn}() should have ${argLength} argument${argLength === 1 ? '' : 's'}.`;

export const unboundCallValue = (fn: string) =>
  `${fn}() calls must be bound to a bare variable.`;
```

PR #1030, opened 2025-04-29
babel-plugin/chore: convert static error messages to template functions

mellyeliu:
Fixes https://github.com/facebook/stylex/issues/1028

- Refactors hardcoded error strings into template functions for consistency and reuse across different APIs (previously many were incorrectly referencing `stylex.create()` calls
- Removes some unused error messages

### DIFF
--- a/packages/babel-plugin/src/shared/messages.js
+++ b/packages/babel-plugin/src/shared/messages.js
@@ -12,64 +12,42 @@
 // This file contains constants to be used within Error messages.
 // The URLs within will eventually be replaced by links to the documentation website for Stylex.
 
-export const ILLEGAL_ARGUMENT_LENGTH =
-  'stylex.create() should have 1 argument.';
-export const ILLEGAL_ARG_LENGTH_FOR_KEYFRAMES =
-  'stylex.keyframes() should have 1 argument.';
-export const NON_STATIC_VALUE =
-  'Only static values are allowed inside of a stylex.create() call.';
-export const NON_STATIC_KEYFRAME_VALUE =
-  'Only static values are allowed inside of a stylex.keyframes() call.';
-export const ESCAPED_STYLEX_VALUE =
-  'Escaping a stylex.create() value is not allowed.';
-export const UNBOUND_STYLEX_CALL_VALUE =
-  'stylex.create calls must be bound to a bare variable.';
-export const ONLY_TOP_LEVEL =
-  'stylex.create() is only allowed at the root of a program.';
-export const NON_OBJECT_FOR_STYLEX_CALL =
-  'stylex.create() can only accept a style object.';
-export const NON_OBJECT_FOR_STYLEX_KEYFRAMES_CALL =
-  'stylex.keyframes() can only accept an object.';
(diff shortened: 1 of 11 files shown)


packages/babel-plugin/src/shared/messages.js
```diff
@@ -12,63 +12,42 @@
...
+  `Only static values are allowed inside of a ${fn}() call.`;
+export const nonExportNamedDeclaration = (fn: string): string =>
+  `The return value of ${fn}() must be bound to a named export.`;
+export const unboundCallValue = (fn: string): string =>
+  `${fn}() calls must be bound to a bare variable.`;
+export const nonStyleObject = (fn: string): string =>
+  `${fn}() can only accept a style object.`;
+
 export const ESCAPED_STYLEX_VALUE =
   'Escaping a stylex.create() value is not allowed.';
```
necolas, line 27: Let's remove all the `stylex.` prefixes from the other messages, as done in https://github.com/facebook/stylex/pull/1019/files#diff-37d9716b1d4880fca33aae3ff3c7ffa046c842e27f748609a01684ff5e758bb4

packages/babel-plugin/src/shared/messages.js
```diff
@@ -12,63 +12,42 @@
...
-export const INVALID_PSEUDO_OR_AT_RULE = 'Invalid pseudo or at-rule.';
-export const NO_CONDITIONAL_SHORTHAND =
-  'You cannot use conditional style values for a shorthand property.';
-export const ILLEGAL_NAMESPACE_TYPE =
-  'Only a string literal namespace is allowed here.';
-export const UNKNOWN_NAMESPACE = 'Unknown namespace';
 export const ILLEGAL_NESTED_PSEUDO =
   "Pseudo objects can't be nested more than one level deep.";
+export const INVALID_PSEUDO = 'Invalid pseudo selector, not on the whitelist.';
+export const INVALID_PSEUDO_OR_AT_RULE = 'Invalid pseudo or at-rule.';
```
necolas, line 34: If we're alpha-sorting, these should appear below the "illegal" messages

packages/babel-plugin/src/shared/messages.js
```diff
@@ -12,63 +12,42 @@
...
+export const illegalArgumentLength = (fn: string, argLength: number): string =>
+  `${fn}() should have ${argLength} argument${argLength === 1 ? '' : 's'}.`;
+export const nonStaticValue = (fn: string): string =>
+  `Only static values are allowed inside of a ${fn}() call.`;
+export const nonExportNamedDeclaration = (fn: string): string =>
+  `The return value of ${fn}() must be bound to a named export.`;
+export const unboundCallValue = (fn: string): string =>
+  `${fn}() calls must be bound to a bare variable.`;
+export const nonStyleObject = (fn: string): string =>
+  `${fn}() can only accept a style object.`;
```
necolas, line 24: Before, we'd say "style object" for create calls but just "object" for keyframes. Looking at how this function is now used, I think "can only accept an object" is the message we want.

necolas (approved):
Looks good, just a few small tweaks requested

PR #1031, opened 2025-05-01
[chore] add flow types for anchor positioning css properties to StyleXCSSTypes.js

Jta26:

This PR adds flow types for CSS-based anchor positioning properties. Prevents the need to $FlowFixMe when using these properties.

I saw that @nmn currently has a PR up for supporting @position-try. This would also be useful for me internally in the short to medium term future.

### DIFF
--- a/packages/stylex/src/types/StyleXCSSTypes.js
+++ b/packages/stylex/src/types/StyleXCSSTypes.js
@@ -577,6 +577,21 @@ type pointerEvents =
   | 'all'
   | 'inherit';
 type position = 'static' | 'relative' | 'absolute' | 'sticky' | 'fixed';
+type positionArea =
+  | 'top'
+  | 'left'
+  | 'bottom'
+  | 'right'
+  | 'center'
+  | 'start'
+  | 'end'
+  | 'span-start'
+  | 'span-end'
+  | 'block-start'
+  | 'block-end'
+  | 'inline-start'
+  | 'inline-end';
+type positionVisibility = 'always' | 'anchors-visible' | 'no-overflow';
 type quotes = string | 'none';
 type resize = 'none' | 'both' | 'horizontal' | 'vertical';
 type rowGap = number | string;
@@ -983,6 +998,8 @@ export type CSSProperties = $ReadOnly<{
   alignTracks?: all | string,
   justifyTracks?: all | string,
   masonryAutoFlow?: all | string,
+
+  anchorName?: all | string,
 
   // Not Allowed:
   // all?: all | all,
@@ -1395,6 +1412,14 @@ export type CSSProperties = $ReadOnly<{
   perspectiveOrigin?: all | perspectiveOrigin,
   pointerEvents?: all | pointerEvents,
   position?: all | position,
+
+  positionAnchor?: all | string,
+  positionArea?: all | positionArea,
+  positionTry?: all | string,
+  positionTryFallbacks?: all | string,
+  positionTryOptions?: all | string,
+  positionVisibility?: all | positionVisibility,
+
   quotes?: all | quotes,
   resize?: all | resize,
   rest?: all | rest,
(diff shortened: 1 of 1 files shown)


packages/stylex/src/types/StyleXCSSTypes.js
```diff
@@ -577,6 +577,21 @@ type pointerEvents =
...
+  | 'right'
+  | 'center'
+  | 'start'
+  | 'end'
+  | 'span-start'
+  | 'span-end'
+  | 'block-start'
+  | 'block-end'
+  | 'inline-start'
+  | 'inline-end';
```
Jta26, line 593: note that this doesn't encapsulate all the possible position areas for anchor positioning. There are a lot of them, and I'm unsure if we should have all of them here.

packages/stylex/src/types/StyleXCSSTypes.js
```diff
@@ -577,6 +577,21 @@ type pointerEvents =
...
 type position = 'static' | 'relative' | 'absolute' | 'sticky' | 'fixed';
+type positionArea =
+  | 'top'
+  | 'left'
+  | 'bottom'
+  | 'right'
+  | 'center'
+  | 'start'
+  | 'end'
+  | 'span-start'
```
mellyeliu, line 588: we probably want to support `span-inline-start`/`span-block-start` over `span-start` and `span-inline-end`/`span-block-end` over `span-end` because they're axis explicit and less error prone

packages/stylex/src/types/StyleXCSSTypes.js
necolas: Probably. But could do that later. Is it going to be very verbose in Flow?

necolas: > Some keywords are ambiguous about what axis they refer to: [center](https://drafts.csswg.org/css-anchor-position-1/#valdef-position-area-center), [span-all](https://drafts.csswg.org/css-anchor-position-1/#valdef-position-area-span-all), and the [start](https://drafts.csswg.org/css-anchor-position-1/#valdef-position-area-start)/etc keywords that don’t specify the block or inline axis explicitly. If the other keyword is unambiguous about its axis, then the ambiguous keyword is referring to the opposite axis. (For example, in block-start center, the center keyword is referring to the inline axis.) If both keywords are ambiguous, however, then the first refers to the [block axis](https://drafts.csswg.org/css-writing-modes-4/#block-axis) of the box’s [containing block](https://drafts.csswg.org/css-display-4/#containing-block), and the second to the [inline axis](https://drafts.csswg.org/css-writing-modes-4/#inline-axis). (For example, span-all start is equivalent to span-all inline-start.)

> If only a single keyword is given, it behaves as if the second keyword is [span-all](https://drafts.csswg.org/css-anchor-position-1/#valdef-position-area-span-all) if the given keyword is unambigous about its axis; otherwise, it behaves as if the given keyword was repeated. (For example, [top](https://drafts.csswg.org/css-anchor-position-1/#valdef-position-area-top) is equivalent to top span-all, but [center](https://drafts.csswg.org/css-anchor-position-1/#valdef-position-area-center) is equivalent to center center.)

Maybe we shouldn't allow the ambiguous keywords

mellyeliu: yeah, we should prefer the inline and block keywords over the ambiguous ones (`span-inline-start` `span-block-start` etc and remove `span-start` and `span-end`) for consistency, since we autofix things like `borderStartWidth` to `borderInlineStartWidth`

Jta26: Okay, we can remove *most* of the ambiguous keywords here, but we do need `center` because there is no way to specify the axis for it. There is no `inline-center` or `block-center`. It appears to be entirely dependent on the other keyword it's used with.

PR #1032, opened 2025-05-01
style-value-parser/chore: refactor media query tests

mellyeliu:
Refactor tests and gauge spec gaps. I think we support pretty much all the query types we'd expect to see in internal now, and can start using it in the compiler and linter
- Mostly rearranging tests
- Added new tests for double inequalities, normalization tests, etc

Not supported: 
- niche keywords like horizontal-viewport-segments, vertical-viewport-segments, environment-blending, video-* media features, dynamic-range, prefers-reduced-data

Pasted Graphic

image

### DIFF
--- a/packages/style-value-parser/src/at-queries/__tests__/media-query-test.js
+++ b/packages/style-value-parser/src/at-queries/__tests__/media-query-test.js
@@ -7,673 +7,483 @@
  * @flow strict
  */
 
-import { mediaInequalityRuleParser } from '../media-query.js';
 import { MediaQuery } from '../media-query.js';
 
-describe('Test CSS Type: @media queries', () => {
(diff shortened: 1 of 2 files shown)


necolas:
What's the thinking behind changing the `expect(parsed.toString())` tests to use snapshots rather than assert against the original input query?

mellyeliu:
it's because for a lot of queries, we're normalizing and simplifying the values (removing unnecessary nested operators, converting range operations like `width <= 400px` to `max-width: 400px`). in the future we might want to do more things like collapsing ranges, simplifying all boolean operators. re: https://github.com/facebook/stylex/pull/997#discussion_r2055258343

necolas (approved):

PR #1033, opened 2025-05-01
style-value-parser/feat: add media query validation

mellyeliu:
- Adds a pre-check for balanced parentheses before parsing media queries to catch early syntax issues. 
- Simplifies parser error handling by always throwing a SYNTAX_ERROR on failure.

we may want to support more granular validation in the future (eg for malformed groups or missing colons), but the current parser makes this tricky. it often ingests invalid tokens in unpredictable ways, which leads to inconsistent or misleading errors across different query shapes.


packages/style-value-parser/src/at-queries/__tests__/validation-media-query-test.js
```diff
@@ -0,0 +1,151 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import { validateMediaQuery } from '../media-query';
```
necolas, line 10: Can we not test the parser directly, instead of this validation function?

mellyeliu: I'm planning to add public entry points for the parser - a validator to be used in the linter/compiler (tested here), and later a normalize() function to be used when processing css rules

necolas: What I'm trying to understand is why this diff introduces a new API that wraps one of the parser functions, rather than this being how the parser functions work.

image

mellyeliu: Good q. I originally tried adding an `orThrow` class function to `TokenParser` to use in `MediaQuery` for more descriptive errors. But I ran into issues - the token stream can be easily disrupted, and stray tokens often get consumed as part of another valid pattern, making error detection brittle and unpredictable.

This wrapper lets us keep the TokenParser focused on structural parsing, while layering on simple linear checks (like paren balancing) for more reliable, human-readable errors. (We didn't use character parser because it's too slow for deep recursive checks, but for a linear pass I think it should be fine) It also gives us a stable entry point for the linter/compiler, and a place to pass options for stricter vs looser validation.

Open to iterate if you have thoughts/suggestions. I'm hoping to wrap this in a broader `validate()` function that would run on all the property checks we're building. I'll try to capture this in our optimizing stylex doc + OSS issues.

necolas: Ah ok that makes sense. Thanks!

necolas (approved):

PR #1041, opened 2025-05-02
Organize public packages into '@stylexjs' workspace

necolas:
Move all the public packages into a `@stylexjs` workspace to create a clearer separation between the existing public and private packages.


Jta26 (approved):

mellyeliu (approved):

nmn:
I assume new packages will be moved to `@stylexjs/` as they become public?

I would like to have the style-value-parser be published as a standalone package eventually as it will be used for both the Babel and ESLint plugins but also be useful for other tools for the ecosystem.

necolas:
> I assume new packages will be moved to @stylexjs/ as they become public?

Yeah if we need more public packages they'll probably live under that org too.

> I would like to have the style-value-parser be published as a standalone package eventually as it will be used for both the Babel and ESLint plugins but also be useful for other tools for the ecosystem.

Yeah it can still be used by the babel and eslint plugins without being made a public package.

Issue #1037, opened 2025-05-02
Finish implementing `defineConsts`.


mellyeliu:

mellyeliu:
Documenting in https://github.com/facebook/stylex/issues/1154, https://github.com/facebook/stylex/issues/1162

PR #1046, opened 2025-05-06
[docs] api notes for defineConsts

mellyeliu:
Will wait until the rest of items in https://github.com/facebook/stylex/issues/1037 to merge

image

### DIFF
--- a/packages/docs/docs/api/javascript/defineConsts.mdx
+++ b/packages/docs/docs/api/javascript/defineConsts.mdx
@@ -1,0 +1,77 @@
+---
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+sidebar_position: 3
+---
+
+# `stylex.defineConsts`
+
+Defines static style constants that can be used directly in `stylex.create` calls anywhere in the codebase. Unlike `stylex.defineVars`, these values are inlined at build time and do not generate actual CSS variables.
+
+*Note: `stylex.defineConsts` is an experimental feature and cannot currently be used with `stylex.firstThatWorks` or when `runtimeInjection` is enabled.*
+
+Common use cases include:
+- Media queries
+- Fixed z-index layers
+- Animation durations or easing curves
+- Static spacing, font sizes, or colors that don’t need theming
+
+```ts
+function defineConsts<Consts extends { [key: string]: string }>(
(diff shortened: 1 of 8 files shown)


packages/docs/docs/api/javascript/defineConsts.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+```ts
+function defineConsts<Consts extends { [key: string]: string }>(
+  consts: Consts,
+): Consts<{ [key in keyof Consts]: string }>;
+```
+
+### Example use
+
+You must define your constants as named exports in a `.stylex.js` (or `.stylex.ts`) file:
```
necolas, line 21: Presumably you can mix defineConsts and defineVars in these files? Maybe we need to be really clear that consts get inclined whereas vars turn into CSS variables

packages/docs/docs/api/javascript/defineConsts.mdx
```diff
@@ -0,0 +1,69 @@
...
+import * as stylex from '@stylexjs/stylex';
+import { breakpoints, zIndexes, animations } from './constants.stylex.js';
+
+const styles = stylex.create({
+  container: {
+    position: 'relative',
+    zIndex: zIndexes.modal,
+    transitionTimingFunction: animations.easeInOut,
+    transitionDuration: animations.fast,
+    '@media': {
```
necolas, line 56: This syntax looks wrong. It should be padding with an object value. Probably move `@media` into the const value too

necolas (changes requested):

packages/docs/docs/api/javascript/defineConsts.mdx
mellyeliu: lack of llm proofreading...

packages/docs/docs/api/javascript/defineConsts.mdx
```diff
@@ -0,0 +1,75 @@
...
+```tsx title="constants.stylex.js"
+import * as stylex from '@stylexjs/stylex';
+
+export const breakpoints = stylex.defineConsts({
+  small: '@media (max-width: 600px)',
+  medium: '@media (min-width: 601px) and (max-width: 1024px)',
+  large: '@media (min-width: 1025px)',
+});
+
+export const zIndexes = stylex.defineConsts({
```
necolas, line 38: nit: indices

necolas (approved):

PR #1049, opened 2025-05-07
[babel-plugin] refactor generateCSSRules logic

mellyeliu:
- just some file restructuring + renaming 
- added basic tests for generateCSSRules in prep for processing defineConsts at-rules


packages/@stylexjs/babel-plugin/src/shared/utils/generate-css-rule.js
```diff
@@ -7,16 +7,46 @@
...
+function buildNestedCSSRule(
+  className: string,
+  decls: string,
+  pseudos: $ReadOnlyArray<string>,
+  atRules: $ReadOnlyArray<string>,
+): string {
+  const pseudo = pseudos.filter((p) => p !== '::thumb').join('');
+
+  let selectorForAtRules =
+    `.${className}` + atRules.map(() => `.${className}`).join('') + pseudo;
```
mellyeliu, line 33: @nmn i'm assuming this is for specificity?

vincentriemer (approved):

packages/@stylexjs/babel-plugin/src/shared/utils/__tests__/generate-css-rule-test.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+import { generateCSSRule } from '../generate-css-rule';
+
+describe('generateCSSRule', () => {
```
necolas, line 12: Is there a reason we need these extra tests instead of relying on the existing public API tests? The vast majority of them snapshot generated CSS metadata already

necolas (changes requested):
Code coverage results didn't change so those extra tests are redundant and can be removed

necolas (approved):

PR #1050, opened 2025-05-07
[babel-plugin] support at-rules in defineConsts

mellyeliu:
https://github.com/facebook/stylex/pull/944 only supports inlining values. Need to properly wrap css rules with media queries, @supports, @container etc

1. We were replacing var(--hash) in the final css ltr strings inside our main babel transform (in `processStylexRules`), but that only touches the output css and doesn’t affect how `stylex-create` wraps declarations
2. `stylex-create` walks the raw style object and builds the ltr strings by wrapping rules in @media {...} based on object keys, but we never looked for var(...) keys there—so any media constants were ignored
4. this diff patches the part that builds the `PreRule`s to detect key segments matching var(--hash) and emits them as literal var(--hash){...} wrappers as a placeholder. once `processStylexRules` runs (where we have the full consts map of { hash: '@media(...)' }), we replace those var(--hash) wrappers with their real queries as usual.
5. We can't inline within `PreRule` itself because we don't know all the constants until we scan all the defineConsts calls. It'd require a major reordering of visitors to capture them before stylex-create. (originally tried this)

TODO:
1. ~Add unit tests for defineConsts that include nested contextual styles with constants.~
2. ~Verify rule sorting precedence in nested contextual styles with pseudo-selectors, regular at-rules, and at-rule constants (eg. ensure :hover inside @media sorts correctly).~
3. ~Investigate why processRules tests appear to wrap media queries unexpectedly even before this patch when using defineConsts, as tested [here](https://github.com/facebook/stylex/commit/c9710b141805970736bc330fb7451441bf920f97)~
    a. This is because we need to simulate running stylex.create calls first instead of inlining the fixture above

Follow up:
- Add type validation for constants, either via a style.types.media registry or a basic allowlist (eg. only accept @media, @supports, @container for now).

## testing
tested on example-nextjs
```
    color: {
      default: 'black',
      [breakpoints.small: {
        default: 'red',
        [breakpoints.big]: {
          default: 'orange',
          '@media (max-width: 1200px)': 'CadetBlue'
        },
      },
   }
```
https://github.com/user-attachments/assets/93c6ea9c-8085-422c-b117-31b14578bd2a

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -9,37 +9,57 @@
 
 jest.autoMockOff();
 
+import path from 'path';
 import { transformSync } from '@babel/core';
 import stylexPlugin from '../src/index';
 
 function transform(source, opts = {}) {
-  const { code, metadata } = transformSync(source, {
-    filename: opts.filename || '/stylex/packages/test.stylex.js',
-    parserOpts: {
-      flow: 'all',
+  const fixturesDir = path.join(__dirname, '__fixtures__');
+
+  const pluginOpts = {
+    ...opts,
+    unstable_moduleResolution: {
+      rootDir: fixturesDir,
+      type: 'commonJS',
     },
-    plugins: [
-      [
-        stylexPlugin,
-        {
-          unstable_moduleResolution: {
-            rootDir: '/stylex/packages/',
-            type: 'commonJS',
-          },
-          ...opts,
-        },
-      ],
-    ],
+  };
+
+  const main = transformSync(source, {
+    filename: path.join(fixturesDir, 'main.stylex.js'),
+    parserOpts: { sourceType: 'module' },
+    babelrc: false,
+    plugins: [[stylexPlugin, pluginOpts]],
   });
-  return { code, metadata };
+
+  const consts = transformSync(
+    `
+    import * as stylex from '@stylexjs/stylex';
+    export const constants = stylex.defineConsts({
+      YELLOW: 'yellow',
+      ORANGE: 'var(--orange)',
+      mediaBig: '@media (max-width: 1000px)',
+      mediaSmall: '@media (max-width: 500px)'
+    });
+    `,
+    {
+      filename: path.join(fixturesDir, 'constants.stylex.js'),
+      parserOpts: { sourceType: 'module' },
+      babelrc: false,
+      plugins: [[stylexPlugin, pluginOpts]],
+    },
+  );
+
+  const metadata = [
+    ...(main.metadata.stylex || []),
+    ...(consts.metadata.stylex || []),
+  ];
+
+  return { code: main.code, metadata };
 }
 
 const fixture = `
 import * as stylex from '@stylexjs/stylex';
-export const constants = stylex.defineConsts({
-  YELLOW: 'yellow',
-  ORANGE: 'var(--orange)'
-});
+import { constants } from './constants.stylex';
 export const styles = stylex.create({
   root: {
     animationName: stylex.keyframes({
@@ -55,7 +75,10 @@ export const styles = stylex.create({
     backgroundColor: 'red',
     borderColor: {
       default: 'green',
-      '@media (min-width:320px)': 'blue'
+      [constants.mediaBig]: {
+        default: 'blue',
+        [constants.mediaSmall]: 'yellow',
+      }
     },
     textShadow: {
       default: '1px 2px 3px 4px red',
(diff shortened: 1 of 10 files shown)


necolas:
I think we need to consider whether supporting things like infinite MQ nesting makes sense for StyleX. You'll rarely see that kind of code used in apps, and I suspect we have close to 0 instances of it at Meta. This kind of pattern will also not be supported on native. So, if we can get 99% of the features while only allowing one MQ (nested within `@supports` if necessary), then we should do that and keep both the compiler and the public API simpler.

necolas:
I'm also having a hard time understanding what each of the bullet points means in this issue

mellyeliu:
sorry, these draft PRs have notes for my own organization, i've reworded them now for clarity. this PR patches the issue explained above.

re: nested contextual styles - agree it’s worth discussing. i don’t think it will impact this diff much either way, since I think `PreRuleSet` handles this logic (just need to double check via testing)

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -38,6 +39,37 @@ function transform(source, opts = {}) {
   return { code, metadata };
 }
 
+function transformWithConsts(source, opts = {}) {
+  const { code, metadata } = transformSync(source, {
+    filename: path.join(__dirname, '__fixtures__/main.stylex.js'), // fake filename to help resolve
```
mellyeliu, line 44: separate fixture to better simulate babel transforms on separate file paths

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -9,37 +9,64 @@
...
 import { transformSync } from '@babel/core';
 import stylexPlugin from '../src/index';
 
-function transform(source, opts = {}) {
-  const { code, metadata } = transformSync(source, {
-    filename: opts.filename || '/stylex/packages/test.stylex.js',
-    parserOpts: {
-      flow: 'all',
+function transformWithConsts(source, opts = {}) {
+  const fixturesDir = path.join(__dirname, '__fixtures__');
```
necolas, line 17: Is `path` really needed for these tests?

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -9,37 +9,64 @@
...
+    ...opts,
+    unstable_moduleResolution: {
+      rootDir: fixturesDir,
+      type: 'commonJS',
+      resolveFilePath(importPath) {
+        if (importPath === './constants.stylex') {
+          return [
+            'themeNameRef',
+            path.join(fixturesDir, 'constants.stylex.js'),
+          ];
```
necolas, line 29: What's this?

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -9,37 +9,64 @@
...
+      parserOpts: { sourceType: 'module' },
+      babelrc: false,
+      plugins: [[stylexPlugin, pluginOpts]],
+    },
+  );
+
+  return stylexPlugin.processStylexRules(
+    [...(main.metadata.stylex || []), ...(consts.metadata.stylex || [])],
+    opts.useLayers,
+  );
```
necolas, line 64: I found it confusing that just in this test the return value of the "transform" function is actually the CSS generated by the process function.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -38,6 +39,37 @@ function transform(source, opts = {}) {
   return { code, metadata };
 }
 
+function transformWithConsts(source, opts = {}) {
```
necolas, line 42: Now in this file, we have another `transformWithConsts` function but this one returns the result of the transform.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -244,4 +276,212 @@ describe('@stylexjs/babel-plugin', () => {
...
+                "ltr": "var(--x1r2wpmh){.xbs0o1n.xbs0o1n{color:blue}}",
+                "rtl": null,
+              },
+              6000,
+            ],
+          ],
+        }
+      `);
+    });
+    test('inlines multiple media queries from constants.stylex', () => {
```
necolas, line 368:
```suggestion
    });

    test('inlines multiple media queries from constants.stylex', () => {
```

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -244,4 +276,212 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
   });
+  describe('[transform] stylex.defineConsts() in stylex.create() ', () => {
```
necolas, line 279:
```suggestion

  describe('use of constants in stylex.create() ', () => {
```

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -244,4 +276,212 @@ describe('@stylexjs/babel-plugin', () => {
...
+                "ltr": "var(--xr4bctk){.x1ru35j7.x1ru35j7{color:yellow}}",
+                "rtl": null,
+              },
+              6000,
+            ],
+          ],
+        }
+      `);
+    });
+    test('inlines nested media query from constants.stylex', () => {
```
necolas, line 426:
```suggestion
    });

    test('inlines nested media query from constants.stylex', () => {
```

packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/PreRule.js
```diff
@@ -68,15 +68,21 @@ export class PreRule implements IPreRule {
     return sortAtRules(unsortedAtRules);
   }
 
+  get constRules(): $ReadOnlyArray<string> {
+    return this.keyPath.filter((key) => key.startsWith('var(--'));
```
necolas, line 72: Is this how consts are marked as placeholders? We should start adding comments to explain code like this, because to me it looks like we're filtering for CSS variables and it's not obvious why consts are all assumed to start with `var(--`

packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/basic-validation.js
```diff
@@ -58,7 +58,14 @@ function validateConditionalStyles(
...
-    if (!(key.startsWith('@') || key.startsWith(':') || key === 'default')) {
+    if (
+      !(
+        key.startsWith('@') ||
+        key.startsWith(':') ||
+        key.startsWith('var(--') ||
+        key === 'default'
+      )
+    ) {
       throw new Error(messages.INVALID_PSEUDO_OR_AT_RULE);
```
necolas, line 69: Here too - if I look at the condition, it's unclear why starting not with `var(--` should result in `INVALID_PSEUDO_OR_AT_RULE` being thrown

necolas (changes requested):
> packages/babel-plugin/package-lock.json

This file shouldn't exist

packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/PreRule.js
mellyeliu: yeah, using `var(--` as a placeholder so we're still emitting valid css if somehow it doesn't get inlined

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
mellyeliu: this is just here to resolve the relative import `'./constants.stylex' in the fixture

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
mellyeliu: renamed for clarity! let me add a comment below on what both of these functions do and why they behave differently, because i needed to do some digging...

mellyeliu:
Some findings when I was investigating how to capture defineConsts behaviour in tests (documenting for posterity)

`transformWithMergedMetadata`
- Transforms both the main file and the constants file, then merges their metadata
- Babel's `transformSync` only transforms the file passed in (in this case, the fixture with the `create` call, not the `defineConsts` call), so the metadata accumulated from `defineConsts` is missing
- `processStylexRules` expects all metadata across files, so we run `transformSync` on both files

`transformWithInlineConsts`
- Only transforms the main file (relies on `evaluatePath` logic to statically resolve and inline imported constants)
- New fixtures file to better mirror real project structure
- Just tests that the plugin correctly inlines constant values at transform time

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
mellyeliu: leftover when I copied the function over, we don't need in this test

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -244,4 +276,215 @@ describe('@stylexjs/babel-plugin', () => {
...
+        };"
+      `);
+
+      expect(metadata).toMatchInlineSnapshot(`
+        {
+          "stylex": [
+            [
+              "xw8d3ix",
+              {
+                "ltr": ".xw8d3ix{background-color:var(--x180gk19)}",
```
necolas, line 310: Looks like this constant wasn't inlined?

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -244,4 +276,215 @@ describe('@stylexjs/babel-plugin', () => {
...
+              {
+                "ltr": ".x1e2nbdu{color:red}",
+                "rtl": null,
+              },
+              3000,
+            ],
+            [
+              "xbs0o1n",
+              {
+                "ltr": "var(--x1r2wpmh){.xbs0o1n.xbs0o1n{color:blue}}",
```
necolas, line 360: MQ wasn't inlined. This is also invalid CSS, so maybe the approach of using a CSS var as a placeholder is not a good one...at least for MQs

necolas (changes requested):
Several snapshots show the constants are not inlined. 

It might be worth rethinking the idea of using const placeholders that look like CSS code, given how confusing that is and that it doesn't prevent invalid CSS for MQs. It's really hard to tell by looking at a snapshot whether the output is correct or not, as sometimes the output should have CSS vars if the code was authored to use them.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
mellyeliu: this is expected, since this gets replaced later downstream in `processStylexRules`. what we want to see at this stage is wrapping it in var(--hash){}. but yes there's probably a better placeholder here, maybe something like: `@media (min-width: 0px /*--hash*/) { ... }` which is a no-op without the inlining. we can scan for this pattern in `processStylexRules` alongside the vars one

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
mellyeliu: this tests the visitor, which runs before `processStylexRules`

mellyeliu:
maybe we can explore something like:
- for all constant values: prefixing the hash with something like var(--STYLEX_CONST_<hash>) so it's semantically distinct from user-authored vars
- for MQs: @media (min-width: 0px /*--STYLEX_CONST_hash*/) { ... } or some similar no-op MQ

or we can simply use some placeholder like STYLEX_CONST(HASH) and add a compiler error if there's anything left after inlining in processStylexRules.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -38,6 +39,37 @@ function transform(source, opts = {}) {
...
+            rootDir: path.join(__dirname, '__fixtures__'),
+            type: 'commonJS',
+            resolveFilePath(importPath, _) {
+              if (importPath === './constants.stylex') {
+                return [
+                  'themeNameRef',
+                  path.join(__dirname, '__fixtures__/constants.stylex.js'),
+                ];
+              }
+              return null;
```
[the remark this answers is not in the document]
mellyeliu: this is an llm hallucination... removed

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -55,7 +75,10 @@ export const styles = stylex.create({
     backgroundColor: 'red',
     borderColor: {
       default: 'green',
-      '@media (min-width:320px)': 'blue'
+      [constants.mediaBig]: {
+        default: 'blue',
+        [constants.mediaSmall]: 'yellow',
+      }
```
necolas, line 81: This is another example of code we shouldn't allow

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -68,57 +91,104 @@ export const styles = stylex.create({
...
-        .xckgs0v:not(#\\#):not(#\\#){animation-name:xi07kvp-B}
+        .xdmqw5o:not(#\\#):not(#\\#){animation-name:x4ssjuf-B}
         .xrkmrrc:not(#\\#):not(#\\#){background-color:red}
         html:not([dir='rtl']) .x1skrh0i:not(#\\#):not(#\\#){text-shadow:1px 2px 3px 4px red}
         html[dir='rtl'] .x1skrh0i:not(#\\#):not(#\\#){text-shadow:-1px 2px 3px 4px red}
         @media (min-width:320px){html:not([dir='rtl']) .x1cmij7u.x1cmij7u:not(#\\#):not(#\\#){text-shadow:10px 20px 30px 40px green}}
-        @media (min-width:320px){html[dir='rtl'] .x1cmij7u.x1cmij7u:not(#\\#):not(#\\#){text-shadow:-10px 20px 30px 40px green}}"
+        @media (min-width:320px){html[dir='rtl'] .x1cmij7u.x1cmij7u:not(#\\#):not(#\\#){text-shadow:-10px 20px 30px 40px green}}
+        @media (max-width: 1000px){.x1l0eizu.x1l0eizu:not(#\\#):not(#\\#):not(#\\#){border-color:blue}}
+        @media (max-width: 500px){@media (max-width: 1000px){.x5i7zo.x5i7zo.x5i7zo:not(#\\#):not(#\\#):not(#\\#):not(#\\#){border-color:yellow}}}"
```
necolas, line 139: Not something to fix in this diff, but look at how bad this generated CSS is.

necolas (approved):

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineConsts-test.js
```diff
@@ -244,4 +267,215 @@ describe('@stylexjs/babel-plugin', () => {
...
+        import { breakpoints, colors } from './constants.stylex';
+
+        export const styles = stylex.create({
+          root: {
+            color: {
+              default: 'black',
+              [breakpoints.big]: {
+                default: colors.red,
+                [breakpoints.small]: colors.blue,
+              },
```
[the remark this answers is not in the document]
mellyeliu: meant to use max-width for `breakpoints.big`, but yes let's revisit this as part of the mq transform/validation work

PR #1055, opened 2025-05-08
Remove addSpecificityLevel

necolas:
Test impact on CSS bundle size of the addSpecificityLevel hack. Before adds the selectors. After removes them


necolas:
Confirmed minimal impact on compressed size

PR #1058, opened 2025-05-15
[docs] Add docs for positionTry API

necolas:


packages/docs/docs/api/javascript/positionTry.mdx
```diff
@@ -0,0 +1,60 @@
...
+});
+
+const styles = stylex.create({
+  anchor: {
+    positionTryFallbacks: fallback,
+  },
+});
+```
+
+To use `positionTry` return values in a separate file, you can use `defineVars` to hold animation names:
```
mellyeliu, line 41:
```suggestion
To use `positionTry` return values in a separate file, you can use `defineVars` to hold position fallback values:
```

packages/docs/docs/api/javascript/positionTry.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+Creates a [`@position-try`](https://developer.mozilla.org/en-US/docs/Web/CSS/@position-try) rule used to define a custom position-try fallback option, which can be used to define positioning and alignment for anchor-positioned elements. One or more sets of position-try fallback options can be applied to the anchored element via the `positionTryFallbacks` property.
+
+```ts
+function positionTry(descriptors: {[key: string]: RawStyles}): string;
+```
+
+The only properties allowed in a `positionTry` call are `positionAnchor`, `positionArea`, inset properties (`top`, `left`, `insetInline`, etc.), margin properties, size properties (`height`, `inlineSize`, etc.), and self-alignment properties (`alignSelf`, `justifySelf`, `placeSelf`).
+
+You must declare your position-try rules in the same file as where you use them. Duplicate declarations will be deduplicated in the generated CSS output.
```
mellyeliu, line 19: I think this is a copy-over from the wording in keyframes, but it feels misleading because later we give an example where you export them with `defineVars`

necolas: Yeah that seems to be a workaround for not being able to export them directly. Fine for now IMO

mellyeliu (approved):

PR #1064, opened 2025-05-16
[babel-plugin] Make 'property-specificity' the default 'styleResolution'

necolas:
Changing the default styleResolution enforces more disallowed shorthands and reduces the size of the generated JavaScript objects.

Summary of the benchmark results is below.

* 10% faster compilation of `create`.
* 35% less generated JS after minification.
* 8% less generated JS after compression.
* No significant change to size of CSS.
* 48% faster runtime merging by default (with caching).
* 630% faster runtime merging without caching.

Every metric we track is significantly improved or neutral as a result of this change.

Notes:

1. This change primarily affects styles using shorthands. Some of the benchmark fixtures (e.g., "basic create") contain a greater proportion of these styles than others, (e.g., "complex create"), which explains some of the differences in results.
2. The generated JS results are "synthetic", because the fixtures are almost 100% StyleX code and not StyleX mixed in with other JS as is the case in practice (e.g., when used in React components). In real-world scenarios we should expect to see far less impact on the total size of JS bundles, as StyleX-generated JS will be a smaller fraction of the total JS. But we can still conclude that the JS generated by StyleX will be reduced by about 35% in total.


mellyeliu (approved):
lgtm! js size improvements look great

maybe we queue this for the next release because the 0.13.0 has some breaking changes already

PR #1065, opened 2025-05-17
[postcss-plugin] Fix importSource object syntax not working properly

javascripter:

Using the `{ from, as }` object form in importSources caused a runtime error due to incorrect .includes usage in [Previous PR](https://github.com/facebook/stylex/pull/1063). This fix ensures both string and object forms are handled properly in shouldTransform.

## Linked PR/Issues

https://github.com/facebook/stylex/pull/1063

## Additional Context

In main branch, move to the example app
```bash
cd examples/example-nextjs
```

Using the object syntax throws when running `npm run example:dev`:
```javascript
   importSources: [
        {
          // Note:Just enough config to reproduce the error, not actually a proper config for stylex
          from: '@stylexjs/stylex',
          as: 'create',
        },
      ],
```

Switch to this branch and now it does not throw and renders correctly.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/postcss-plugin/src/bundler.js
+++ b/packages/@stylexjs/postcss-plugin/src/bundler.js
@@ -20,7 +20,7 @@ module.exports = function createBundler() {
       if (typeof importSource === 'string') {
         return sourceCode.includes(importSource);
       }
-      return importSource.includes(sourceCode.from);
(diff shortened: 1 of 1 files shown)


necolas (approved):
We need unit tests for this plugin

Issue #1034, opened 2025-05-01
0.13.0 task list


necolas:
List of tasks to complete before releasing 0.13.0

- [x] Deprecate `@stylexjs/shared` package on npm.
- [x] Stop publishing `@stylexjs/style-value-parser` to npm.
- [x] Replace `v0.12.0` tag with `0.12.0`.
- [x] Decide other features to include in 0.13.0 
  - [x] babel-plugin
    - [ ] ~https://github.com/facebook/stylex/pull/1005~
    - [x] https://github.com/facebook/stylex/pull/1059
    - [x] https://github.com/facebook/stylex/pull/1019
    - [ ] ~https://github.com/facebook/stylex/pull/856/~
  - [x] eslint-plugin
    - [x] https://github.com/facebook/stylex/pull/1039
- [x] Make sure docs are up-to-date.
  - [x] `defineConsts` docs
    - [x] https://github.com/facebook/stylex/pull/1046
  - [x] `positionTry` docs
    - [x] https://github.com/facebook/stylex/pull/1058
- [x] Update CHANGELOG as part of the `0.13.0` commit.
  - [x] https://github.com/facebook/stylex/pull/1060

mellyeliu:
- We could do a patch release in the meantime, since we made the fix for `firstThatWorks` (https://github.com/facebook/stylex/pull/1009) and the style injection improvements (https://github.com/facebook/stylex/pull/1017), and deprecate `@stylexjs/shared` alongside the patch 

- I'll update `@stylexjs/style-value-parser` to be private and unpublish the npm package, since both `@stylexjs/eslint-plugin` and `@stylexjs/babel-plugin`  will use it

necolas:
Going to punt on https://github.com/facebook/stylex/pull/1005 as defineConsts doesn't fit well outside the compiler. Also punting #856 due to lack of follow up

mellyeliu:
published `0.13.0` to npm

PR #1066, opened 2025-05-18
[release] v0.13.0

mellyeliu:
```
$ npm run release -- -v 0.13.0
```
since https://github.com/facebook/stylex/issues/1034 is wrapped up

will add release notes in separate pr

### DIFF
--- a/examples/example-cli/package.json
+++ b/examples/example-cli/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "example-cli",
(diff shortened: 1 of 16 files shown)


necolas (changes requested):
There's also a patch to the postcss-plugin that should be added to the changelog

mellyeliu:
modified the changelog here: https://github.com/facebook/stylex/pull/1068

Issue #1062, opened 2025-05-15
[babel-plugin] Better options for processStylexRules


necolas:
In particular, we should be able to support generating CSS files that:

* Order CSS correctly but don't use `@layer` or a polyfill.
  * CSS layers aren't needed for many use cases.
* Use native `@layer`.
  * Useful for explicitly controlling order, including when generating multiple sheets. 
* Polyfill `@layer`.
  * If layers are needed but browser support is not adequate.
  * Use PostCSS instead of rolling our own solution: https://www.npmjs.com/package/@csstools/postcss-cascade-layers

jeongminsang:
fixed this issue! https://github.com/facebook/stylex/pull/1072

nmn:
> Order CSS correctly but don't use @layer or a polyfill.

This is not entirely possible. Based on the StyleX design for media queries and pseudo classes, all longhand styles should have higher specificity than shorthand styles.

This is also needed for RN support to work correctly.

Therefore, we need a plain `.margin-top` to override a `.margin:hover`

This means we need some way to bump the specificity, which means a `@layer` or a polyfill for bumping specificity.

jeongminsang:
I've carefully reviewed the feedback on the [#1072](https://github.com/facebook/stylex/pull/1072) and am currently working on addressing the points. I'll update the PR soon.

@nmn,

I now fully understand that for StyleX's design and RN support, ensuring longhand styles override shorthand styles requires achieving higher specificity, which in turn necessitates the use of `@layer` or a polyfill.

Considering this situation, I'd appreciate hearing the maintainers' thoughts on the future direction of this issue.

* Would it be best to proceed with the current [#1072](https://github.com/facebook/stylex/pull/1072), excluding the "order CSS correctly without `@layer` or a polyfill" requirement?
* Alternatively, regarding the StyleX specificity issue that would require `@layer` or a polyfill, I'm open to discussing and exploring solutions if maintainers can provide clear guidelines.

I'm keen to help resolve this issue.

necolas:
You can do it without adding "none" for now. We don't use any layers stuff internally, but that won't be the case for long so we might not need "none" for long anyway

PR #1072, opened 2025-05-21
[postcss-plugin] Add @layer and polyfill support via @csstools/postcss-cascade-layers

jeongminsang:

Previously, useCSSLayers offered limited choices, and its polyfill implementation was handled internally. 
This update replaces the custom polyfill with a well-maintained solution: @csstools/postcss-cascade-layers.

This change allows us to effectively support the following CSS Layer generation scenarios:

Order CSS correctly without @layer or a polyfill
Use native @layer
Polyfill @layer

## Linked PR/Issues

Fixes https://github.com/facebook/stylex/issues/1062

## Additional Context

useCSSLayers Option Refactoring

'none' (default): Handles CSS specificity using :not(#\\#) without relying on @layer or a polyfill.
'native': Utilizes native CSS @layer for specificity control. (This previously corresponded to useCSSLayers: true).
'polyfill': Implements CSS layer polyfilling using @csstools/postcss-cascade-layers. (This previously corresponded to useCSSLayers: false for the custom polyfill).
Example configuration:
```jsx
//postcss.config.js
module.exports = {
  plugins: {
    "@stylexjs/postcss-plugin": {
      include: [
        "./**/*.{js,jsx,ts,tsx}",
        "app/**/*.{js,jsx,ts,tsx}",
        "components/**/*.{js,jsx,ts,tsx}",
      ],
      useCSSLayers: "none",  // "none" | "native" | "polyfill"
    },
    autoprefixer: {},
  },
};
```

Polyfill Implementation Details
The previously self-implemented polyfill logic has been removed:

```jsx
// babal-plugin/src/index.js
if (!useLayers) {
	ltrRule = addSpecificityLevel(ltrRule, index);
	rtlRule = rtlRule && addSpecificityLevel(rtlRule, index);
}
```
The polyfill functionality is now powered by @csstools/postcss-cascade-layers, ensuring a more robust and compliant solution:

```jsx
// postcss-plugin/src/index.js
let processedCss = css;
if (useCSSLayers === 'polyfill') {
  const result = await postcss([
    cascadeLayers({
      onRevertLayerKeyword: 'warn',
      onConditionalRulesChangingLayerOrder: 'warn',
      onImportLayerRule: 'warn',
    }),
  ]).process(css);
  processedCss = result.css;
}
```

Transfile Example
[Testcode](https://github.com/facebook/stylex/compare/main...jeongminsang:stylex:main?expand=1#diff-a80b7764f34f4913554600107b195c7b9dd11fcf9adf590ff7672eb9ccdace53)

This change introduces breaking changes due to the modification of the useCSSLayers option's type and expected values. If backward compatibility is a critical concern, alternative implementation approaches might need to be considered.
How about we rework this, considering the breaking changes?

### DIFF
--- a/packages/@stylexjs/cli/src/config.js
+++ b/packages/@stylexjs/cli/src/config.js
@@ -23,7 +23,7 @@ export type CliConfig = {
   babelPluginsPre?: $ReadOnlyArray<any>,
   babelPluginsPost?: $ReadOnlyArray<any>,
   modules_EXPERIMENTAL: $ReadOnlyArray<ModuleType>,
-  useCSSLayers?: boolean,
+  useCSSLayers?: 'none' | 'native' | 'polyfill',
   styleXConfig?: StyleXOptions,
 };
 
--- a/packages/@stylexjs/cli/src/index.js
+++ b/packages/@stylexjs/cli/src/index.js
@@ -71,7 +71,7 @@ const modules_EXPERIMENTAL: $ReadOnlyArray<ModuleType> =
 const babelPresets: $ReadOnlyArray<any> = args.babelPresets;
 const babelPluginsPre: $ReadOnlyArray<any> = args.babelPluginsPre;
 const babelPluginsPost: $ReadOnlyArray<any> = args.babelPluginsPost;
-const useCSSLayers: boolean = args.useCSSLayers;
+const useCSSLayers: 'none' | 'native' | 'polyfill' = args.useCSSLayers;
 const styleXConfig: StyleXOptions = (config.styleXConfig as $FlowFixMe) ?? {};
 
 const cliArgsConfig: CliConfig = {
--- a/packages/@stylexjs/cli/src/options.js
+++ b/packages/@stylexjs/cli/src/options.js
@@ -35,8 +35,8 @@ const options = {
   useCSSLayers: {
     alias: 'l',
     describe: 'Use CSS layers to optimize CSS rendering',
-    type: 'boolean',
-    default: false,
+    type: 'string',
+    default: 'none',
   },
   babelPresets: {
     describe:
--- a/packages/@stylexjs/cli/src/transform.js
+++ b/packages/@stylexjs/cli/src/transform.js
@@ -81,7 +81,7 @@ export async function compileDirectory(
 
   const compiledCSS = await styleXPlugin.processStylexRules(
     Array.from(config.state.styleXRules.values()).flat(),
-    config.useCSSLayers,
+    config.useCSSLayers !== 'none',
   );
 
   const cssBundlePath = path.join(config.output, config.styleXBundleName);
--- a/packages/@stylexjs/postcss-plugin/README.md
+++ b/packages/@stylexjs/postcss-plugin/README.md
@@ -130,8 +130,13 @@ and specify desired options. Refer to
 ### useCSSLayers
 
 ```js
-useCSSLayers: boolean; // Default: false
+useCSSLayers: 'none' | 'native' | 'polyfill'; // Default: 'none'
 ```
 
-Enabling this option switches Stylex from using `:not(#\#)` to using `@layers`
-for handling CSS specificity.
+The `useCSSLayers` option controls how StyleX handles CSS specificity and layer
+management. It supports three strategies:
+
+- `'none'` (default): Uses `:not(#\#)` to handle CSS specificity without layers
+- `'native'`: Uses native CSS `@layer` for handling CSS specificity
+- `'polyfill'`: Uses `@csstools/postcss-cascade-layers` to polyfill CSS layers
+  for browsers that don't support them
--- a/packages/@stylexjs/postcss-plugin/package.json
+++ b/packages/@stylexjs/postcss-plugin/package.json
@@ -15,5 +15,8 @@
     "fast-glob": "^3.3.2",
     "glob-parent": "^6.0.2",
     "is-glob": "^4.0.3"
+  },
+  "devDependencies": {
+    "@csstools/postcss-cascade-layers": "^5.0.1"
   }
 }
--- a/packages/@stylexjs/postcss-plugin/package.json
+++ b/packages/@stylexjs/postcss-plugin/package.json
@@ -11,12 +11,10 @@
   "dependencies": {
     "@babel/core": "^7.26.8",
     "@stylexjs/babel-plugin": "0.12.0",
+    "@csstools/postcss-cascade-layers": "^5.0.1",
     "postcss": "^8.4.49",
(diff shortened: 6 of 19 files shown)


packages/@stylexjs/cli/src/config.js
```diff
@@ -23,7 +23,7 @@ export type CliConfig = {
   babelPluginsPre?: $ReadOnlyArray<any>,
   babelPluginsPost?: $ReadOnlyArray<any>,
   modules_EXPERIMENTAL: $ReadOnlyArray<ModuleType>,
-  useCSSLayers?: boolean,
+  useCSSLayers?: 'none' | 'native' | 'polyfill',
```
necolas, line 26:
```suggestion
  layers?: 'none' | 'native' | 'polyfill',
```

packages/@stylexjs/cli/src/index.js
```diff
@@ -71,7 +71,7 @@ const modules_EXPERIMENTAL: $ReadOnlyArray<ModuleType> =
 const babelPresets: $ReadOnlyArray<any> = args.babelPresets;
 const babelPluginsPre: $ReadOnlyArray<any> = args.babelPluginsPre;
 const babelPluginsPost: $ReadOnlyArray<any> = args.babelPluginsPost;
-const useCSSLayers: boolean = args.useCSSLayers;
+const useCSSLayers: 'none' | 'native' | 'polyfill' = args.useCSSLayers;
```
necolas, line 74:
```suggestion
const layers: 'none' | 'native' | 'polyfill' = args.layers;
```

packages/@stylexjs/cli/src/options.js
```diff
@@ -35,8 +35,8 @@ const options = {
   useCSSLayers: {
```
necolas, line 35:
```suggestion
  layers: {
```

packages/@stylexjs/cli/src/transform.js
```diff
@@ -81,7 +81,7 @@ export async function compileDirectory(
 
   const compiledCSS = await styleXPlugin.processStylexRules(
     Array.from(config.state.styleXRules.values()).flat(),
-    config.useCSSLayers,
+    config.useCSSLayers !== 'none',
```
necolas, line 84: This arg should be changed to an options object, and here we'd pass `{ layers }`

packages/@stylexjs/postcss-plugin/README.md
```diff
@@ -130,11 +130,16 @@ and specify desired options. Refer to
 ### useCSSLayers
 
 ```js
-useCSSLayers: boolean; // Default: false
+useCSSLayers: 'none' | 'native' | 'polyfill'; // Default: 'none'
```
necolas, line 133: We should output layers by default

```suggestion
layers: 'none' | 'native' | 'polyfill'; // Default: 'native'
```

packages/@stylexjs/postcss-plugin/README.md
```diff
@@ -130,11 +130,16 @@ and specify desired options. Refer to
...
 ```
 
-Enabling this option switches Stylex from using `:not(#\#)` to using `@layers`
-for handling CSS specificity.
+The `useCSSLayers` option controls how StyleX handles CSS specificity and layer
+management. It supports three strategies:
```
necolas, line 137:
```suggestion
The `layers` option controls how StyleX handles CSS order. It supports three strategies:
```

packages/@stylexjs/postcss-plugin/README.md
```diff
@@ -130,11 +130,16 @@ and specify desired options. Refer to
...
+
+- `'none'` (default): Uses `:not(#\#)` to handle CSS specificity without layers
```
necolas, line 139: `none` should not add layers at all and rely only on generated ordering using the "priority"

packages/@stylexjs/postcss-plugin/__tests__/index-test.js
```diff
@@ -56,19 +59,37 @@ describe('@stylexjs/postcss-plugin', () => {
   });
 
   test('supports CSS layers', async () => {
-    const result = await runStylexPostcss({ useCSSLayers: true });
+    const result = await runStylexPostcss({ useCSSLayers: 'native' });
```
necolas, line 62:
```suggestion
    const result = await runStylexPostcss({ layers: 'native' });
```

packages/@stylexjs/postcss-plugin/__tests__/index-test.js
```diff
@@ -56,19 +59,37 @@ describe('@stylexjs/postcss-plugin', () => {
...
+      @layer priority2{
+      .xg6iff7{min-height:100vh}
+      .x1gan7if{padding-bottom:32px}
+      .x1miatn0{padding-top:32px}
       }"
     `);
   });
 
+  test('supports polyfill for CSS layers', async () => {
+    const result = await runStylexPostcss({ useCSSLayers: 'polyfill' });
```
necolas, line 81:
```suggestion
    const result = await runStylexPostcss({ layers: 'polyfill' });
```

packages/@stylexjs/postcss-plugin/package.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "@babel/core": "^7.26.8",
     "@stylexjs/babel-plugin": "0.13.0",
+    "@csstools/postcss-cascade-layers": "^5.0.1",
```
necolas, line 17: This polyfill should be used by `processStylexRules` rather than the postcss plugin

packages/@stylexjs/postcss-plugin/src/builder.js
```diff
@@ -151,7 +151,7 @@ function createBuilder() {
       }),
     );
 
-    const css = bundler.bundle({ useCSSLayers });
+    const css = bundler.bundle({ useCSSLayers: useCSSLayers !== 'none' });
```
necolas, line 154:
```suggestion
    const css = bundler.bundle({ layers });
```

necolas (changes requested):
Thanks for giving this a try.

The layers polyfill should be part of the babel-plugin `processStylexRules` method. The unit tests for that API should also be updated to snapshot the output from the 3 different options. The other packages like the CLI and postcss-plugin can then call into processStylexRules and pass an options object instead of the current boolean.

jeongminsang:
@necolas,
I'm considering implementing the polyfill for  [@csstools/postcss-cascade-layers](https://www.npmjs.com/package/@csstools/postcss-cascade-layers)  within the processStylexRules method, as per your review. However, this change would require processStylexRules to become an async function, which could lead to significant breaking changes across the codebase. Are these extensive changes what you intended?

PR #1078, opened 2025-05-23
feat: add initial implementation for stylex.viewTransitionClass API

vincentriemer:
## Overview

This PR adds an initial implementation of the `stylex.viewTransitionClass` API as described in https://github.com/facebook/stylex/discussions/1011. This implementation is not feature complete, most notably missing the ability to integrate media queries. It does however include the integrations necessary for accepting `stylex.keyframes` animations and accepting constants.

Adding things like media query integration in a scalable way (i.e. not duplicating the implementation that already exists for `stylex.create`) would likely require a non-trivial refactor of `stylex.create`'s transform so I'm putting this change up as an initial step and people can start experimenting with it.

### Unit Tests

Added new unit tests which are passing

### Manual Testing

Locally modified the NextJS example project & `Counter.tsx` file to utilize view transitions and verified that it is rendering correctly. I did not include these changes in this PR as it requires an experimental NextJS configuration.

Here's what the new source of `Counter.tsx` looks like:

```ts
/**
 * Copyright (c) Meta Platforms, Inc. and affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 *
 *
 */

'use client';

import * as stylex from '@stylexjs/stylex';
import { spacing, text, globalTokens as $ } from './globalTokens.stylex';
import { colors } from '@stylexjs/open-props/lib/colors.stylex';
import {
  startTransition,
  useState,
  unstable_ViewTransition as ViewTransition,
  unstable_addTransitionType as addAnimationType,
} from 'react';

const animationDuration = '300ms';
const animationTimingFunction = 'cubic-bezier(0.25, 1, 0.5, 1)'; // ease-out-quart

const fadeInFromLeft = stylex.keyframes({
  from: {
    opacity: 0,
    transform: 'translateX(30px)',
  },
  to: {
    opacity: 1,
    transform: 'translateX(0px)',
  },
});

const fadeInFromRight = stylex.keyframes({
  from: {
    opacity: 0,
    transform: 'translateX(-30px)',
  },
  to: {
    opacity: 1,
    transform: 'translateX(0px)',
  },
});

const fadeOutToRight = stylex.keyframes({
  from: {
    opacity: 1,
    transform: 'translateX(0px)',
  },
  to: {
    opacity: 0,
    transform: 'translateX(-30px)',
  },
});

const fadeOutToLeft = stylex.keyframes({
  from: {
    opacity: 1,
    transform: 'translateX(0px)',
  },
  to: {
[...]

PR #1081, opened 2025-05-27
feat: allow inline keyframe expressions inside viewTransitionClass

vincentriemer:
## Overview

In https://github.com/facebook/stylex/pull/1078 I added an initial implementation of `viewTransitionClass` and when I was doing some additional manual exploration I noticed that using the `keyframes` API inline in the `viewTransitionClass` object wasn't working so this diff resolves that by adding support. This is a small thing but can really help keep the view transition configuration (relatively) consise.

## Test Plan

Added unit test covering the scenario and it is passing.

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-viewTransitionClass-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-viewTransitionClass-test.js
@@ -65,6 +65,7 @@ describe('@stylexjs/babel-plugin', () => {
         }
       `);
     });
+
     test('local variables used in view transition class', () => {
       const { code, metadata } = transform(`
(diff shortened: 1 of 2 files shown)


packages/@stylexjs/babel-plugin/__tests__/transform-stylex-viewTransitionClass-test.js
```diff
@@ -156,6 +155,61 @@ describe('@stylexjs/babel-plugin', () => {
         }
       `);
     });
+    test('using inline keyframes', () => {
```
mellyeliu, line 158:
```suggestion
    });
    
    test('using inline keyframes', () => {
```

necolas (approved):

Issue #752, opened 2024-10-15
Do not compile logical styles into `ltr` and `rtl` equivalents.


nmn:
Currently, when using `styleResolution: 'legacy-expand-shorthands'`, CSS logical properties are compiled into two `ltr` and `rtl` equivalent CSS rules.

All new browsers natively support logical CSS properties, so we should just use those.

The one exception is `border-*-*-radius`, which is not supported in older browsers.
In that case, we should polyfill it after CSS generation if it makes sense.

nmn:
@mobalti Could you please explain why this is a priority for you? `styleResolution: 'legacy-expand-shorthands'` is something that no one should be using and it's a setting that we use internally for esoteric reasons.

If you use any other value for `styleResolution` we already do not generate separate `ltr` and `rtl` styles.

Do you have a usecase for using `styleResolution: 'legacy-expand-shorthands'` in your own codebase?

mobalti:
I thought this issue was related to a discussion I started a few months ago about why the library converts all logical properties to their physical equivalents, such as changing "block-size" to "height" and "inline-size" to "width."

It would be helpful if the library offered an option to disable this feature because logical properties represent how we think when we design with CSS, not just browser behavior.

Thank you for considering this.

nmn:
> the library converts all logical properties to their physical equivalents, such as changing "block-size" to "height" and "inline-size" to "width."

@mobalti This is not completely accurate. Logical styles such as `margin-inline-start` etc are preserved. The current limitation is that StyleX does not support vertical writing modes. Logical styles along the horizontal axis are fully supported.

We are planning to fix this limitation.

PR #1083, opened 2025-05-28
[babel-plugin] add `enableLegacyDirectionPolyfill` config to legacy-e…

mellyeliu:
## Context
Let's remove the LTR/RTL polyfill (https://github.com/facebook/stylex/issues/752), as most browsers support logical styles now:
- [margin-inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-start)

`legacy-expand-shorthands` works in a few steps: 
1. In `legacy-expand-shorthands.js`: We split shorthands into respective longhands. There are no shorthands left after this.
2. In `/physical-rtl` we split these into ltr: `marginLeft` and rtl: `marginRight`. 

Follow ups:
- In 1., we do some transforms (`marginInlineStart` -> to RN equivalent `marginStart`, I **believe** this is for legacy reasons, before RN rolled out support for standard inline properties?) that we need to remove
- We should consider adding an exception to [`border-*-*-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-start-start-radius#browser_compatibility) properties, though it's on the edge of browser support. They fit the internal browser requirements for every app except WhatsApp.

I'm setting `enableLegacyDirectionPolyfill` to true to begin until the above items are complete.

## Testing
image

For these tests, we expect to see the LTR/RTL polyfill when `enableLegacyDirectionPolyfill: true`. We expect to see the equivalent (directional for vertical mode, logical for  longhands when `enableLegacyDirectionPolyfill: false`.

Note: We want to migrate over to `application-order` (and later `property-specificity`), but it'll be a long process to test and ship `@layer`. In the meantime, let's remove the logical styles polyfill as a way to test experimental stylesheets. We use the `legacy-expand-shorthands` style resolution  over `application-order` or `property-specificity` internally for this reason:
- `legacy-expand-shorthands` gets rid of the most common shorthands and expands them, since without `@layer` use in `processStylexRules`, shorthands get overridden by later longhands


necolas (commented):
The mapping to marginStart etc is not really for RN, it's just the legacy non-standard names we used internally before moving into the standard ones.

We could also move the related polyfill tests into the 'legacy' directory

packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -218,6 +218,14 @@ export default class StateManager {
         'options.enableMinifiedKeys',
       );
 
+    const enableLegacyDirectionPolyfill: StyleXStateOptions['enableLegacyDirectionPolyfill'] =
+      z.logAndDefault(
+        z.boolean(),
+        options.enableLegacyDirectionPolyfill ?? true,
+        true,
+        'options.enableLegacyDirectionPolyfill',
```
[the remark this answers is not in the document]
necolas: I don't think it's implying that we're adding a feature. We've already done this for several other flags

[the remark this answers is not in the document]
mellyeliu: sorry, still a bit confused - can you clarify? i think my PR desc might also have been unclear

this flag (when `true`) preserves the current behaviour that converts logical styles into physical ones, so it's not really a polyfill _for_ logical styles. when `false`, it keeps user authored logical properties as is. just thinking the name might sound like we're polyfilling support for them rather than removing them and inserting physical properties instead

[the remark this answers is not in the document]
necolas: Removing them and using physical properties is just how the polyfill currently works

packages/@stylexjs/babel-plugin/__tests__/legacy/transform-logical-properties-test.js
```diff
@@ -640,4 +706,1042 @@ describe('@stylexjs/babel-plugin', () => {
...
+              styleResolution: 'legacy-expand-shorthands',
+              enableLegacyDirectionPolyfill: true,
+            },
+          ),
+        ).toMatchInlineSnapshot(`
+          "import _inject from "@stylexjs/stylex/lib/stylex-inject";
+          var _inject2 = _inject;
+          import stylex from 'stylex';
+          _inject2(".x1hm9lzh{margin-left:10px}", 3000, ".x1hm9lzh{margin-right:10px}");
+          _inject2(".x1sa5p1d{margin-right:10px}", 3000, ".x1sa5p1d{margin-left:10px}");
```
necolas, line 730: Is it expected that the classname hash is the same for 2 different rules?

packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -218,6 +219,14 @@ export default class StateManager {
         'options.enableMinifiedKeys',
       );
 
+    const enableLegacyDirectionPolyfill: StyleXStateOptions['enableLegacyDirectionPolyfill'] =
+      z.logAndDefault(
+        z.boolean(),
+        options.enableLegacyDirectionPolyfill ?? true,
```
necolas, line 225:
```suggestion
        options.enableLegacyDirectionPolyfill ?? false,
```

packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
```diff
@@ -53,19 +53,19 @@ describe('Legacy-shorthand-expansion resolution', () => {
         var _inject2 = _inject;
         import stylex from 'stylex';
         _inject2(".x123j3cw{padding-top:5px}", 4000);
-        _inject2(".x1mpkggp{padding-right:5px}", 3000, ".x1mpkggp{padding-left:5px}");
+        _inject2(".x1gabggj{padding-right:5px}", 3000, ".x1gabggj{padding-left:5px}");
```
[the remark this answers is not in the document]
mellyeliu: just dug into this a bit. in the `stylex-create` visitor, we build the hash using the property name after shorthand expansion (`padding` -> `padding-top`, `padding-start`), but before ltr/rtl polyfill (see below). 

the hashes changed because i made a change to expand to `padding-inline-start` instead of `padding-start`, but the polyfill later changes it back to `paddingLeft` and `paddingRight`

packages/@stylexjs/babel-plugin/src/shared/physical-rtl/generate-ltr.js
```diff
@@ -7,94 +7,148 @@
...
+};
+
 const propertyToLTR: $ReadOnly<{
   [key: string]: ($ReadOnly<[string, string]>) => $ReadOnly<[string, string]>,
 }> = {
   'margin-start': ([_key, val]: $ReadOnly<[string, string]>) => [
     'margin-left',
     val,
   ],
-  // 'margin-inline-start': ([_key, val]) => ['margin-left', val],
```
mellyeliu, line 22: these are previous "aliases" that didn't need to be explicitly included, because we converted instances of `marginInlineStart` to `marginStart` while expanding shorthands. since we're expanding to `marginInlineStart` now, we need to explicitly include this in the mapping

packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/legacy-expand-shorthands.js
```diff
@@ -245,34 +245,34 @@ const shorthands: $ReadOnly<{ [key: string]: (TStyleValue) => TReturn }> = {
 
     return [
       ['marginTop', top],
-      ['marginEnd', right],
+      ['marginInlineEnd', right],
       ['marginBottom', bottom],
-      ['marginStart', left],
```
mellyeliu, line 250: this is where we previously expanded `margin` to `marginStart` anyway, since it all got translated to `marginLeft`/`marginRight` downstream

packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
```diff
@@ -53,19 +53,19 @@ describe('Legacy-shorthand-expansion resolution', () => {
         var _inject2 = _inject;
         import stylex from 'stylex';
         _inject2(".x123j3cw{padding-top:5px}", 4000);
-        _inject2(".x1mpkggp{padding-right:5px}", 3000, ".x1mpkggp{padding-left:5px}");
+        _inject2(".x1gabggj{padding-inline-end:5px}", 3000);
```
necolas, line 56: Maybe we should turn the feature flag on for these tests, if they're basically to test that the way we current use StyleX internally is not changed?

necolas (approved):

PR #1087, opened 2025-05-30
[babel-plugin] Avoid putting certain @-rules in @layer

### LINKED ISSUE #1071: [babel-plugin] CSS `@layer` code containing `@property`
### Describe the issue

See https://github.com/facebook/react-strict-dom/pull/308#discussion_r2097011056

Although this appears to be valid CSS, it causes a warning in Expo and may be tripping up some CSS validators. If we don't need the `@property` code in a layer, it might be a good idea to avoid that for now.

```
| Warning: Unknown at rule: @property (src/app/stylex.css:16:10):
|   14 |
|   15 |  @layer priority1{
| > 16 | @property --backgroundColor { syntax: "*"; inherits: false;}
|      |          ^
|   17 | @property --opacity { syntax: "*"; inherits: false;}
|   18 | @property --transform { syntax: "*"; inherits: false;}
|   19 | :root, .xxw3bu5{--primary100-x1ud21w4:black;--primary200-x19lv8z8:white;}
| Warning: Unknown at rule: @property (src/app/stylex.css:17:10):
|   15 |  @layer priority1{
|   16 | @property --backgroundColor { syntax: "*"; inherits: false;}
| > 17 | @property --opacity { syntax: "*"; inherits: false;}
```

### Expected behavior

Compatibility with Expo would be improved if `@property` declarations were not nested inside `@layer`

### Steps to reproduce

See above

### Test case

_No response_

### Additional comments

_No response_

necolas:
Not all CSS validators understand @property, @keyframes, @position-try inside @layer.

Fix #1071


packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineVars-test.js
```diff
@@ -506,7 +506,7 @@ describe('@stylexjs/babel-plugin', () => {
                 "ltr": ":root, .xop34xu{--xwx8imx:red;}",
                 "rtl": null,
               },
-              0,
+              0.1,
```
mellyeliu, line 509: what does the 0.1 represent/come from?

necolas: There's code elsewhere that multiplies the "at rule priority" by 0.1 😂 And I changed the default at rule priority from 0 to 1, so we get 0.1 here.

mellyeliu: classic...

PR #1092, opened 2025-06-05
[babel-plugin] polyfill all nonstandard properties in 'legacy-expand-…

mellyeliu:
Adding some missing polyfills for some of the more obscure nonstandard properties we use internally. Previously these would get translated to LTR/RTL, but we need to add mappings to standard logical properties. 

- borderStart
- start, end -> insetInlineStart, insetInlineEnd
- float: start, end -> float: inline-start, inline-end
- borderTopStartRadius -> borderStartStartRadius

Synced with the flag turned on and off internally, all screenshot tests passing. Added some tests to to document the expected behaviour. These tests are a bit verbose and the structure hard to navigate, and I'm hoping we can clean them up and reorganize them soon after we ship logical styles.
- `transform-legacy-polyfills-test` contains tests for all nonstandard properties
   - added both variants of `legacy-expand-shorthands`
- `stylex-transform-legacy-shorthands-test.js` contains all expanded shorthands
   - added `enableLogicalStylesPolyfill: false` tests

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
@@ -28,7 +28,7 @@ function transform(source, opts = {}) {
         {
           runtimeInjection: true,
           styleResolution: 'legacy-expand-shorthands',
-          enableLogicalStylesPolyfill: true,
+          enableLogicalStylesPolyfill: opts.enableLogicalStylesPolyfill ?? true,
           ...opts,
         },
       ],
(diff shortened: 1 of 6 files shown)


packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
```diff
@@ -462,3 +462,475 @@ describe('Legacy-shorthand-expansion resolution', () => {
     });
   });
 });
+
+describe('legacy-shorthand-expansion resolution (enableLogicalStylesPolyfill: false)', () => {
+  describe('while using RN non-standard shorthands', () => {
```
necolas: `margin` and `padding` aren't RN non-standard shorthands

packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
```diff
@@ -462,3 +462,475 @@ describe('Legacy-shorthand-expansion resolution', () => {
...
+
+    test('padding: with longhand property collisions', () => {
+      expect(
+        transform(
+          `
+          import stylex from 'stylex';
+          const styles = stylex.create({
+            foo: {
+              padding: 5,
+              paddingEnd: 10,
```
necolas, line 476: We didn't already codemod these away? Is there a reason to do this rather than finish the code migration first?

packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
```diff
@@ -462,3 +462,475 @@ describe('Legacy-shorthand-expansion resolution', () => {
...
+          export const styles = stylex.create({
+            foo: {
+              paddingInline: 5
+            }
+          });
+          stylex(styles.foo);
+        `,
+          { enableLogicalStylesPolyfill: false },
+        ),
+      ).toMatchInlineSnapshot(`
```
necolas: Tests like this don't see to be in the right place, if this file is for testing "legacy shorthands"

necolas (approved):
Stamping to unblock. This will be faster than blocking on codemodding. And all these tests for legacy features will get deleted eventually

mellyeliu:
Most of the common rules `margin/padding/border` are codemodded, but there are a few dozen stragglers from autogen code or from folders where the lint rule is not enabled. `valid-styles` also never had autofixes for the less common nonstandard properties, only `margin/padding/border` properties. Opened a PR to add autofixes for the remaining properties: https://github.com/facebook/stylex/pull/1093

Issue #1070, opened 2025-05-20
TypeScript error while using `stylex.create` after 0.13.0 update - the inferred type cannot be named without a reference


pawelblaszczyk5:
### Describe the issue

Hello 👋
It seems to be somewhat similar to https://github.com/facebook/stylex/issues/857 - in 0.13.0 entry point with StyleXTypes were removed. When using `stylex.create` I'm getting type errors because TypeScript can't generated type without having access to e.g. `StyleXClassNameFor` (in my tests re-exporting it from package fixed this particular error but I'm unsure whether there're some more cases where it'd fail.

If I use any `stylex.create` and TS is configured to generate types for given package it errors:

![Image](https://github.com/user-attachments/assets/ca0d3144-ca26-4be5-9368-dbd159150bc2)

![Image](https://github.com/user-attachments/assets/ad35ea4c-96ab-4c4e-816e-c03b9620eebc)

### Expected behavior

Project can be properly typechecked via tsc when declaration generation is enabled and `stylex.create` results are exported

### Steps to reproduce

1. Install StyleX
2. Enable declaration generating in your tsconfig by setting declaration to true
3. Try to create any styles with stylex and export them 
4. Observe error in IDE/tsc failing

I've created small reproduction: https://github.com/pawelblaszczyk5/stylex-incorrect-types-repro

### Test case

```tsx
export const example = stylex.create({
  test: {
    textAlign: "center",
  },
});

```

and TypeScript with `declaration` option enabled

### Additional comments

As mentioned - it seems like the whole entry point with types got removed from `package.json` - it seems like some of these types are needed for TS so it can properly build project. In my case I'm using monorepo and having shared styles in a package which is built and I need `d.ts` files for it

necolas:
The types are all in the package and the export map points to them. Have a look through the published code and let me know if you can tell why you're getting that error now - https://www.npmjs.com/package/@stylexjs/stylex/v/0.13.0?activeTab=code

pawelblaszczyk5:
> The types are all in the package and the export map points to them. Have a look through the published code and let me know if you can tell why you're getting that error now - https://www.npmjs.com/package/@stylexjs/stylex/v/0.13.0?activeTab=code

[...]

PR #1095, opened 2025-06-10
fix: export additional types to enable TypeScript declaration builds

pawelblaszczyk5:

It was discussed a bit here: https://github.com/facebook/stylex/issues/1070

For generating declarations by TypeScript all types that would be needed by the output needs to be exported so it can properly reference them in `d.ts` file. It's really important to works correctly for e.g. in monorepo scenarios where part of your code can live in a separate package. 

These two types are used by the output of `styles.create` calls - you can observe this by hovering styles created with this function call. Example:

image

(this is slightly modified fragment of code inside of repository `examples/example-nextjs/typetests/theming`)

Hence TypeScript couldn't generate the same type as in hover in `d.ts` file - because these types aren't exported from the latest release. 

Before changes (NOTE: I had to enable `declaration` in tsconfig of this example to observe this error):

image

After changes:

image

(no error 😃)

## Linked PR/Issues

Fixes https://github.com/facebook/stylex/issues/1070

## Additional Context

In theory I see that there's a `example-nextjs` with directory named `typesets` but I see them being already broken (importing from the removed `lib/StyleXTypes` entry point. And not being verified anywhere. If you'd like me to create examples for cases which this change fixes inside of repo just let me know where I can place them. Thanks a lot in advance!

## Pre-flight checklist


pawelblaszczyk5:
Hello 👋 
Rebased the branch so it's up to date with the main. Is there anything I can do to move this PR forward? 

Thanks in advance!

pawelblaszczyk5:
It seems like the CI fail is completely unrelated to the changes? 🤔 Any ideas what's happening here?

Issue #1075, opened 2025-05-22
TypeScript Error When Using CSS Variables in Props


michaltarasiuk:
### Describe the issue

I'm passing styles as props, but I ran into a TypeScript issue. When the passed styles use a CSS variable, it throws a TypeScript error (Error: Type 'Color<...>' Not Assignable to 'Readonly<string | null | undefined>' in Style Prop)

### Expected behavior

A

### Steps to reproduce

Version: 0.11.1

### Test case

```tsx
function Parent() {
  return <Child style={styles.base} />;
}

function Child({style}: {style: stylex.StyleXStyles}) {
  return <div {...stylex.props(style)} />;
}

const baseColors = stylex.defineVars({
  background: stylex.types.color('#000000'),
});

const styles = stylex.create({
  base: {
    backgroundColor: baseColors.background,
  },
});
```

### Additional comments

Extend the type 'Readonly<string | null | undefined>' to include a CSS variable

michaltarasiuk:
It can be solved this way, but I'm not sure it's the best way.

```tsx
function Parent() {
  return <Child style={styles.base} />;
}

function Child({
  style,
}: {
  style: stylex.StyleXStyles<{
    backgroundColor: Readonly<string | null | undefined> | stylex.Types.Color<string>;
  }>;
}) {
  return <div {...stylex.props(style)} />;
}

const styles = stylex.create({
  base: {
    backgroundColor: baseColors.background,
  },
});

// tokens.stylex.ts
const baseColors = stylex.defineVars({
  background: stylex.types.color('#000000'),
});
```

PR #1096, opened 2025-06-10
fix: allow stylex.types.* in StyleXStyles by resolving CSSType to underlying type

pawelblaszczyk5:

Currently at type level (at least in TypeScript, I'm not fully sure about Flow so if it impacts it in some way - I'm open for any feedback 😃) you can't use vars created by `styles.defineVars` inside of variable typed as `StyleXStyles`(which is useful for e.g. component props). 

Example (adjusted code from `examples/example-nextjs/typetests/theming`):

```ts
export const ButtonTokens = stylex.defineVars({
  bgColor: stylex.types.color('red'),
  color: 'currentcolor',
  height: 'var(--button-height-medium)',
  opacity: '1',
});
```

in other file:

```tax
  test1: {
    padding: 4,
    color: ButtonTokens.bgColor,
    backgroundColor: `color-mix(in oklch, ${ButtonTokens.bgColor}, 'white')`,
  },
});

const test: stylex.StyleXStyles = styles.test1;
```

This causes TypeScript error:

image

This happens because through some generics it mostly accepts `CSSPropertiesWithExtras` which are defined for each CSS property inside of `StyleXCSSTypes.js`. All of them accept some union/subset of string/number/null.

WIth change from this PR:

image

(no error 😃)

~I'm not 100% sure about this approach but it seems to work and I'm fully open for any feedback. I thought that maybe ideally would be listing supported `CSSType` subset inside of each property - so e.g. `<angle>` defined by `stylex.types.angle` could only be accepted by properties that in reality accepts angles. But I have two thoughts here - 1. It'd work differently than just `styles.create` where you can pass any type created to `styles.types.*` to any property. 2. It'd be a huge effort so I wanted to get some feedback before trying to do so 😅 But I'm open to implementing it this way if it makes sense?~

Changed the approach from initial one - there's a type helper that resolves type of created styles to `StyleXClassNameFor` when using `stylex.create` and there's already some logic responsible for mapping them at a type level so they're normalized. Added logic for resolving `CSSType` inside of `StyleXVar` as well

## Linked PR/Issues

Fixes https://github.com/facebook/stylex/issues/1075

## Additional Context

Similarly to this PR - https://github.com/facebook/stylex/pull/1095 I saw there's some example which in theory has dir named `typetests` but they're already borked - I'd love to add some test cases for types if guided to do so 😃 

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/stylex/src/types/StyleXCSSTypes.js
+++ b/packages/@stylexjs/stylex/src/types/StyleXCSSTypes.js
@@ -9,8 +9,6 @@
 
 'use strict';
 
-import type { CSSType } from './VarTypes';
-
 type CSSCursor =
   | 'auto'
   | 'default'
@@ -102,7 +100,7 @@ type alignSelf =
   | 'safe center'
   | 'unsafe center'
   | all;
-type all = null | 'initial' | 'inherit' | 'unset' | CSSType;
+type all = null | 'initial' | 'inherit' | 'unset';
 type animationDelay = time;
 type animationDirection = singleAnimationDirection;
 type animationDuration = time;
--- a/packages/@stylexjs/stylex/src/types/StyleXCSSTypes.js
+++ b/packages/@stylexjs/stylex/src/types/StyleXCSSTypes.js
@@ -9,6 +9,8 @@
 
 'use strict';
 
+import type { CSSType } from './VarTypes';
+
 type CSSCursor =
   | 'auto'
   | 'default'
@@ -100,7 +102,7 @@ type alignSelf =
   | 'safe center'
   | 'unsafe center'
   | all;
-type all = null | 'initial' | 'inherit' | 'unset';
+type all = null | 'initial' | 'inherit' | 'unset' | CSSType;
 type animationDelay = time;
 type animationDirection = singleAnimationDirection;
 type animationDuration = time;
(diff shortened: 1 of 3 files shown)


pawelblaszczyk5:
Hello 👋 
Rebased the branch so it's up to date with the main. Is there anything I can do to move this PR forward? 

Thanks in advance!

packages/@stylexjs/stylex/src/types/StyleXCSSTypes.js
```diff
@@ -100,7 +102,7 @@ type alignSelf =
   | 'safe center'
   | 'unsafe center'
   | all;
-type all = null | 'initial' | 'inherit' | 'unset';
+type all = null | 'initial' | 'inherit' | 'unset' | CSSType;
```
necolas: [`all` is a CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/all). So this change doesn't looks right

pawelblaszczyk5: Oh - I missed the intention here definitely. Any clues where to put it best so it’s allowed everywhere? I’ll also take a look at it tomorrow. Thanks!

I also see the test failing - missed it previously, will resolve that as well

pawelblaszczyk5: Hello 👋

Changed that approach to one that I'm a lot more confident about - there was already a type helper `ComplexStyleValueType` which when stuff like `StyleXVar` was used inside of `stylex.create` it resolved it to the underlaying value at the type level. The logic of resolving `CSSType` fits perfectly here imho. Wdyt?

necolas (approved):

PR #1099, opened 2025-06-17
[babel-plugin] Improve createTheme and defineVars tests

necolas:
Unit tests to better capture the way that theming currently works in StyleX.

Background:

* [#630](https://github.com/facebook/stylex/issues/601) describes how StyleX theming used to work, and that it wasn't matching user expectations.
* A solution was found to match expectations.
* [5f14ff58](https://github.com/facebook/stylex/commit/5f14ff580127769b7d6c315eed4e0a090230418a) is a patch that made a change to how StyleX theming works.
* I recently noticed in the React Strict DOM example app that StyleX-based theming regressed on web.

Patch:

* Make sure we capture the complete metadata in the createTheme tests. Previously, we did not have visibility into all the CSS rules generated.
* Add extra tests to capture scenarios where multiple variables and themes are created across files. These tests surface that a different "root theme" class name is created for each different file the contains `defineVars` calls.

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-createTheme-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-createTheme-test.js
@@ -44,14 +44,16 @@ function transform(source, opts = {}) {
  * StyleX exports to make sure they are transformed as expected.
  */
 
-function createFixture(fixture, { varsFilename, varsLiterals } = {}) {
-  const createKey = (key) => (varsLiterals ? `'--${key}'` : key);
+function transformWithFixture(fixture, fixtureOptions, pluginOptions) {
+  const { varsFilename, varsLiterals } = fixtureOptions || {};
+
   // Generate defineVars output first.
   // This is inlined into the fixture so that createTheme works.
-  const options = {
+  const createKey = (key) => (varsLiterals ? `'--${key}'` : key);
+  const fixtureTransformOptions = {
     filename: varsFilename || '/stylex/packages/vars.stylex.js',
   };
-  const defineVarsOutput = transform(
+  const { code: _code, metadata: _metadata } = transform(
     `
     import * as stylex from '@stylexjs/stylex';
     export const vars = stylex.defineVars({
@@ -67,13 +69,21 @@ function createFixture(fixture, { varsFilename, varsLiterals } = {}) {
       ${createKey('radius')}: 10
     });
   `,
-    options,
-  ).code;
+    fixtureTransformOptions,
+  );
 
-  return `
-    ${defineVarsOutput}
+  // Generate the final transform
+  const { code, metadata } = transform(
+    `
+    ${_code}
     ${fixture}
-  `;
+  `,
+    pluginOptions,
+  );
+
+  _metadata.stylex.push(...metadata.stylex);
+
+  return { code, metadata: _metadata };
 }
 
 const themeObject = `{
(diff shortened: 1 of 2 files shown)


packages/@stylexjs/babel-plugin/__tests__/transform-stylex-createTheme-test.js
```diff
@@ -67,13 +69,21 @@ function createFixture(fixture, { varsFilename, varsLiterals } = {}) {
...
+    ${_code}
     ${fixture}
-  `;
+  `,
+    pluginOptions,
+  );
+
+  _metadata.stylex.push(...metadata.stylex);
+
+  return { code, metadata: _metadata };
```
necolas, line 86: This test abstraction exists because we're trying to test the integration between defineVars and createTheme. But, we were only taking snapshots of partial metadata, missing the "default theme" CSS rules. To fix that, I chose to refactor this helper

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-createTheme-test.js
```diff
@@ -113,6 +122,30 @@ describe('@stylexjs/babel-plugin', () => {
...
+              0.1,
+            ],
+            [
+              "xop34xu-bdddrq",
+              {
+                "ltr": "@media print{:root, .xop34xu{--xwx8imx:white;}}",
+                "rtl": null,
+              },
+              0.1,
+            ],
```
necolas, line 148: This is an example of "default theme" CSS that now shows up in the snapshots

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-createTheme-test.js
```diff
@@ -167,6 +203,30 @@ describe('@stylexjs/babel-plugin', () => {
       expect(metadata).toMatchInlineSnapshot(`
         {
           "stylex": [
+            [
+              "xop34xu",
```
necolas, line 207: This specific hash / class name - `xop34xu` - appears in most of these snapshots because we frequently use `/stylex/packages/vars.stylex.js` as the default file name

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-createTheme-test.js
```diff
@@ -197,50 +257,76 @@ describe('@stylexjs/babel-plugin', () => {
...
-      const options = {
+      const fixtureOptions = {
         varsFilename: '/stylex/packages/src/css/vars.stylex.js',
       };
-      const fixture = createFixture(`
+      const { code, metadata } = transformWithFixture(
+        `
         export const theme = stylex.createTheme(vars, ${themeObject});
-      `);
-      const { code, metadata } = transform(fixture, options);
```
necolas, line 206: The options here were meant to be passed to the previous transform call (inside createFixture)

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-createTheme-test.js
```diff
@@ -197,50 +257,76 @@ describe('@stylexjs/babel-plugin', () => {
...
       `);
       expect(metadata).toMatchInlineSnapshot(`
         {
           "stylex": [
             [
-              "x4aw18j",
+              "x1xohuxq",
               {
-                "ltr": ".x4aw18j, .x4aw18j:root{--xwx8imx:green;--xaaua2w:antiquewhite;--xbbre8:6px;}",
+                "ltr": ":root, .x1xohuxq{--xt4ziaz:blue;--x1e3it8h:grey;--x1onrunl:10;}",
```
necolas, line 289: We fixed this test to use the vars filename correctly. Now we see that the "default theme" class name hash has changed from `xop34xu` to `x1xohuxq`.

PR #1100, opened 2025-06-17
[babel-plugin] Add themes to processStylexRules tests

necolas:


packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -45,24 +37,63 @@ function transform(source, opts = {}) {
...
+      large: '8px'
+    });
+    `,
+    {
+      filename: '/src/app/otherTokens.stylex.js',
+      parserOpts: { flow: 'all' },
+      babelrc: false,
+      plugins: [[stylexPlugin, pluginOpts]],
+    },
+  );
```
necolas, line 62: Added a second file of vars and some `createTheme` calls in the main fixture.

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -95,61 +127,96 @@ export const styles = stylex.create({
...
+        export const spacing = {
+          small: "var(--small-x19twipt)",
+          medium: "var(--medium-xypjos2)",
+          large: "var(--large-x1ec7iuc)",
+          __themeName__: "xbiwvf9"
+        };"
+      `);
+      expect(stylexPlugin.processStylexRules(metadata)).toMatchInlineSnapshot(`
+        ":root, .xsg933n{--blue-xpqh4lw:blue;}
+        :root, .xbiwvf9{--small-x19twipt:2px;--medium-xypjos2:4px;--large-x1ec7iuc:8px;}"
```
necolas, line 152: This output is the reason why theming doesn't work properly. Different `defineVars` calls in different files produce different "root theme" class names.

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -95,61 +127,96 @@ export const styles = stylex.create({
...
+          __themeName__: "xbiwvf9"
+        };
+        export const themeColor = {
+          $$css: true,
+          xsg933n: "x6xqkwy xsg933n"
+        };
+        export const themeSpacing = {
+          $$css: true,
+          xbiwvf9: "x57uvma xbiwvf9"
+        };
```
necolas, line 183: There are a few of details to point out about these objects created by StyleX:

* Each key is unique to the specific vars created by a given `defineVars` call.
    * This means merging 2 themes for the same vars will result in the last one replacing any prior ones.
* Each value is 2 class names.
    * The first is the "overrides" of the vars.
    * The second is the "root theme" key of the default vars.
* Multiple themes can be merged together on the same element (`createTheme` really creates themes partials for a given block of vars).
   * `const metaTheme = [ metaButtonTheme, metaCardTheme, metaTextTheme, ... ]`
   * Doing this today adds a lot of extra class names to the HTML element, because each `defineVars` use generates a new "root theme" class name.

StyleX currently doesn't generate a single class name that matches what `:root` does, i.e., apply all the variables that are created throughout an app by `defineVars` calls. (See my comments in #601 for why you need to apply both the *entire* default theme directly to an element, along with the overrides class, to make themes behave as expected.

What this means in practice is that themes applied lower in the DOM tree *do not* fully reset themes applied higher in the DOM tree. The DOM tree ends up looking like this:

```html
<div class="x6xqkwy xsg933n">
  <div class="x57uvma xbiwvf9"></div>
</div>
```

The inner `div` only partially resets the variables of the "root theme". It would need to also add the class `xsg933n` (and every other class name for each defineVars partial) to behave as desired.

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -95,61 +127,96 @@ export const styles = stylex.create({
...
-        html:not([dir='rtl']) .x1skrh0i:not(#\\#):not(#\\#){text-shadow:1px 2px 3px 4px red}
-        html[dir='rtl'] .x1skrh0i:not(#\\#):not(#\\#){text-shadow:-1px 2px 3px 4px red}
-        @media (min-width:320px){html:not([dir='rtl']) .x1cmij7u.x1cmij7u:not(#\\#):not(#\\#){text-shadow:10px 20px 30px 40px green}}
-        @media (min-width:320px){html[dir='rtl'] .x1cmij7u.x1cmij7u:not(#\\#):not(#\\#){text-shadow:-10px 20px 30px 40px green}}
-        @media (max-width: 1000px){.xwguixi.xwguixi:not(#\\#):not(#\\#):not(#\\#){border-color:var(--x1i1e39s)}}
-        @media (max-width: 500px){@media (max-width: 1000px){.x5i7zo.x5i7zo.x5i7zo:not(#\\#):not(#\\#):not(#\\#):not(#\\#){border-color:yellow}}}"
+        @keyframes xi07kvp-B{0%{box-shadow:1px 2px 3px 4px red;color:yellow;}100%{box-shadow:10px 20px 30px 40px green;color:var(--orange);}}
+        @keyframes xi07kvp-B{0%{box-shadow:-1px 2px 3px 4px red;color:yellow;}100%{box-shadow:-10px 20px 30px 40px green;color:var(--orange);}}
+        :root, .xsg933n{--blue-xpqh4lw:blue;}
+        :root, .xbiwvf9{--small-x19twipt:2px;--medium-xypjos2:4px;--large-x1ec7iuc:8px;}
```
necolas, line 206: The fix is to make sure the `:root` selector's companion class names are all the same. This will also reduce the number of theming-related class names applied to the element.

mellyeliu (approved):

vincentriemer (approved):

PR #1101, opened 2025-06-17
[babel-plugin] Fix theming in dev/debug mode

necolas:
This fixes a bug in StyleX theming caused by how evaluate-path works in dev/debug mode, which produced bad debug class names for `createTheme` calls.

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-import-export-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-import-export-test.js
@@ -251,8 +251,8 @@ describe('@stylexjs/babel-plugin', () => {
           }
         };
         const theme = {
-          $$css: true,
-          xop34xu: "xfnndu4 xop34xu"
+          xop34xu: "xfnndu4 xop34xu",
+          $$css: true
         };
         stylex.props(styles.root, theme);"
       `);
@@ -386,8 +386,8 @@ describe('@stylexjs/babel-plugin', () => {
           }
         };
         const theme = {
-          $$css: true,
-          xop34xu: "xfnndu4 xop34xu"
+          xop34xu: "xfnndu4 xop34xu",
+          $$css: true
         };
         foo.props(styles.root, theme);"
       `);
@@ -436,8 +436,8 @@ describe('@stylexjs/babel-plugin', () => {
           }
         };
         const theme = {
-          $$css: true,
-          xop34xu: "xfnndu4 xop34xu"
+          xop34xu: "xfnndu4 xop34xu",
+          $$css: true
         };
         props(styles.root, theme);"
       `);
@@ -495,8 +495,8 @@ describe('@stylexjs/babel-plugin', () => {
           }
         };
         const theme = {
-          $$css: true,
-          xop34xu: "xfnndu4 xop34xu"
+          xop34xu: "xfnndu4 xop34xu",
+          $$css: true
         };
         _props(styles.root, theme);"
       `);
@@ -536,8 +536,8 @@ describe('@stylexjs/babel-plugin', () => {
           }
         };
         const theme = {
-          $$css: true,
-          xop34xu: "xfnndu4 xop34xu"
+          xop34xu: "xfnndu4 xop34xu",
+          $$css: true
         };
         stylex.props(styles.root, theme);"
       `);
@@ -590,8 +590,8 @@ describe('@stylexjs/babel-plugin', () => {
           }
         };
         const theme = {
-          $$css: true,
-          xop34xu: "xfnndu4 xop34xu"
+          xop34xu: "xfnndu4 xop34xu",
+          $$css: true
         };
         css.props(styles.root, theme);"
       `);
@@ -631,8 +631,8 @@ describe('@stylexjs/babel-plugin', () => {
           }
         };
         const theme = {
-          $$css: true,
-          xop34xu: "xfnndu4 xop34xu"
+          xop34xu: "xfnndu4 xop34xu",
+          $$css: true
         };
         stylex.props(styles.root, theme);"
       `);
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -174,12 +174,12 @@ describe('@stylexjs/babel-plugin', () => {
           __themeName__: "xbiwvf9"
         };
         export const themeColor = {
-          $$css: true,
-          xsg933n: "x6xqkwy xsg933n"
+          xsg933n: "x6xqkwy xsg933n",
+          $$css: true
         };
         export const themeSpacing = {
-          $$css: true,
-          xbiwvf9: "x57uvma xbiwvf9"
+          xbiwvf9: "x57uvma xbiwvf9",
+          $$css: true
         };
         export const styles = {
           root: {
@@ -243,12 +243,12 @@ describe('@stylexjs/babel-plugin', () => {
           __themeName__: "xbiwvf9"
         };
         export const themeColor = {
-          $$css: true,
-          xsg933n: "x6xqkwy xsg933n"
+          xsg933n: "x6xqkwy xsg933n",
+          $$css: true
         };
         export const themeSpacing = {
-          $$css: true,
-          xbiwvf9: "x57uvma xbiwvf9"
+          xbiwvf9: "x57uvma xbiwvf9",
+          $$css: true
         };
         export const styles = {
           root: {
--- a/packages/@stylexjs/babel-plugin/src/shared/stylex-create-theme.js
+++ b/packages/@stylexjs/babel-plugin/src/shared/stylex-create-theme.js
@@ -28,7 +28,7 @@ export default function styleXCreateTheme(
 ): [{ $$css: true, +[string]: string }, { [string]: InjectableStyle }] {
   if (typeof themeVars.__themeName__ !== 'string') {
     throw new Error(
-      'Can only override variables theme created with stylex.defineVars().',
+      'Can only override variables theme created with defineVars().',
     );
   }
 
@@ -85,8 +85,10 @@ export default function styleXCreateTheme(
   const themeClass = `${overrideClassName} ${themeVars.__themeName__}`;
 
   return [
-    // $FlowFixMe[invalid-computed-prop]
-    { $$css: true, [themeVars.__themeName__]: themeClass },
+    {
+      [themeVars.__themeName__]: themeClass,
+      $$css: true,
+    },
     stylesToInject,
   ];
 }
--- a/packages/@stylexjs/babel-plugin/src/shared/stylex-vars-utils.js
+++ b/packages/@stylexjs/babel-plugin/src/shared/stylex-vars-utils.js
@@ -37,7 +37,7 @@ export function collectVarsByAtRule(
     return;
   }
   if (Array.isArray(value)) {
-    throw new Error('Array is not supported in stylex.defineVars');
+    throw new Error('Array is not supported in defineVars');
   }
   if (typeof value === 'object') {
     if (value.default === undefined) {
@@ -76,7 +76,7 @@ export function getDefaultValue(value: VarsConfigValue): ?string {
     return null;
   }
   if (Array.isArray(value)) {
-    throw new Error('Array is not supported in stylex.defineVars');
+    throw new Error('Array is not supported in defineVars');
   }
   if (typeof value === 'object') {
     if (value.default === undefined) {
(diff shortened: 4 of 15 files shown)


packages/@stylexjs/babel-plugin/src/shared/stylex-vars-utils.js
```diff
@@ -19,6 +19,8 @@ export type VarsConfig = $ReadOnly<{
 
 const SPLIT_TOKEN = '__$$__';
 
+export const ROOT_VARS_CLASSNAME = 'stylexvars';
```
necolas: I hardcoded the class name because it felt difficult to understand the snapshots when everything was a hashed string, and this key needs to exist independently from the hashed strings used to identify the product of `defineVars` calls.

mellyeliu:
Just to clarify, the intention is to create a global variable pool, have each theme reset all variables, and move towards a "composite theme" model? 

In that case, we should probably change the function signature of `createTheme`, since the `vars` param no longer scopes anything. We might want to also double check with DSP to see if this affects how they plan on applying the themes.

necolas:
That's not really correct. This is a fix for a broken implementation of how theming was meant to work in 0.9.3. It doesn't change the fact that there has always been a pool of global variables. And it doesn't change the fact that defineVars still scopes the variables created by createTheme to the specific input vars. You can look at the changes (and non-changes) to the snapshots to validate this

packages/benchmarks/size/rollup.config.mjs
```diff
@@ -22,7 +22,7 @@ const config = {
   },
   // See all options in the babel plugin configuration docs:
   // https://stylexjs.com/docs/api/configuration/babel-plugin/
-  plugins: [stylexPlugin({ fileName: 'stylex.css' })],
+  plugins: [stylexPlugin({ fileName: 'stylex.css', useCSSLayers: true })],
```
necolas: I will revert this so it doesn't change the perf workflow

mellyeliu:
I think this PR does change how theming works downstream. When you add a `createTheme` call to var group, it resets all other variables to their default, even those from different groups.

Say we have two variable groups:

```
export const colorTokens = stylex.defineVars({
  red: 'red',
  blue: 'blue',
});

export const sizeTokens = stylex.defineVars({
  small: '16px',
  big: '24px',
});
```

And themes:
```
const darkColors = stylex.createTheme(colorTokens, {
  red: 'darkred',
});

const largeText = stylex.createTheme(sizeTokens, {
  big: '36px',
});
```

With this implementation, adding the `largeText` theme resets all `colorTokens` to the default, even though `largeText` is meant to only apply to`sizeTokens`, which is a little unintuitive. Thus all previously applied themes are ignored within the block. 
```
  text: {
    fontSize: fontSizes.big,
    color: colors.red,
  },
});

<div {...stylex.props(darkColors, styles.text)}>
    This is dark red
    <div {...stylex.props(largeText, styles.text}>
       This is red again and 36px (previously dark red and 36px)
    </div>
</div>
```

I think this is because each theme emits the new `.stylexvars` class every time, which fixes the theming issue in #601, but also has this side effect. But would need to dig in a bit more to verify.

necolas:
> When you add a createTheme call to var group, it resets all other variables to their default, even those from different groups.

Yes, that's essential to have themes work as expected per #601. What was landed in StyleX for 0.9.3 is not useful in practice. No one uses themes like this; it's what variants are for:

```
  <div {...stylex.props(darkColors, style.hero)}>
    <div {...stylex.props(style.h1)}>
      This is dark red
      <div {...stylex.props(largeText, style.h1)}>
         This is red again and 36px (previously dark red and 36px)
      </div>
    </div>
  </div>
```

Instead, the expectation is that themes are a combination of colors, sizes, font styles, etc., that get applied to an ancestor to style the subtree of elements that are themselves referencing vars (not themes) as values. This is how you create and retheme a themeable tree of components. It's the same mental model as React Context.

> I think this is because each theme emits the new .stylexvars class every time

`defineVars` calls emit `.stylexvars`, not `createTheme`. It's might be helpful to think of `defineVars` as "the default theme" or "add some global variables". That's why it used to be that all `defineVars` calls only emitted vars under the `:root` selector - they've always been one big bag of global variables. 

An earlier name for `createTheme` was `overrideVars` - all it does it create some new rules that set new values for the previously-created custom properties.

```css
/* this is what defineVars generates */
:root, .stylexvars {--blue-xpqh4lw:blue;}
/* this is what createTheme generates for the above vars */
.x6xqkwy:root, .x6xqkwy {--blue-xpqh4lw:lightblue;}
```

But to fix the problems in #601, we need a way to reference *all* these variables - the "default theme" - on subtrees that are trying to set a non-default theme.

I don't know why the patch for 0.9.3 was engineered to make each `defineVars` emit a different companion class to the `:root` selector, but it meant there was still no way to get a handle to all the vars defined by an app. The StyleX theming APIs are really flexible, so you can use one `defineVars` in your app and create `n` mutually exclusive themes using one `createTheme` call each. The key name for each generated theme is the same, since the input vars are the same. Or, you can use many `defineVars` in independent components, and then use multiple `createTheme` in your app to define overrides using your design system tokens. But you must then create a "composite theme" by stitching together the overrides for each different bag of `vars` (i.e., the array of override styles that I've shown before). This way of theming *also* works, because the keys on the generated style objects are all different since each of the input vars objects are different.

The way Meta does themes internally right now is to add pretty much all vars using a single `defineVars` call, and then create a handful of themes using those vars. Meta probably hasn't encounter the scenario where you set a different theme on a subtree within a top-level theme, and expect it to *render the same as it does when it's the top-level theme*.

Imagine this scenario. We have a themeable "ad" component where styles reference "design system" tokens. In Facebook, the root of the app is wrapped in the "Facebook theme" so all those ads looks like the rest of Facebook. On Instagram, we use the same ad component but since Instagram is wrapped in "Instagram theme", they look like the rest of Instagram. This scenario works for both the pre-0.9.3 theming, the post-0.9.3 theming, and this PR. But if the ads team wants to show a preview of the "ad" component within Ads Manager, and they want to display that ad as it will look on Facebook OR Instagram, now the bug is encountered. Ads Manager might be wrapped in "Meta theme", so by default all the ads would take on the that theme, which is not what they want. So they wrap the ad preview container in "Facebook theme", expecting it to theme ad component to match Facebook. But it won't work. The current way that theming works in StyleX for this scenario is completely broken - themes simple don't work the same if they are used on a subtree within another theme. This patch fixes that.

RSD's implementation of stylex is another way to help explain how things should work and why the current system is broken when using themes-within-themes in the component tree.

* [defineVars](https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/src/native/stylex/index.js#L463-L475) adds all vars to a global registry, like we do in CSS for StyleX.
* [createTheme](https://github.com/facebook/react-strict-dom/blob/main/packages/react-strict-dom/src/native/stylex/index.js#L439-L450) simply generates a new object with the same keys, but new values, like we do in CSS for StyleX.
* And the rest is [using React context](https://github.com/facebook/react-strict-dom/blob/d745d1d11f6d7f3775f521e8b9dbb560637772bc/packages/react-strict-dom/src/native/modules/useNativeProps.js#L19) to merge default vars with the theme overrides.

mellyeliu:
That all makes sense, and I think this approach makes sense to resolve the subtree theming issues, and more in line with how themes are expected to work in practice.

I do think it’d be helpful to update the docs and examples around theming. The [docs](https://stylexjs.com/docs/learn/theming/creating-themes/) have this idea of variable groups, which gives the user the impression that `defineVars` creates an isolated group of variables that function independently. I’m guessing that was the model behind @nmn’s intention in 0.9.3. but that framing doesn’t reflect internal usage patterns you pointed out (which tend toward a unified variable set). Either way, the current notes give a misleading mental model for consumers.

We can also clarify that the `vars` param in createTheme no longer scopes which variables are affected, it just determines which ones are overridden, with everything else is implicitly reset to default via the base `.stylexvars` class

packages/@stylexjs/babel-plugin/src/shared/stylex-vars-utils.js
```diff
@@ -76,13 +78,13 @@ export function getDefaultValue(value: VarsConfigValue): ?string {
     return null;
   }
   if (Array.isArray(value)) {
-    throw new Error('Array is not supported in stylex.defineVars');
+    throw new Error('Array is not supported in defineVars');
```
mellyeliu, line 79: can do later but these should prob be moved into the `messages.js` file

mellyeliu (approved):

nmn:
I'm reviewing this now. Please don't merge because I think we went through a whole discussion about this before.

packages/@stylexjs/babel-plugin/__tests__/transform-import-export-test.js
```diff
@@ -251,8 +251,9 @@ describe('@stylexjs/babel-plugin', () => {
           }
         };
         const theme = {
-          $$css: true,
-          xop34xu: "xfnndu4 xop34xu"
+          xop34xu: "xfnndu4",
+          __rootVars__: "stylexvars",
+          $$css: true
```
nmn: This looks exactly like what we discussed a while ago. Every theme sets the variables in that theme definition *and* sets all other variables back to their default values.

This would mean that if you have two `VarGroup`s: `colors` and `density`.

And on one element you have applied a `dense` theme, applying `dark` theme on a child element will reset that back to the default density values.

This is not how anyone expects theming to work and would be a major breaking change.

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -148,8 +148,8 @@ describe('@stylexjs/babel-plugin', () => {
         };"
       `);
       expect(stylexPlugin.processStylexRules(metadata)).toMatchInlineSnapshot(`
-        ":root, .xsg933n{--blue-xpqh4lw:blue;}
-        :root, .xbiwvf9{--small-x19twipt:2px;--medium-xypjos2:4px;--large-x1ec7iuc:8px;}"
+        ":root, .stylexvars{--blue-xpqh4lw:blue;}
+        :root, .stylexvars{--small-x19twipt:2px;--medium-xypjos2:4px;--large-x1ec7iuc:8px;}"
```
nmn: Again, the previous implementation was what we landed on after a lot of discussion as a theme would only reset variable in its *own* `VarGroup` back to their default values.

This new implementation will reset everything. Even variables in other `VarGroup`s. This breaks the entire purpose of having separate groups of variables and themes for those groups.

packages/@stylexjs/babel-plugin/__tests__/transform-import-export-test.js
nmn: When I originally implemented the PR that resets the VarGroup to its default values, I had originally implemented behind a config and this was one of the ways it could work. I then had a long discussion with you and we settled on how it works today.

I know of another solution that would fix the "reactive variables" expectation without causing this regression. We didn't do that until now, because variables are not reactive in CSS and we agreed it was OK to model the CSS behaviour. However, if this issue keeps coming up, I'll implement a solution that will solve it without these regressions.

packages/@stylexjs/babel-plugin/src/shared/stylex-create-theme.js
```diff
@@ -85,8 +85,10 @@ export default function styleXCreateTheme(
   const themeClass = `${overrideClassName} ${themeVars.__themeName__}`;
 
   return [
-    // $FlowFixMe[invalid-computed-prop]
-    { $$css: true, [themeVars.__themeName__]: themeClass },
+    {
+      [themeVars.__themeName__]: themeClass,
+      $$css: true,
+    },
```
necolas, line 91: This change avoids the need for the flowfixme  comment

packages/@stylexjs/babel-plugin/src/utils/evaluate-path.js
```diff
@@ -175,14 +175,17 @@ function evaluateThemeRef(
...
-    const varSafeKey = (
-      key[0] >= '0' && key[0] <= '9' ? `_${key}` : key
-    ).replace(/[^a-zA-Z0-9]/g, '_');
+    const varSafeKey =
+      key === '__themeName__'
+        ? ''
+        : (key[0] >= '0' && key[0] <= '9' ? `_${key}` : key).replace(
+            /[^a-zA-Z0-9]/g,
+            '_',
+          ) + '-';
```
necolas, line 184: This is a fix for dev/debug mode. The way theming is currently designed to work is broken in dev because this generates bad class names that don't map to anything in the generated CSS

mellyeliu:
Dev/debug changes lgtm

PR #1104, opened 2025-06-23
[babel-plugin] Fix theming in dev/debug mode

necolas:
This fixes a bug in StyleX theming. Before, defineVars was creating a `:root` rule with a classname that is unique to a given defineVars call. However, when debug mode is enabled, calls to createTheme were generating bad debug class names and the unique classname for vars was not being applied. This bug isn't caught by the unit tests because they are using fixtures and not using the bad logic in evaluate-path.js


necolas:
Duplicates #1101

PR #1111, opened 2025-07-01
fix: export additional type for dynamic styles in TS declaration build

pawelblaszczyk5:

That's analogous to this one: https://github.com/facebook/stylex/pull/1095

User reported that one more thing is still erroring in 0.14 - dynamic styles. Must've missed them because I don't use them too much 😅 In theory I feel like TS should work fine with `InlineStyles` being exported but if I understand correctly - type aliases are often resolved to their underlaying value - same happens with e.g. hover tips.

Hence, I added one more export, before (you can see what TS would like to generate):

image

After change from MR:

image

(no error - it works)

## Linked PR/Issues

Fixes https://github.com/facebook/stylex/issues/1070#issuecomment-3021954185

This issue is closed from my previous PR - as mentioned, I've missed one use case 

## Additional Context

Same as previous MR - I feel like it'd be useful to do some typetesting (also in CI) that would enforce nothing breaks between releases etc. I'd be really eager to contribute this if given some guidance how best fit it into this repository :)

## Pre-flight checklist


pawelblaszczyk5:
I’ll need to take a deeper look because Flow structure is a bit different there than TypeScript one

Edit: this approach seems to work - open for feedback obviously 😃

packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
```diff
@@ -193,7 +193,9 @@ export type CompiledStyles =
 
 declare const StyleXInlineStylesTag: unique symbol;
 
-export type InlineStyles = typeof StyleXInlineStylesTag;
+export type InlineStyles = string & {
```
necolas: Why is it `string &`?

pawelblaszczyk5: Maybe it’s unneeded - my thought was to make it similar to other opaque types here (it’s basically a token type not meaning anything anyway). I’ll verify whether it’s needed later today

necolas:
> I feel like it'd be useful to do some typetesting (also in CI) that would enforce nothing breaks between releases etc. I'd be really eager to contribute this if given some guidance how best fit it into this repository :)

Yes that would be useful. Try putting some alongside the flow tests in the `@stylexjs/stylex` package

pawelblaszczyk5:
> Yes that would be useful. Try putting some alongside the flow tests in the `@stylexjs/stylex` package

What do you think about separate PR with this? Or would you prefer this being a part of this as well?

necolas:
A separate PR is fine. Best to get this fix into a patch release ASAP I think. Thanks!

pawelblaszczyk5:
Thanks - I’ll take a look in a few hours in the opaque type, when at home. And will make a follow-up with TS tests in following days

packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
pawelblaszczyk5: Yeah, `string &` isn't required here - so adjusted that if the similarity to other opaque types isn't concern here. Works correctly in my testing and intellisense/hover tip is also still nice without unnecessary expansion:

image

necolas:
Are you happy that the latest iteration of the type does the trick?

pawelblaszczyk5:
Yup :)

necolas (approved):
Thanks for the help improving the experience for typescript!

pawelblaszczyk5:
My pleasure! @necolas Also one more question about the mentioned typescript tests stuff. I'm not sure whether putting them next to `flow-tests` makes sense now. Some of TS types are currently codegen so it's not as easy as just importing stuff from `src`. Wdyt about putting them more similarly to e.g. `benchmark` package - separate one, which depends on the `@stylexjs/stylex` stuff? This way the tests would be closer to normal usage, especially with codegen and stuff like some types not being exported as a public API.

necolas:
ok sure

PR #1115, opened 2025-07-02
[babel-plugin] add message template for filename error

mellyeliu:


packages/@stylexjs/babel-plugin/src/shared/messages.js
```diff
@@ -22,6 +22,8 @@ export const nonExportNamedDeclaration = (fn: string): string =>
   `The return value of ${fn}() must be bound to a named export.`;
 export const unboundCallValue = (fn: string): string =>
   `${fn}() calls must be bound to a bare variable.`;
+export const missingFilename = (fn: string): string =>
+  `No filename found for generating ${fn}() key name.`;
```
necolas: Maybe we can take this opportunity to provide a better error message? Not sure what exactly, but this message is always a bit confusing when it pops up while writing unit tests (hopefully people aren't encountering it in other scenarios).

necolas (approved):

packages/@stylexjs/babel-plugin/src/shared/messages.js
mellyeliu: Good callout. I added some context based on the checks in `fileNameForHashing` to (hopefully) be more descriptive to users.

necolas: lgtm

Issue #1091, opened 2025-06-03
[babel-plugin] Some shorthand expansions broken with `legacy-expand-shorthands`


mellyeliu:
### Describe the issue

A lot of the shorthand properties in `legacy-expand-shorthands` are actually not expanded correctly. Properties like `margin`, `padding`, and `borderColor` seem to be fine, but many other properties like `paddingInline`, `marginInline`, and other values are never expanded. Technically we discourage use of multivalue shorthands anyway, but the rule is only turned on in xplat-react folders.

This leads to properties like `paddingInline: "5px 10px"` being expanded to `paddingLeft: "5px 10px"` and `paddingRight: "5px 10px"`. Not ideal, whatever the circumstance.

### Expected behavior

`paddingInline: "5px 10px"` should be `paddingLeft: "5px"` and `paddingRight: "10px"` in ltr and flipped in rtl

### Steps to reproduce

See generated CSS when passing `paddingInline: "5px 10px"` with `styleResolution: legacy-expand-shorthands`

PR #1118, opened 2025-07-03
[babel-plugin][legacy] fix expansion for marginInline, paddingInline, etc

mellyeliu:
This is a longstanding bug for `marginInline`, `marginBlock`, `paddingInline`, `paddingBlock` (https://github.com/facebook/stylex/issues/1091). I'm guessing it's an oversight because the rest of the shorthands (`margin`, `padding`, `gap`, etc.) are handled with `splitValue`. 

Currently, we don't split or handle the values passed to any of these properties. The `valid-shorthands` linter discourages them anyway, but internally it's not enabled everywhere and not strictly enforced.

```
paddingBlock: "5px 10px" ->

_inject2(".x4tyy86{padding-top:5px 10px}", 4000);
_inject2(".xr7q0qe{padding-bottom:5px 10px}", 4000);
```

This PR handles these values appropriately:

```
paddingBlock: "5px 10px" ->

_inject2(".x4tyy86{padding-top:5px}", 4000);
_inject2(".xr7q0qe{padding-bottom:10px}", 4000);
```

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/legacy/stylex-transform-legacy-shorthands-test.js
@@ -236,6 +236,36 @@ describe('legacy-shorthand-expansion style resolution (enableLogicalStylesPolyfi
       `);
     });
 
+    test('paddingInline: basic multivalue shorthand', () => {
+      expect(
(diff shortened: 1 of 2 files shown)


necolas:
This is just for the legacy strategy right? We should start making that explicit in commit messages so we remember it's just for meta migration issues

necolas (approved):

PR #1123, opened 2025-07-03
Update npm-release documentation

necolas:
We can't do everything in one command because we can't push commits directly to `main`. And the GitHub PR UI creates a new commit on merge, which means tags have to be added to the post-merge commit on `main`.

### DIFF
--- a/tools/npm/release.js
+++ b/tools/npm/release.js
@@ -109,7 +109,7 @@ if (commit) {
   // commit
   execSync(`git commit -m "${pkgVersion}" --no-verify`, { stdio: 'inherit' });
   // tag
-  execSync(`git tag -fam ${pkgVersion} "${pkgVersion}"`, { stdio: 'inherit' });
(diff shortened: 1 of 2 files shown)


tools/npm/release.js
```diff
@@ -109,7 +109,7 @@ if (commit) {
   // commit
   execSync(`git commit -m "${pkgVersion}" --no-verify`, { stdio: 'inherit' });
   // tag
-  execSync(`git tag -fam ${pkgVersion} "${pkgVersion}"`, { stdio: 'inherit' });
+  // execSync(`git tag -fam ${pkgVersion} "${pkgVersion}"`, { stdio: 'inherit' });
```
mellyeliu, line 112: are we keeping this for any reason?

necolas: Maaybe we'll want to relax blocking push-to-main for code owners in the future, if just for the release process now that it can be automated and shouldn't really need reviews of the commit. Then we can go back to just one command that does it all

PR #1124, opened 2025-07-03
test: add TypeScript tests

pawelblaszczyk5:

As discussed in a few PRs related to TypeScript lately - I've made some setup for testing TypeScript types and added some initial test cases. I've added it as a separate package (as discussed here https://github.com/facebook/stylex/pull/1111) so it can properly work with codegen types and validate whether it behaves properly when it doesn't have access to some internal types.

Initial test cases:

- `exports.ts` - verifies whether everything works with TS declaration enabled, so that everything can be properly exported without errors, protects from issues such as https://github.com/facebook/stylex/issues/1070 and https://github.com/facebook/stylex/issues/857
- `styles.styles.ts` - verifies whether styles can be created and later used with stuff like `stylex.StyleXStyles` or `stylex.StaticStyles`. Protects from such edge cases such as https://github.com/facebook/stylex/issues/1075 It also catched one error which could be maybe resolved in the future, but I'm not yet sure how to approach it
- `theming.ts` - verifies whether themese/vars/consts can be properly created, extended, typed. Also verified the functionality of unique type identity themes/vars

I also looked through flow tests and the existing TypeScript tests which are in the next.js example (should we remove them? They don't do anything anyway and make next example harder for users?), but weren't properly run anywhere and missed a lot of edge cases - I think I covered all the same cases. Happy to add some more if needed.

I added a command so TypeScript can be easily run and added it to existing `test` script. Also added proper step to the CI, analogous to the Flow step.

Two open questions:

1. Spotted that every file has the Meta licensing header, should I add it to these files as well?
2. Should I add a readme with some explanation, e.g. similar to the above so it's not lost in PR description only? I saw that some packages have readme and some don't

## Linked PR/Issues

N/A

## Additional Context

Currently everything passes:

image

If something fails we'll get an error

image

## Pre-flight checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,10 @@
     "prepare": "husky install tools/husky",
     "build": "npm run build --workspaces --if-present",
     "flow": "flow",
-    "prettier": "prettier --write \"**/*.js\" \"**/*.ts\"",
-    "prettier:report": "prettier --check \"**/*.js\"",
+    "prettier": "prettier --write \"**/*.js\" \"**/*.flow\"",
+    "prettier:report": "prettier --check \"**/*.js\" \"**/*.flow\"",
     "lint": "npm run lint:report -- --fix",
-    "lint:report": "eslint . --ext .js,.jsx,.mjs",
+    "lint:report": "eslint . --ext .flow,.js,.jsx,.mjs",
     "release": "tools/npm/release.js",
     "test:packages": "npm run test --workspaces --if-present",
     "test": "npm run build && npm run flow && npm run test:packages && npm run prettier:report && npm run lint:report"
@@ -56,6 +56,7 @@
         "files": [
           "*.js",
           "*.jsx",
+          "*.mjs",
           "*.flow"
         ],
         "options": {
--- a/packages/typescript-tests/package.json
+++ b/packages/typescript-tests/package.json
@@ -1,0 +1,17 @@
+{
+  "private": true,
+  "name": "typescript-tests",
+  "version": "0.14.1",
+  "type": "module",
+  "scripts": {
+    "test": "tsc"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@stylexjs/stylex": "0.14.1",
+    "typescript": "^5.8.3"
+  },
+  "engines": {
+    "node": ">=20.11.0"
+  }
+}
--- a/packages/typescript-tests/tsconfig.json
+++ b/packages/typescript-tests/tsconfig.json
@@ -1,0 +1,21 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "compilerOptions": {
+    "skipLibCheck": true,
+    "moduleDetection": "force",
+    "isolatedModules": true,
+    "incremental": true,
+    "module": "NodeNext",
+    "target": "ESNext",
+    "outDir": "${configDir}/node_modules/.cache/tsbuildinfo/",
+    "noEmit": true,
+    "declaration": true,
+
+    "strict": true,
+    "exactOptionalPropertyTypes": true,
+    "noUncheckedIndexedAccess": true,
+
+    "noErrorTruncation": true
+  },
+  "include": ["src"]
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,6 +36,7 @@ module.exports = {
     'dist',
     'flow-typed',
     'lib',
+    'flow_modules',
     'node_modules',
     'next-env.d.ts',
     'next.config.js',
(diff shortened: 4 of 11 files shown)


packages/typescript-tests/src/stylex-styles.ts
```diff
@@ -0,0 +1,277 @@
...
+  styles.combinedConditions,
+] satisfies stylex.StyleXStylesWithout<{ height: unknown }>;
+
+[
+  styles.basic,
+  // @ts-expect-error - should disallow dynamic styles
+  styles.dynamicWithStringParam('100%'),
+  // @ts-expect-error - should disallow dynamic styles
+  styles.dynamicWithNumericParam(100),
+  // styles.empty - adding this breaks TypeScript,
```
pawelblaszczyk5: The one mentioned case which I spotted breaks TS here 😄 `styles.empty` for some reason overwrites the type of the whole array to `Array<Readonly<{}>>` which in TS means any non-nullish value so it passes. Not sure why it overwrites it only in that case and not in these above. I'm also not yet sure how it could be solved so maybe I'll try to tackle it in the future 😉

packages/typescript-tests/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "compilerOptions": {
+    "skipLibCheck": true,
+    "moduleDetection": "force",
+    "isolatedModules": true,
+    "incremental": true,
+    "module": "NodeNext",
+    "target": "ESNext",
+    "outDir": "${configDir}/node_modules/.cache/tsbuildinfo/",
```
necolas, line 10: Is it standard in ts-world to set the outdir to a path within `node_modules`?

packages/typescript-tests/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "private": true,
+  "name": "typescript-tests",
+  "version": "0.14.1",
+  "type": "module",
+  "scripts": {
+    "typescript": "tsc"
```
necolas: This script could be called `test` and then it will automatically run without needing changes to the root scripts or github workflow

package.json
```diff
@@ -8,13 +8,14 @@
     "prepare": "husky install tools/husky",
     "build": "npm run build --workspaces --if-present",
     "flow": "flow",
+    "typescript": "npm run typescript --workspaces --if-present",
```
necolas: This is really a test, and not a monorepo task. You should be able to revert the changes to this file and change the script name in the new package

.github/workflows/tests.yml
```diff
@@ -18,6 +18,16 @@ jobs:
...
+  typescript:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22.x'
+      - run: npm install
+      - run: npm run typescript
+
```
necolas: Revert this change too

packages/typescript-tests/src/stylex-styles.ts
```diff
@@ -0,0 +1,277 @@
+import * as stylex from '@stylexjs/stylex';
```
necolas: Let's call this file `styles.ts`

packages/typescript-tests/src/theming.ts
```diff
@@ -0,0 +1,148 @@
+import * as stylex from '@stylexjs/stylex';
```
necolas: Let's call this file `themes.ts`

necolas (commented):
Thanks. This looks pretty good!

(We should probably do something like this for the flow-tests too. At the moment they don't test the public API properly, and they look like mess.)

packages/typescript-tests/tsconfig.json
pawelblaszczyk5: It's not an output you'd expect here. We disabled emitting so only thing TypeScript outputs is a `tsbuildinfo` file which is used for faster subsequent type checking. You don't want to put it in source control, and a `node_modules/.cache` dir is an easy solution for this, and it's also used by many other tools. And it's not my own practice, it has some relevance in TS community imho - especially for folks coming from monorepo background. Here's example resource from a folk which recommends such setup - https://www.shew.dev/monorepos/guardrails/typescript?utm_source=chatgpt.com#set-up-a-typescript-workspace

packages/typescript-tests/package.json
pawelblaszczyk5: Good idea - though about this as well when implementing, so sounds reasonable to do it this way

pawelblaszczyk5:
Resolved all of the stuff 😃 Also re flow tests - I'm not as familiar with flow but it would be a cool challenge to learn a bit and try to create similar tests for flow if you'd be open for this and given some small tips on how to approach this

PR #1136, opened 2025-07-11
[eslint-plugin] add style validation for length properties

mellyeliu:
Let's add basic lint validation for length properties - numeric string literal patterns like `margin: "10"` which are safely incorrect. Starting with the linter over the compiler so we don't break anything 

- patterns like `margin: "10"` which are syntactically valid but semantically incorrect (missing unit)
- applied to all length-related properties (margin, padding, width, height, etc.)

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -1857,5 +1857,89 @@ revert`,
         },
       ],
     },
+    {
+      code: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          invalidStyle: {
+            margin: "10",
+            height: "10",
+          },
+        });
+      `,
+      errors: [
+        {
+          message:
+            'margin value must be one of:\n' +
+            'a number literal or math expression\n' +
+            'a non-numeric string\n' +
+            'null\n' +
+            'initial\n' +
+            'inherit\n' +
+            'unset\n' +
+            'revert',
+        },
+        {
+          message:
+            'height value must be one of:\n' +
+            'a non-numeric string\n' +
+            'a number literal or math expression\n' +
+            'available\n' +
+            'min-content\n' +
+            'max-content\n' +
+            'fit-content\n' +
+            'auto\n' +
+            'a number ending in px, mm, in, pc, pt\n' +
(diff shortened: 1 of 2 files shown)


packages/@stylexjs/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -140,6 +140,17 @@ const isRelativeLength: RuleCheck = (
 
 const isLength = makeUnionRule(isAbsoluteLength, isRelativeLength);
 
+const isNonNumericString: RuleCheck = (node: Node): RuleResponse => {
+  if (node.type === 'Literal' && typeof node.value === 'string') {
+    if (/^[-+]?\d+(\.\d+)?$/.test(node.value)) {
```
nmn: This is very accurate, but I think anything that is `23.34.2334` is also invalid, so you could look for `[\d\.]` if you wanted.

Also, I think you might want to change the `+` after the first `\d` to a `*` as `.50` is a numeric string.

nmn (approved):
lgtm

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -1857,5 +1857,89 @@ revert`,
...
+            height: "10",
+          },
+        });
+      `,
+      errors: [
+        {
+          message:
+            'margin value must be one of:\n' +
+            'a number literal or math expression\n' +
+            'a non-numeric string\n' +
```
necolas, line 1875: These messages could align more with the formal CSS syntax for a given property, e.g., "a CSS <length>, <length-percentage>" etc. https://developer.mozilla.org/en-US/docs/Web/CSS/width#formal_syntax

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -1857,5 +1857,89 @@ revert`,
...
+          message:
+            'height value must be one of:\n' +
+            'a non-numeric string\n' +
+            'a number literal or math expression\n' +
+            'available\n' +
+            'min-content\n' +
+            'max-content\n' +
+            'fit-content\n' +
+            'auto\n' +
+            'a number ending in px, mm, in, pc, pt\n' +
```
necolas, line 1892: Curious why this message is present for height but not margin? Aligning on official syntax might help with consistency

PR #1137, opened 2025-07-14
[feat] Hoist stylex.create definitions and nested objects within functions

nmn:

Today all StyleX transformation happens *in-place*. This usually works fine, but there are edge cases where this breaks down a bit.
- Using `stylex.create` within Typescript namespaces doesn't work
- Dynamic style function recreate the object for static classNames on every call, de-opting caching.

To solve these issues, this PR refactors and add AST utilities for hoisting expressions to the top level of a file and provide an identifier reference to it instead.

As of right now this is only used for two use-cases:
1. `stylex.create` can now be defined within blocks such as functions. The compiler will define the actual object at the top level of the file, and use a stable reference where it is defined.
2. Dynamic styles are tuples of static classNames objects and dynamic inline styles. The static classNames are now hoisted out so they can maintain a static reference.

Tests have been added and updated to show these changes.

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
@@ -368,13 +368,14 @@ describe('Evaluation of imported values works based on configuration', () => {
         import stylex from 'stylex';
         import 'otherFile.stylex';
         import { MyTheme } from 'otherFile.stylex';
+        const _temp = {
+          "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__b69i2g" : color,
+          $$css: true
+        };
         _inject2(".__hashed_var__b69i2g{--__hashed_var__1jqb1tb:var(----__hashed_var__1jqb1tb)}", 1);
         _inject2("@property ----__hashed_var__1jqb1tb { syntax: \\"*\\"; inherits: false;}", 0);
         const styles = {
-          color: color => [{
-            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__b69i2g" : color,
-            $$css: true
-          }, {
+          color: color => [_temp, {
             "----__hashed_var__1jqb1tb": color != null ? color : undefined
           }]
         };
--- a/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
@@ -368,14 +368,13 @@ describe('Evaluation of imported values works based on configuration', () => {
         import stylex from 'stylex';
         import 'otherFile.stylex';
         import { MyTheme } from 'otherFile.stylex';
-        const _temp = {
-          "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__b69i2g" : color,
-          $$css: true
-        };
         _inject2(".__hashed_var__b69i2g{--__hashed_var__1jqb1tb:var(----__hashed_var__1jqb1tb)}", 1);
         _inject2("@property ----__hashed_var__1jqb1tb { syntax: \\"*\\"; inherits: false;}", 0);
         const styles = {
-          color: color => [_temp, {
+          color: color => [{
+            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__b69i2g" : color,
+            $$css: true
+          }, {
             "----__hashed_var__1jqb1tb": color != null ? color : undefined
           }]
         };
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
@@ -2944,12 +2944,13 @@ describe('@stylexjs/babel-plugin', () => {
         `);
         expect(code).toMatchInlineSnapshot(`
           "import * as stylex from '@stylexjs/stylex';
+          const _temp = {
+            kWkggS: "xrkmrrc",
+            kMwMTN: "xfx01vb",
+            $$css: true
+          };
           export const styles = {
-            root: color => [{
-              kWkggS: "xrkmrrc",
-              kMwMTN: color != null ? "xfx01vb" : color,
-              $$css: true
-            }, {
+            root: color => [_temp, {
               "--color": color != null ? color : undefined
             }]
           };"
@@ -3000,11 +3001,12 @@ describe('@stylexjs/babel-plugin', () => {
         `);
         expect(code).toMatchInlineSnapshot(`
           "import * as stylex from '@stylexjs/stylex';
+          const _temp = {
+            kMwMTN: "xfx01vb",
+            $$css: true
+          };
           export const styles = {
-            one: color => [{
-              kMwMTN: color != null ? "xfx01vb" : color,
-              $$css: true
-            }, {
+            one: color => [_temp, {
               "--color": color != null ? color : undefined
             }],
             two: {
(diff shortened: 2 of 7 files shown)


packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -109,6 +109,57 @@ describe('@stylexjs/babel-plugin', () => {
         `);
```
mellyeliu, line 109: can we add some tests for `stylex.create` calls in nested fns and maybe multiple stylex.create calls in a file? to ensure the hoisting works then

packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
```diff
@@ -0,0 +1,204 @@
...
+    ]),
+  );
+  return nameIdent;
+}
+
+export function pathReplaceHoisted(
+  path: NodePath<>,
+  astExpression: t.Expression,
+): void {
+  // If the object is already hoisted, leave it inlined.
```
mellyeliu:
```suggestion
  // If the object is already hoisted, leave it inlined
```

packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
```diff
@@ -0,0 +1,204 @@
...
+    if (programPath.parentPath) {
+      programPath = programPath.parentPath;
+    } else {
+      return null;
+    }
+  }
+  return programPath;
+}
+
+export function addNamedImport(
```
mellyeliu, line 79: can we abstract some of the shared code between `addNamedImport` and `addDefaultImport`?

packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
```diff
@@ -0,0 +1,204 @@
...
+  lastImport.insertAfter(
+    t.variableDeclaration('var', [
+      t.variableDeclarator(importName, identifier),
+    ]),
+  );
+
+  return importName;
+}
+
+export function isProgramLevel(path: NodePath<>): boolean {
```
mellyeliu, line 172: maybe rename, since it also returns if it's an `ExportDeclaration`?

packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
```diff
@@ -0,0 +1,204 @@
...
+export function hoistExpression(
+  path: NodePath<>,
+  astExpression: t.Expression,
+): t.Expression {
+  const programStatementPath = getProgramStatement(path);
+  if (programStatementPath == null) {
+    return astExpression;
+  }
+
+  const nameIdent = path.scope.generateUidIdentifier();
```
mellyeliu: maybe something like
```suggestion
  const hoistedIdentifier = path.scope.generateUidIdentifier();
```

packages/@stylexjs/babel-plugin/__tests__/validation-stylex-create-test.js
```diff
@@ -71,7 +71,7 @@ describe('@stylexjs/babel-plugin', () => {
             const styles = stylex.create({});
           }
        `);
-      }).toThrow(messages.ONLY_TOP_LEVEL);
+      }).not.toThrow(messages.ONLY_TOP_LEVEL);
```
mellyeliu: do we even need this message at all then?

packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+    if (programPath.isFunction()) {
+      return false;
+    }
+    programPath = parentPath;
+  }
+  return true;
+}
+
+export function getProgramStatement(path: NodePath<>): NodePath<> {
```
mellyeliu, line 194: i _think_ we need to check `programPath.isStatement()` here too. but please double check

packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
```diff
@@ -368,13 +368,14 @@ describe('Evaluation of imported values works based on configuration', () => {
         import stylex from 'stylex';
         import 'otherFile.stylex';
         import { MyTheme } from 'otherFile.stylex';
+        const _temp = {
+          "--__hashed_var__1jqb1tb": color == null ? null : "__hashed_var__b69i2g",
```
necolas: Isn't `color` undefined in this scope?

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -109,6 +109,57 @@ describe('@stylexjs/babel-plugin', () => {
...
       });
 
+      test('nested referenced style object', () => {
+        const { code, metadata } = transform(`
+          import * as stylex from '@stylexjs/stylex';
+          function fooBar() {
+            const styles = stylex.create({
```
necolas, line 116: We should probably have a discussion about the pros and cons of allowing create within arbitrary functions. It might set incorrect expectations for users that create can be used dynamically

necolas (changes requested):
Please could we break this down into different PRs? It would be good if we could fix the deopt for dynamic styles separately from a big change that might no longer require create calls at the top level. We should also talk about whether that should be limited to nesting within TS vs being allowed for arbitrary functions or blocks

packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
nmn: They are thin wrappers around functions from a library, I don’t think there’s much more scope for abstraction but I’ll take a look.

packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
nmn: That is also program level. It’s just a special case as export statements contain other statements. AFAIK, it’s the only kind of statement with a direct statement child rather than an expression.

packages/@stylexjs/babel-plugin/__tests__/validation-stylex-create-test.js
nmn: Probably not, just didn’t start deleting this yet.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
nmn: I should’ve marked this PR as a draft as it’s mostly a POC right now and the actual design needs discussion.

packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
nmn: Perhaps this can’t be hoisted at all then. This used to work differently at one point where this object was fully static, but there was a subtle bug with that.

I’ll revert this change.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
nmn: OK. I've made changes now. The important thing here is that there is no change to the actual API. The ESLint rule should still error when using `stylex.create` not at the top level.

However, I've added these tests for two reasons:
1. Flow doesn't support namespaces and our tests aren't set up for TS syntax
2. To make the transformation more reliable when used with various bundlers and transformations. It's not uncommon for various bundlers to wrap each module in a self-executing function or something like that. Today StyleX tends to break if things aren't *just* right. Making it work in more scenarios should help make the babel plugin more reliable.

packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
nmn: I don't think so. Everything is in a program and if not we want to return the top level anyway. Otherwise, things will break.

mellyeliu: don't we call `insertBefore` later on the node returned by `getProgramStatement`? if that ever returns a non-Statement then Babel will throw

this doesn't affect this PR specifically because we only call it from CallExpression paths but since this is a general purpose util it could break things in the future

nmn: A program can never have a non statement child anyway afaik. Is there an edge-case where that's not true?

mellyeliu: babel workings... yes correct

PR #1143, opened 2025-07-16
[eslint-plugin] Sort keys based on property name instead of alphabetically

lotusjohn:
It's unintuitive to read css when they are sorted alphabetically. 
This sorts based on ordering by stylelint-config-clean-order.

Please include relevant motivation and context

## Linked PR/Issues

Fixes # (issue)

## Additional Context

**Screenshot of before lint --fix**
image
image

**After lint --fix**
image

image

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
@@ -27,13 +27,13 @@ eslintTester.run('stylex-sort-keys', rule.default, {
       import * as stylex from '@stylexjs/stylex';
       const styles = stylex.create({
         main: {
+          display: 'flex',
           borderColor: {
             default: 'green',
             ':hover': 'red',
             '@media (min-width: 1540px)': 1366,
           },
           borderRadius: 10,
-          display: 'flex',
         },
         dynamic: (color) => ({
           backgroundColor: color,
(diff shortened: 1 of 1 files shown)


packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
```diff
@@ -65,10 +65,10 @@ eslintTester.run('stylex-sort-keys', rule.default, {
       const obj = { fontSize: '12px' };
       const styles = cr({
         button: {
-          alignItems: 'center',
           display: 'flex',
-          ...obj,
+          alignItems: 'center',
           alignSelf: 'center',
+          ...obj,
```
nmn, line 71: this is a bug because the object spread should not lose its position.

packages/@stylexjs/eslint-plugin/src/utils/getPropertyPriorityAndType.js
```diff
@@ -11,14 +11,484 @@
...
+  'maxWidth',
+  'minInlineSize',
+  'maxInlineSize',
+  'height',
+  'blockSize',
+  'minHeight',
+  'maxHeight',
+  'minBlockSize',
+  'maxBlockSize',
+  'margin',
```
nmn, line 210: The physical properties, such as margin-left should come after the logical properties such as margin-in line-start. This is an important guarantee to maintain as that is what we will end up applied.

nmn (changes requested):
thanks for the contribution!

Firstly, the styles are not sorted entirely by alphabetical order today, they are sorted with a combination of property specificity with a  fallback on alphabetical order.


This new approach, based on stylelint, can also make sense but:
1. It should be added as a configuration option.
2. It should continue to respect the property specificity of styleX’s compiler.

PR #1147, opened 2025-07-18
[babel-plugin] add "last media query" wins behaviour behind feature flag

mellyeliu:
combing thru the h1 stragglers...

Context: 
- in https://github.com/facebook/stylex/pull/898 we added a media query transformer to implement "last media query" wins behaviour which is expected amongst devs but didn't integrate it into the babel plugin
- previously devs would have to manually ensure this behaviour, but it's increasingly unsustainable esp with `defineConsts`/ shareable media queries

`media-query-transformer` fixes:
- reduced dfs depth by one, as by the time we're in `flatten-raw-style-obj.js` the namespace has been stripped
- also added an array check to avoid messing with `firstThatWorks` arrays

added in `babel-plugin`
- add `style-value-parser` as dependency
- added processing in `flatten-raw-style-obj.js` and update tests
- gated behind `enableLastMediaQueryWins` param


.github/workflows/tests.yml
```diff
@@ -36,6 +38,7 @@ jobs:
         with:
           node-version: '22.x'
       - run: npm install
+      - run: npm run build --workspaces --if-present
```
necolas: This already runs after install https://github.com/facebook/stylex/blob/main/package.json#L7

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "postinstall": "npm run build",
     "prepare": "husky install tools/husky",
-    "build": "npm run build --workspaces --if-present",
+    "build": "npm run build -w style-value-parser && npm run build --workspaces --if-present",
```
necolas: If we need the parser package built first it can be listed first in the workspaces array, like we did with `shared`

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -1829,12 +1830,71 @@ describe('@stylexjs/babel-plugin', () => {
...
+              root: {
+                backgroundColor: {
+                  default: 'red',
+                  '@media (min-width: 1000px)': 'blue',
+                  '@media (min-width: 2000px)': 'purple',
+                }
+              },
+            });
+          `,
+            { enableLastMediaQueryWins: true },
```
necolas: Is this test meant to have the flag enabled?

packages/@stylexjs/babel-plugin/package.json
```diff
@@ -22,7 +22,8 @@
     "@babel/types": "^7.26.8",
     "@dual-bundle/import-meta-resolve": "^4.1.0",
     "@stylexjs/stylex": "0.14.2",
-    "postcss-value-parser": "^4.1.0"
+    "postcss-value-parser": "^4.1.0",
+    "style-value-parser": "*"
```
necolas: This should probably be pinned to the current version like we do with other packages

packages/style-value-parser/package.json
```diff
@@ -1,23 +1,29 @@
 {
-  "private": true,
   "name": "style-value-parser",
   "version": "0.14.2",
-  "main": "src/index.js",
+  "private": true,
```
necolas: Prefer private packages to have this field first so it's immediately obvious they are private

packages/style-value-parser/package.json
```diff
@@ -1,23 +1,29 @@
...
+  "type": "module",
+  "main": "lib/index.js",
   "license": "MIT",
   "scripts": {
+    "prepublishOnly": "npm run build",
```
necolas: Why is this needed?

mellyeliu:
sorry this is a draft, will mark when ready

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -1817,8 +1817,66 @@ describe('@stylexjs/babel-plugin', () => {
...
+                  {
+                    "ltr": ".xrkmrrc{background-color:red}",
+                    "rtl": null,
+                  },
+                  3000,
+                ],
+                [
+                  "x1wa28t1",
+                  {
+                    "ltr": "@media (max-width: 1000px) and (not (max-width: 500px)){.x1wa28t1.x1wa28t1{background-color:blue}}",
```
nmn: This should be simplified here.

mellyeliu:
let me write out a merge intervals approach for min/max simplification

package.json
```diff
@@ -39,7 +40,7 @@
     "jest": "^30.0.0-alpha.7",
     "jest-environment-jsdom": "^30.0.0-alpha.7",
     "lint-staged": "^13.0.3",
-    "prettier": "^3.5.0",
+    "prettier": "^3.2.5",
```
mellyeliu, line 43: this version of prettier was causing some parsing issues, after quick search I pinned it to this which seemed more stable

mellyeliu:
graveyard of force pushes from debugging workflow/rebase struggles...

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -1817,8 +1817,75 @@ describe('@stylexjs/babel-plugin', () => {
...
+                backgroundColor: {
+                  default: 'red',
+                  '@media (max-width: 900px)': 'blue',
+                  '@media (max-width: 500px)': 'purple',
+                  '@media (max-width: 400px)': 'green',
+                }
+              },
+            });
+          `,
+            { enableLastMediaQueryWins: true },
```
nmn, line 1835: We might need a better name for this.

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -43,30 +39,26 @@ describe('Media Query Transformer', () => {
...
-            '@media (max-width: 768px)': '1 / -1',
-          },
+      gridColumn: {
+        default: '1 / 2',
+        '@media (min-width: 1024.01px) and (max-width: 1440px)': {
           '@media (min-width: 768.01px) and (max-width: 1024px)': '1 / 3',
           '@media (max-width: 768px)': '1 / -1',
         },
+        '@media (min-width: 768.01px) and (max-width: 1024px)': '1 / 3',
+        '@media (max-width: 768px)': '1 / -1',
```
nmn: The media queries can be flattened in this case, but this is a contrived example. Using `height` queries in the nested object is probably more useful.

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -16,24 +16,20 @@ const stylex = {
...
     const originalStyles = {
-      foo: {
-        gridColumn: {
-          default: '1 / 2',
-          '@media (max-width: 1440px)': '1 / 4',
-          '@media (max-width: 1024px)': '1 / 3',
-          '@media (max-width: 768px)': '1 / -1',
-        },
+      gridColumn: {
+        default: '1 / 2',
```
nmn, line 16: Please add examples which use a series of `min-width` instead.

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -184,22 +164,18 @@ describe('Media Query Transformer', () => {
...
     const expectedStyles = {
-      colorMode: {
-        mode: {
-          default: 'normal',
-          '@media (color) and (not (monochrome))': 'colorful',
-          '@media (monochrome)': 'grayscale',
-        },
+      mode: {
+        default: 'normal',
+        '@media (color) and (not (monochrome))': 'colorful',
```
nmn, line 250: Not a blocker for this PR, but note that this should be simplified in the future. color and monochrome should be considered mutually exclusive.

nmn (commented):
Please add a few more examples to at least show that `min-width`, `max-width` and the combination of both is implemented. We can ignore the other edge-cases for now.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
mellyeliu: maybe `enableMediaQueryOrder`

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -438,4 +421,30 @@ describe('Media Query Transformer', () => {
...
+          '@media (max-width: 1040px) and (min-width: 600px)': '1 / 3',
+        },
+      },
+    };
+
+    const expectedStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (min-width: 900px) and (max-width: 1440px) and (not ((min-width: 600px) and (max-width: 1040px)))':
```
mellyeliu, line 436: this still needs to be simplified, will handle separately

mellyeliu: https://github.com/facebook/stylex/pull/1170

vincentriemer (approved):

PR #1152, opened 2025-07-19
[style-value-parser] merge overlapping widths in media query clauses

mellyeliu:
merge intervals to simplify and normalize mqs 
- collect all min/max width rules into intervals (eg. min-width: 500px -> [500, inf], max-width: 1000px -> [-inf, 1000]
- handle not rules by inverting intervals: not (min-width: 500px) -> [-inf, 499.99]
- find intersection of all intervals: highest lower bound and lowest upper bound
- if intersection is empty (lower > upper), return not all media query
- otherwise, convert back to min/max width rules from the intersection bounds
- integrate this merging into the normalize method for 'and' rules

needed for https://github.com/facebook/stylex/pull/1147 

also fixed a bug with the 0.01 epsilon adjustment for min/max width for inequalities like 400 < width < 700

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27313,7 +27313,24 @@
       },
       "devDependencies": {
         "@babel/node": "^7.23.9",
-        "benchmark": "^2.1.4"
+        "benchmark": "^2.1.4",
+        "prettier": "^2.8.8"
+      }
+    },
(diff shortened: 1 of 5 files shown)


mellyeliu:
@nmn ik the flow types are bad....

Issue #1146, opened 2025-07-17
[babel-plugin] Using dynamic styles with undefined values doesn't match expectations


necolas:
### Describe the issue

We encountered a bug internally that was root-caused to how StyleX has implemented dynamic styles.

The issue manifests when the value used with dynamic styles is `undefined` or `null`, which results in no value being set for the custom property.

```
const styles = stylex.create({
  width: (width: ?number) => ({
    width
  })
});

styles.width(undefined)
```

The browser does not fallback to a non-var value when the var is undefined (https://codepen.io/necolas/pen/VYvvBmj/). However, in StyleX we don't set the non-dynamic class names on the element anyway, because during a property merge the class for `width` is set to the one using the custom property.

You can see that here, where there is no logic checking for undefined values to decide which block of compiled styles to use. It's always the same block.

https://github.com/facebook/stylex/blob/0.14.2/packages/%40stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js#L2032-L2053

When using inline styles with React DOM, if the value is undefined it doesn't get set and whatever value is present in your static CSS will get used. So to match this expectation is StyleX, we likely need to include more logic in the compiled output for dynamic styles, so that the relevant atomic class is not included if the input value is undefined.

We've previously discussed dynamic styles in the context of optimizing them, but the solution there should also be the foundation for fixing this bug: https://github.com/facebook/stylex/issues/737

### Expected behavior

See above

### Steps to reproduce

See above

### Test case

See above

mellyeliu:
yea, I believe this behaviour is unintentional and inconsistent, we have other tests where the classname is appropriately wrapped with null (https://github.com/facebook/stylex/blob/0.14.2/packages/%40stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js#L2156-L2168). 

I was eyeing this part of dynamic styles classname generation as strange for a few reasons. The issue is due to a pretty brittle reverse lookup of dynamic styles in the create visitor which results in classnames not actually being wrapped in null as expected. Put up a rewrite of the dynamic styles logic that walks through the styles and properly detects static vs dynamic styles that should fix the issue: https://github.com/facebook/stylex/pull/1153

https://github.com/facebook/stylex/issues/737 would still be a good follow up to https://github.com/facebook/stylex/pull/1153 at some point

nmn:
This bug was discovered previously and fixed. The css variable should not be set to anything and the actual property should be set to undefined as well which should get skipped.

[...]

Issue #737, opened 2024-10-09
Optimise dynamic styles into static styles when possible


nmn:
See #727 

---

There are cases where a developer may use dynamic styles syntax with a function but the styles themselves are conditional styles that can be known at compile time.

In such scenarios, we can compile every possible value into static classNames.

Here are some examples to clarify what should happen:

---

Source:

```ts
const styles = stylex.create({
  root: (isCondition) => ({
    width: '100%',
    maxWidth: 768,
    display: 'flex',
    color: isCondition ? 'red' : 'blue'
  }),
});
```

Output:

```ts
const _styles_root = {
  width: 'w-full', // sets `width: 100%`
  maxWidth: 'mw-768',
  display: 'd-flex',
};
const _styles_root1 = {
  $$css: true,
  color: 'c-red', // className that sets color: red;
};
const _styles_root2 = {
  $$css: true,
  color: 'c-blue', // className that sets color: blue;
};
const styles = {
  root: (isCondition) => [
    _styles_root,
    isCondition ? _styles_root1 : _styles_root2,
  ],
};
```
---

---

Source:

```ts
const styles = stylex.create({
  root: (isA, isB) => ({
    width: '100%',
    maxWidth: 768,
    display: 'flex',
    color: isA ? 'red' : 'blue',
    backgroundColor: isB ? 'yellow' : 'green',
  }),
});
```

Output:

```ts
const _styles_root = {
  width: 'w-full', // sets `width: 100%`
  maxWidth: 'mw-768',
  display: 'd-flex',
};
const _styles_root1_1 = {
  $$css: true,
  color: 'c-red', // className that sets color: red;
};
const _styles_root1_2 = {
  $$css: true,
  color: 'c-blue',
};
const _styles_root2_1 = {
  $$css: true,
  backgroundColor: 'c-yellow',
};
const _styles_root2_2 = {
  $$css: true,
  backgroundColor: 'c-green',
};
const styles = {
  root: (isA, isB) => [
    _styles_root,
    isA ? _styles_root1_1 : _styles_root1_2,
    isB ? _styles_root2_1 : _styles_root2_2,
  ],
};
```
---

The overall logic can be:
1. Separate the dynamic style object into separate chunks:
   - The complete static part
   - The parts that are conditional styles
   - The truly dynamic parts
2. Compile each of the parts separately
3. Return them all within an array.

lucas-teks:
[...]

PR #1153, opened 2025-07-20
[babel-plugin] rewrite dynamic styles construction 

mellyeliu:
Rewrite for dynamic styles handling that is sort of long overdue. previous approach is quite imprecise and brittle in the way strings are built up. this has a few changes to the existing behaviour:
- old version mapped classNames back to expressions via `origClassPaths` dict (kind of brittle reverse look up of stringified path arrays that were pretty broken - according to the tests, never fully worked) (causing issues like https://github.com/facebook/stylex/issues/1146)
- new version just inspects the metadata directly to build a list of style keys by determining which styles are expressions instead of static
- also rebuilds the string from scratch with conditional logic instead of mutating in place strangely

Testing:
- tested in `example-nextjs`
- confirmed
   - no width omitted
   - null check catches only undefined and null (meaning 0 and false still omits classname)
   - computed expressions like (width + 100) work as expected
         - undefined + 100 -> NaNpx
         - null + 100 → 100 -> 100px
   - multiple dynamic values
   - static values within a dynamic style are preserved

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
@@ -372,7 +372,7 @@ describe('Evaluation of imported values works based on configuration', () => {
         _inject2("@property ----__hashed_var__1jqb1tb { syntax: \\"*\\"; inherits: false;}", 0);
         const styles = {
           color: color => [{
-            "--__hashed_var__1jqb1tb": color == null ? null : "__hashed_var__b69i2g",
+            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__b69i2g" : color,
             $$css: true
           }, {
             "----__hashed_var__1jqb1tb": color != null ? color : undefined
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -191,7 +191,7 @@ describe('@stylexjs/babel-plugin', () => {
             $$css: "app/main.js:31"
           },
           dynamic: color => [{
-            "color-kMwMTN": "color-xfx01vb",
+            "color-kMwMTN": color != null ? "color-xfx01vb" : color,
             $$css: "app/main.js:56"
           }, {
             "--color": color != null ? color : undefined
@@ -257,7 +257,7 @@ describe('@stylexjs/babel-plugin', () => {
             $$css: "app/main.js:31"
           },
           dynamic: color => [{
-            "color-kMwMTN": "color-xfx01vb",
+            "color-kMwMTN": color != null ? "color-xfx01vb" : color,
             $$css: "app/main.js:56"
           }, {
             "--color": color != null ? color : undefined
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
@@ -2044,7 +2044,7 @@ describe('@stylexjs/babel-plugin', () => {
           export const styles = {
             root: color => [{
               kWkggS: "xrkmrrc",
-              kMwMTN: "xfx01vb",
+              kMwMTN: color != null ? "xfx01vb" : color,
               $$css: true
             }, {
               "--color": color != null ? color : undefined
@@ -2099,7 +2099,7 @@ describe('@stylexjs/babel-plugin', () => {
           "import * as stylex from '@stylexjs/stylex';
           export const styles = {
             one: color => [{
-              kMwMTN: "xfx01vb",
+              kMwMTN: color != null ? "xfx01vb" : color,
               $$css: true
             }, {
               "--color": color != null ? color : undefined
@@ -2157,8 +2157,8 @@ describe('@stylexjs/babel-plugin', () => {
           "import * as stylex from '@stylexjs/stylex';
           export const styles = {
             root: (bgColor, otherColor) => [{
-              "--background-color": bgColor == null ? null : "x15mgraa",
-              "--otherColor": otherColor == null ? null : "x1qph05k",
+              "--background-color": bgColor != null ? "x15mgraa" : bgColor,
+              "--otherColor": otherColor != null ? "x1qph05k" : otherColor,
               $$css: true
             }, {
               "----background-color": bgColor != null ? bgColor : undefined,
@@ -2221,7 +2221,7 @@ describe('@stylexjs/babel-plugin', () => {
             "import * as stylex from '@stylexjs/stylex';
             export const styles = {
               root: width => [{
-                kzqmXN: "x1bl4301",
+                kzqmXN: width != null ? "x1bl4301" : width,
                 $$css: true
               }, {
                 "--width": (val => typeof val === "number" ? val + "px" : val != null ? val : undefined)(width)
@@ -2243,6 +2243,80 @@ describe('@stylexjs/babel-plugin', () => {
                   "--width",
                   {
                     "ltr": "@property --width { syntax: "*"; inherits: false;}",
+                    "rtl": null,
+                  },
+                  0,
+                ],
+              ],
+            }
+          `);
+        });
+
+        test('set mixed values', () => {
+          const { code, metadata } = transform(`
+            import * as stylex from '@stylexjs/stylex';
+            export const styles = stylex.create({
+              root: (width) => ({
+                width,
+                backgroundColor: 'red',
+                height: width + 100,
+              })
+            });
+          `);
+          // Check that dynamic number values get units where appropriate
+          expect(code).toMatchInlineSnapshot(`
+            "import * as stylex from '@stylexjs/stylex';
+            export const styles = {
+              root: width => [{
+                kzqmXN: width != null ? "x1bl4301" : width,
+                kWkggS: "xrkmrrc",
+                kZKoxP: width + 100 != null ? "x1f5funs" : width + 100,
+                $$css: true
+              }, {
+                "--width": (val => typeof val === "number" ? val + "px" : val != null ? val : undefined)(width),
+                "--height": (val => typeof val === "number" ? val + "px" : val != null ? val : undefined)(width + 100)
+              }]
+            };"
+          `);
+          expect(metadata).toMatchInlineSnapshot(`
+            {
+              "stylex": [
+                [
+                  "x1bl4301",
+                  {
+                    "ltr": ".x1bl4301{width:var(--width)}",
+                    "rtl": null,
+                  },
+                  4000,
+                ],
+                [
+                  "xrkmrrc",
+                  {
+                    "ltr": ".xrkmrrc{background-color:red}",
+                    "rtl": null,
+                  },
+                  3000,
+                ],
+                [
+                  "x1f5funs",
+                  {
+                    "ltr": ".x1f5funs{height:var(--height)}",
+                    "rtl": null,
+                  },
+                  4000,
+                ],
+                [
+                  "--width",
+                  {
+                    "ltr": "@property --width { syntax: "*"; inherits: false;}",
+                    "rtl": null,
+                  },
+                  0,
+                ],
+                [
+                  "--height",
+                  {
+                    "ltr": "@property --height { syntax: "*"; inherits: false;}",
                     "rtl": null,
                   },
                   0,
@@ -2275,7 +2349,7 @@ describe('@stylexjs/babel-plugin', () => {
             import { vars } from 'vars.stylex.js';
             export const styles = {
               root: width => [{
-                "--x1anmu0j": width == null ? null : "x14vhreu",
+                "--x1anmu0j": width != null ? "x14vhreu" : width,
                 $$css: true
               }, {
                 "----x1anmu0j": width != null ? width : undefined
@@ -2326,8 +2400,8 @@ describe('@stylexjs/babel-plugin', () => {
             "import * as stylex from '@stylexjs/stylex';
             export const styles = {
               root: color => [{
-                kWkggS: "x1ttfofm",
-                kMwMTN: "x74ai9j",
+                kWkggS: color != null ? "x1ttfofm" : color,
+                kMwMTN: color != null ? "x74ai9j" : color,
                 $$css: true
               }, {
                 "--1e2mv7m": color != null ? color : undefined,
@@ -2393,7 +2467,7 @@ describe('@stylexjs/babel-plugin', () => {
             "import * as stylex from '@stylexjs/stylex';
             export const styles = {
               root: (hover, active, focus) => [{
-                kMwMTN: "x74ai9j x19c4yy1 x10peeyq x126ychx",
+                kMwMTN: (hover != null ? "x74ai9j" : hover) + (active != null ? "x19c4yy1" : active) + (focus != null ? "x10peeyq" : focus) + "x126ychx",
                 $$css: true
               }, {
                 "--1113oo7": hover != null ? hover : undefined,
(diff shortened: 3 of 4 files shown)


packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
```diff
@@ -372,8 +372,8 @@ describe('Evaluation of imported values works based on configuration', () => {
         _inject2("@property ----__hashed_var__1jqb1tb { syntax: \\"*\\"; inherits: false;}", 0);
         const styles = {
           color: color => [{
-            "--__hashed_var__1jqb1tb": color == null ? null : "__hashed_var__b69i2g",
-            $$css: true
+            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__b69i2g " : "",
+            $$css: ""
```
nmn: Not quite what we want here. We don’t want to fall back to an empty string. We want to fallback to the same null-ish value, undefined or null.

nmn (changes requested):

packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
mellyeliu: sure, changed to set it to the expr itself, either undefined or null

packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
```diff
@@ -372,8 +372,8 @@ describe('Evaluation of imported values works based on configuration', () => {
         _inject2("@property ----__hashed_var__1jqb1tb { syntax: \\"*\\"; inherits: false;}", 0);
         const styles = {
           color: color => [{
-            "--__hashed_var__1jqb1tb": color == null ? null : "__hashed_var__b69i2g",
-            $$css: true
+            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__b69i2g" : color,
+            $$css: ""
```
necolas: `""` should be `true`

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -191,8 +191,8 @@ describe('@stylexjs/babel-plugin', () => {
             $$css: "app/main.js:31"
           },
           dynamic: color => [{
-            "color-kMwMTN": "color-xfx01vb",
-            $$css: "app/main.js:56"
+            "color-kMwMTN": color != null ? "color-xfx01vb" : color,
+            $$css: "app/main.js:56 "
```
necolas: Extra whitespace added here

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -2252,6 +2252,80 @@ describe('@stylexjs/babel-plugin', () => {
...
+            });
+          `);
+          // Check that dynamic number values get units where appropriate
+          expect(code).toMatchInlineSnapshot(`
+            "import * as stylex from '@stylexjs/stylex';
+            export const styles = {
+              root: width => [{
+                kzqmXN: width != null ? "x1bl4301" : width,
+                kWkggS: "xrkmrrc ",
+                kZKoxP: width + 100 != null ? "x1f5funs" : width + 100,
```
necolas: If `width` is `undefined` then `width + 100` will be `NaN`, and the class will incorrectly get applied

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -2252,6 +2252,80 @@ describe('@stylexjs/babel-plugin', () => {
...
+                $$css: ""
+              }, {
+                "--width": (val => typeof val === "number" ? val + "px" : val != null ? val : undefined)(width),
+                "--height": (val => typeof val === "number" ? val + "px" : val != null ? val : undefined)(width + 100)
```
necolas, line 2277: This will produce `NaNpx` if `width` is `undefined`

necolas (changes requested):

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
nmn: I almost missed this. The extra whitespace is in the string value for `$$css`

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -257,8 +257,8 @@ describe('@stylexjs/babel-plugin', () => {
...
```
nmn: Extra whitespace in this string as well.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -2043,9 +2043,9 @@ describe('@stylexjs/babel-plugin', () => {
           "import * as stylex from '@stylexjs/stylex';
           export const styles = {
             root: color => [{
-              kWkggS: "xrkmrrc",
-              kMwMTN: "xfx01vb",
-              $$css: true
+              kWkggS: "xrkmrrc ",
```
nmn: And there's an extra space after this className.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -2157,9 +2157,9 @@ describe('@stylexjs/babel-plugin', () => {
           "import * as stylex from '@stylexjs/stylex';
           export const styles = {
             root: (bgColor, otherColor) => [{
-              "--background-color": bgColor == null ? null : "x15mgraa",
-              "--otherColor": otherColor == null ? null : "x1qph05k",
-              $$css: true
+              "--background-color": bgColor != null ? "x15mgraa" : bgColor,
+              "--otherColor": otherColor != null ? "x1qph05k" : otherColor,
+              $$css: ""
```
nmn: `$$css` should fall back to `true` when a file location isn't available.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
nmn: What else should we do here? If `width` is undefined, then the original source code is setting the value of `height` to `NaN`. Validating values here would be too expensive overall.

That said, I think we should detect binary expressions such as `+` and treat them as values that can never be nullish and skip the ternary entirely.

So, in this case, I think we should just have:
```suggestion
                kZKoxP: "x1f5funs",
```

Because the addition of two things can never be undefined.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
nmn: Again, I think we should accept it as if `width` is undefined, there is a bug in the code given to StyleX itself. Handling it manually in this expression would bloat the code everywhere for one edge-case.

packages/@stylexjs/babel-plugin/src/visitors/stylex-create.js
```diff
@@ -262,76 +267,54 @@ export default function transformStyleXCreate(
...
+                  if (expr) {
+                    exprList.push(
+                      t.conditionalExpression(
+                        t.binaryExpression('!=', expr, t.nullLiteral()),
+                        t.stringLiteral(cls),
+                        expr,
+                      ),
+                    );
+                  } else {
+                    exprList.push(t.stringLiteral(cls + ' '));
```
nmn: This is where the extra space is likely coming from.

```suggestion
                    exprList.push(t.stringLiteral(cls));
```

packages/@stylexjs/babel-plugin/src/visitors/stylex-create.js
```diff
@@ -262,76 +267,54 @@ export default function transformStyleXCreate(
...
+                  )?.expression;
+
+                  if (expr) {
+                    exprList.push(
+                      t.conditionalExpression(
+                        t.binaryExpression('!=', expr, t.nullLiteral()),
+                        t.stringLiteral(cls),
+                        expr,
+                      ),
+                    );
```
nmn, line 301: Probably something like this might work.

```suggestion
                    if (t.isIdentifier) {
                      exprList.push(
                        t.conditionalExpression(
                          t.binaryExpression('!=', expr, t.nullLiteral()),
                          t.stringLiteral(cls),
                          expr,
                        ),
                      );
                    } else {
                      exprList.push(t.stringLiteral(cls))
                    }
```

This might be too broad, so maybe look for `BinaryExpression` specifically with `+`, `*`, `-`, `/`, `**` and `UnaryExpression` with a leading `-` and `+`

nmn (changes requested):
Some small changes needed, but mostly on track!

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
mellyeliu: I avoided making this change in this PR because I wanted to get us on par with the behaviour before things broke (ie. checking the value of the expression) and mainly wanted to illustrate edge case behaviour that we could iterate on with this test. It's probably too much overhead to manually validate every malformed expression resulting from errors in the source code. If the user is authoring code that looks like `undefined + int` than it's something they should likely fix upstream anyway

I'll push up the logic for binary and urary expressions as it seems safe enough. There are also a few more optimizations (conditional exprs for one) that I can address in a follow up

necolas: The original bug we encountered is when the value is `undefined`. React DOM with inline styles will not render `height: NaNpx` if it receives `undefined + 100`. We could change our codemod to account for this shortcoming in StyleX's implementation of dynamic styles. There are lots of other problems with how dynamic styles are designed and implemented.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
necolas: It would be simple enough to not call the function if the inputs are `null` or `undefined`

mellyeliu: I considered this approach initially, but there are lots of cases where this may not work as simply as we might think. The case might be there for `undefined` (I haven't traced all possible outcomes or expressions and compared them to React DOM behaviour) but it's a lot murkier for `null` inputs.

Here's an example where the function could have difficulties. In this example, if the offset is `null` here, we probably still want to apply `padding: 100px` to the element:

```
     root: (offset) => ({
          padding: offset + 100,
        })
      });
})
```

Or something simpler like fallback cases. Here we would be omitting the classname/width when height is `null` or `undefined`, when the user would expect 100. 

```
     root: (height) => ({
          width: height ?? 100,
        })
      });
})
```

I think the issue is that nullish values don't always mean "ignore", they sometimes mean "use fallback logic". So instead of skipping function calls, we should probably preserve execution and selectively guard className output only when necessary.

Exactly when is worth a careful discussion, just need to outline the cases carefully. So just using the entire expr as a starting point. Can punt on this and work on follow up optimizations/edge cases!

packages/@stylexjs/babel-plugin/src/visitors/stylex-create.js
mellyeliu: ~I think this could be valuable in a follow up, would be good to carefully think through all these cases first imo~ 

alright fine I will add this optimization here...

mellyeliu:
Fixes for `$$css` and whitespace

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
necolas: > In this example, if the offset is null here, we probably still want to apply padding: 100px to the element

That's a good example of the issue in question. When using inline styles in React DOM the padding would *not* be `100px` if `undefined`.

> Or something simpler like fallback cases.

The reported bug is an example of where the fallback was a static style that was not applied when expected. Maybe StyleX shouldn't allow conditions within dynamic styles, as they could exist in the component body that calls the style function instead.

```
styles.root(height ?? 100)
```

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -3550,10 +3624,10 @@ describe('@stylexjs/babel-plugin', () => {
           export const styles = {
             default: margin => [{
               kWkggS: "xrkmrrc",
-              k71WvV: "x1555q52 x1bi16m7",
-              k1K539: "x1hvr6ea x3skgmg",
-              keTefX: "x1feukp3 xgzim5p",
-              keoZOQ: "x17zef60",
+              k71WvV: (margin != null ? "x1555q52" : margin) + (margin + 4 != null ? "x1bi16m7" : margin + 4),
```
necolas: These expressions could be shorter if we didn't repeat the evaluated condition if false.

As @nmn pointed out, some of these conditions can never be `null` though.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
mellyeliu: In JS, for operations like `null + 100`, the null is coerced to 0. So we would expect the React DOM inline style behaviour of ` <div style={{ padding: null + 100 }} />` -> ` <div style={{ padding: 100 }} />` to be respected, as outlined in the test plan 

What we should and shouldn't allow in dynamic styles is a separate discussion, but there are lots of instances of explicit fallback cases in dynamic styles that would break if we stopped supporting them here, so imo it doesn't seem tenable to change that here without more discussion. Happy to discuss the rest in a follow up! This is just meant to patch the current behaviour

mellyeliu:
See output here for regression and prior behaviour (https://github.com/facebook/stylex/pull/875/files#diff-1ae52714c572f4282c394af92227b4198e93ffdccd6bf7c38129b36f09250280L1756-L1760). This PR is just scoped to fix this as a baseline, and the optimization (https://github.com/facebook/stylex/issues/737) may require a longer discussion!

mellyeliu:
pushed changes for `binary` and `unary` optimization

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
necolas: Sorry I was thought we were still talking about when the value is `undefined`. You're right that the result in RDOM is different if the value is `null` due to coercion.

mellyeliu: No worries! can iterate on this

packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
mellyeliu: this is also a deviation from what we had before: https://github.com/facebook/stylex/pull/875/files#diff-1ae52714c572f4282c394af92227b4198e93ffdccd6bf7c38129b36f09250280L1756-L1760

iirc we do handle null and undefined classnames slightly differently from "" when merging styles and this is probably what we do want but just flagging another change to past behaviour

packages/@stylexjs/babel-plugin/src/visitors/stylex-create.js
```diff
@@ -32,6 +32,14 @@ import { messages } from '../shared';
 import { evaluateStyleXCreateArg } from './parse-stylex-create-arg';
 import flatMapExpandedShorthands from '../shared/preprocess-rules';
 
+function isSafeToSkipNullCheck(expr: t.Expression) {
+  return (
+    (t.isBinaryExpression(expr) &&
+      ['+', '-', '*', '/', '**'].includes(expr.operator)) ||
+    (t.isUnaryExpression(expr) && ['-', '+'].includes(expr.operator))
```
nmn-oai, line 39: Maybe omit cases where have `Ternary` expressions, `&&` and `??` where at least of the branches is known to be non-null?

One possibility is to only include the `null-ish` checking for scenrios where the value is an identifier.

This needs more thought so let's create an issue to start figuring this out.

mellyeliu: yeah, that's what I was thinking re: conditional exprs ^ https://github.com/facebook/stylex/pull/1153#discussion_r2221585245 

let me try to map this out a bit more cleanly

PR #1138, opened 2025-07-14
[Docs] Explain how `--` can be used to define manual names for CSS vars

nmn:
## What changed / motivation ?

Added new content to `learn` and `api` sections.

Also made a small tweak the latest release notes.

PR #1155, opened 2025-07-21
[docs] Documentation for creating CSS vars with stable names

### LINKED ISSUE #1116: [docs] Documentation for creating CSS vars with stable names
### Describe the issue

StyleX has a feature that allows you to opt-out of the generated, scoped var names created by `defineVars` - you create a string key with '--' prefix:

```js
export const tokens = defineVars({
  '--this-key-is-manually-created': 10
});
```

That key name can then be used as a string `create` calls:

```js
const styles = create({
  root: {
    padding: 'var(--this-key-is-manually-created)'
  }
});
```

This was introduced for internal migration purposes, but it's probably going to stick around for a while and have a similar use case for external adopters. We should document it.

necolas:
Fix #1116
Close #1138

### DIFF
--- a/packages/docs/docs/api/javascript/defineVars.mdx
+++ b/packages/docs/docs/api/javascript/defineVars.mdx
@@ -8,7 +8,7 @@ sidebar_position: 4
 
 # `stylex.defineVars`
 
-Creates global CSS Custom Properties (Variables) that can be imported and used
+Creates global CSS Custom Properties (variables) that can be imported and used
 within `create` calls anywhere within a codebase.
 
 ```ts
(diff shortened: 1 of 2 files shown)


vincentriemer (approved):

packages/docs/docs/api/javascript/defineVars.mdx
```diff
@@ -17,6 +17,9 @@ function defineVars<Styles extends {[key: string]: Value}>(
 ): Vars<{[key in keyof Styles]: string}>;
 ```
 
+By default, `defineVars` will create unique, hashed variable names. To create
+variables with stable names use a key that starts with `--`.
```
nmn, line 21: "stable" is misleading as all variable names are "stable". "custom" or "manual" are better words IMO.

PR #1161, opened 2025-07-23
[style-value-parser] merge height intervals for media queries

mellyeliu:
just adding height processing using similar logic from https://github.com/facebook/stylex/pull/1152

### DIFF
--- a/packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
+++ b/packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
@@ -182,6 +182,60 @@ describe('Media Query Transformer', () => {
     expect(JSON.stringify(result)).toBe(JSON.stringify(expectedStyles));
   });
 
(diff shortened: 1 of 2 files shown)


vincentriemer (approved):

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -182,6 +182,60 @@ describe('Media Query Transformer', () => {
...
+  test('basic usage: multiple heights', () => {
+    const originalStyles = {
+      foo: {
+        gridRow: {
+          default: '1 / 2',
+          '@media (max-height: 1200px)': '1 / 4',
+          '@media (max-height: 900px)': '1 / 3',
+          '@media (max-height: 600px)': '1 / -1',
+        },
+      },
```
nmn, line 194: Missing example which are chains of `min-height` and also where the input has ranges already.

Issue #1162, opened 2025-07-24
[babel-plugin] Make `defineConsts` work with firstThatWorks


mellyeliu:
Need to add special handling within `processStylexRules` for `defineConsts` `firstThatWorks` integration. Either with multiple regex handling or token parser

nmn:
IMO, this is not a high priority as constants should be treated as literal constants only. This is already not supported for `defineVars`. This is because of how CSS variables work where pretty much any value is always valid. The recommendation is to use `@supports` queries explicitly instead of `firstThatWorks` for these cases.

Although technically it is possible to support `firstThatWorks` for `defineConsts` I think it's actually better to be consistent with vars and not support it.

mellyeliu:
let's punt on this then

PR #1163, opened 2025-07-24
[babel-plugin] document fail cases for defineConsts

mellyeliu:
Let's add some documentation for known fail cases for `defineConsts` as I circle around to them

See https://github.com/facebook/stylex/issues/1154, https://github.com/facebook/stylex/issues/1162

### DIFF
--- a/packages/docs/docs/api/javascript/defineConsts.mdx
+++ b/packages/docs/docs/api/javascript/defineConsts.mdx
@@ -10,7 +10,7 @@ sidebar_position: 3
 
 Defines static style constants that can be used directly in `create` calls anywhere in the codebase. Unlike `defineVars`, these values are inlined at build time and do not generate actual CSS variables.
 
-*Note: `defineConsts` is an experimental feature and cannot currently be used with `firstThatWorks` or when `runtimeInjection` is enabled.*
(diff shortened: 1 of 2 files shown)


Jta26 (approved):

packages/docs/docs/api/javascript/defineConsts.mdx
```diff
@@ -10,7 +10,7 @@ sidebar_position: 3
 
 Defines static style constants that can be used directly in `create` calls anywhere in the codebase. Unlike `defineVars`, these values are inlined at build time and do not generate actual CSS variables.
 
-*Note: `defineConsts` is an experimental feature and cannot currently be used with `firstThatWorks` or when `runtimeInjection` is enabled.*
+*Note: `defineConsts` is an experimental feature and cannot currently be used with `firstThatWorks`, dynamic styles, or when `runtimeInjection` is enabled.*
```
nmn, line 13: 1. `firstTheWorks` will never be supported. It's not really supported for variables in general and you're supposed to use `@supports` to check what is valid instead.

2. dynamic Styles should work very easily if we just export object value given to `defineConsts` as is in the generated JS.

Issue #1069, opened 2025-05-20
0.13.0 Broken defineVars type check


abrcdf1023:
### Describe the issue

defineVars type check broken when use number

![Image](https://github.com/user-attachments/assets/3e55eee8-1432-4b67-8455-ca0108ec4795)

### Expected behavior

Should not broken type check

### Steps to reproduce

Just definVar with number value

import * as stylex from '@stylexjs/stylex'
import { type VarGroup } from '@stylexjs/stylex'

export type ZIndex = Readonly<{
  appBar: number,
  drawer: number,
  backdrop: number,
  fab: number,
  mobileStepper: number,
  modal: number,
  snackbar: number,
  speedDial: number,
  tooltip: number,
}>

export const zIndex: VarGroup<ZIndex> = stylex.defineVars({
  appBar: 1100,
  drawer: 1200,
  backdrop: 1201,
  fab: 1050,
  mobileStepper: 1000,
  modal: 1300,
  snackbar: 1400,
  speedDial: 1050,
  tooltip: 1500,
})

hrasoa:
Same issue here:

<img width="889" height="523" alt="Image" src="https://github.com/user-attachments/assets/c74353de-5982-4b1c-bdcf-e480e00ed13d" />

hrasoa:
Here is a repro: https://github.com/hrasoa/stylex-repro, check `App.tsx` and `zIndex-token.styles.ts`. It's broken from `0.13.0`

nmn:
Thanks for the repro. I can see what the issue is and a few possible ways to fix it.

The issue is specific to `zIndex` and other properties that don't accept strings as values.
We enforce the usage of strings when defining variables to communicate that we never convert numbers to pixels in that context since we can't know how you expect to use the value.

This case isn't handled properly. Also, I don't we handle `stylex.types.*` types correctly today.

nmn:
@hrasoa Hey, looks like the bug is already fixed in the latest version of StyleX, although you will need to make a small change to your code.

You will need to use `stylex.types.number` to wrap the number values in your `defineVars` call:

```ts
export const zIndexToken = stylex.defineVars({
  "z-1": stylex.types.number(1),
});
```

Once you make this, it should work correctly with StyleX 14.x.

---

In case you need a smaller patch in the shorter term, you will need to patch the type `ComplexStyleValueType` within `StyleXTypes.d.ts`.

```ts
type ComplexStyleValueType<+T> =
  T extends StyleXVar<infer U>
[...]

PR #1165, opened 2025-07-28
[fix] Improve stylex.types.* types and fix typescript types/setup

nmn:

The tests for Typescript were broken due to a broken import. 
This has been fixed by updating the `exports` declaration in the `@stylexjs/stylex` package.

Further, `stylex.types.number` and `stylex.types.integer` were incorrectly allowing arbitrary strings to be passed in. This has been updated to only allow numbers. (I'm looking to check if there are specific constants like `PI` in CSS that should be allowed)

I've added another typescript test which validates the case mentioned in #1069

Finally, all the existing tests under the `nextjs-example` have been moved to `typescript-tests`.

## Pre-flight checklist


pawelblaszczyk5:
I thought the entry point was removed on purpose and that was also the reason for settling up these: https://github.com/facebook/stylex/pull/1124 😃

nmn:
Thanks for the catch. I'll verify that the typescript tests are where they should be.

PR #1166, opened 2025-07-28
[fix] Fix media query parser for conditions with number values

nmn:

There was a bug in the media query parser where it couldn't parse queries such as
- `@media (-webkit-min-device-pixel-ratio: 2)`

After some investigation, it turned out that this was because the parser did not support bare numbers as values in media query conditions. This has now been fixed.

## Additional Context

Additional unit tests have been added to ensure more media queries can now be parsed as expected.

## Pre-flight checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
       "prettier-plugin-hermes-parser"
     ],
     "proseWrap": "always",
+    "trailingComma": "all",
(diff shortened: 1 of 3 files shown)


mellyeliu (approved):
amazing!!!!!

PR #1168, opened 2025-07-28
[docs] Update outdated or misleading theming docs

nmn:

There were some docs about theming that were outdated and incorrect. Thanks to @BMCwebdev for bringing it to our attention in https://github.com/facebook/stylex/discussions/947

I updated that and clarified some sentences that could be misleading.

## Pre-flight checklist

### DIFF
--- a/packages/docs/docs/api/javascript/defineVars.mdx
+++ b/packages/docs/docs/api/javascript/defineVars.mdx
@@ -18,7 +18,9 @@ function defineVars<Styles extends {[key: string]: Value}>(
 ```
 
 By default, `defineVars` will create unique, hashed variable names. To create
-variables with stable names use a key that starts with `--`.
+variables with custom names use a key that starts with `--`. These will generate
+CSS custom properties with the provided name instead of generating a globally
+unique name.
 
 ### Example use:
 
--- a/packages/docs/docs/learn/05-theming/01-defining-variables.mdx
+++ b/packages/docs/docs/learn/05-theming/01-defining-variables.mdx
@@ -48,9 +48,11 @@ export const tokens = stylex.defineVars({
 });
 ```
 
-This function, too, is processed at compile-time and unique CSS variable names are automatically generated. These values can then be imported and used within `create` calls.
+This function, too, is processed at compile-time and unique CSS variable names are automatically generated. 
+These values can then be imported and used within `create` calls.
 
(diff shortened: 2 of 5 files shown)


packages/docs/docs/learn/05-theming/01-defining-variables.mdx
```diff
@@ -48,9 +48,11 @@ export const tokens = stylex.defineVars({
 });
 ```
 
-This function, too, is processed at compile-time and unique CSS variable names are automatically generated. These values can then be imported and used within `create` calls.
+This function, too, is processed at compile-time and unique CSS variable names are automatically generated. 
+These values can then be imported and used within `create` calls.
 
-To create variables with stable names that match the exact strings provided, use a key that starts with `--`. Doing so will mean that those variables are not unique to a given `defineVars` call.
+To create variables with custom stable names that match the exact strings provided, use a key that starts with `--`.
+Doing so will mean that StyleX cannot guarantee that those variables names are unique to a given `defineVars` call.
```
mellyeliu, line 55: this is a good callout bc iirc it caused some dsp breakage cc: @irenegu9823

packages/docs/docs/learn/05-theming/03-creating-themes.mdx
```diff
@@ -61,9 +61,8 @@ const styles = stylex.create({
 <div {...stylex.props(dracula, styles.container)}>{children}</div>;
 ```
 
-**NOTE:** All variables in a variable group must be overridden when creating a
-theme. This choice has been to help catch accidental omissions. Even at the cost
-of occasional tedium.
+**NOTE:** Any variables that are not overridden will revert back to their
```
mellyeliu, line 64: should we highlight the empty {} reset option (ik it's in recipes but could be useful to surface)

packages/docs/docs/api/javascript/defineVars.mdx
```diff
@@ -18,7 +18,9 @@ function defineVars<Styles extends {[key: string]: Value}>(
 ```
 
 By default, `defineVars` will create unique, hashed variable names. To create
-variables with stable names use a key that starts with `--`.
+variables with custom names use a key that starts with `--`. These will generate
```
mellyeliu, line 21:
```suggestion
variables with custom names, use a key that starts with `--`. These will generate
```

PR #1169, opened 2025-07-28
[babel-plugin] optimize dynamic styles output for conditionals and te…

mellyeliu:
Let's optimize for dynamic styles outputs ie. don't emit a ternary wrapping dynamic styles when the expression is "safe" (ie. we can confirm the expression is not `null` or `undefined`). 

We have coverage for `binary` and `unary` expressions, but there are a few more cases we can optimize for:
- template literals
- logical expressions (recursively check sides) 
   - nullish coalescing `??` (if left side is safe)
   - logical or `||`    (if either side is safe)
   - logical `&&` (both sides are safe ie. a string/binary/unary/template literal of course). 
   
   

image


mellyeliu:
double check my boolean logic plz....

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -2029,6 +2029,550 @@ describe('@stylexjs/babel-plugin', () => {
...
+          `);
+          expect(code).toMatchInlineSnapshot(`
+            "import * as stylex from '@stylexjs/stylex';
+            export const styles = {
+              root: color => [{
+                kWkggS: "xr5ldyu",
+                kMwMTN: "xfx01vb",
+                $$css: true
+              }, {
+                "--backgroundColor": \`\${color}\` != null ? \`\${color}\` : undefined,
```
nmn, line 2051: The inline style should be same to simplify too.

mellyeliu:
note: we can also extend this to the inline styles part as well

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -2029,6 +2029,550 @@ describe('@stylexjs/babel-plugin', () => {
...
+          `);
+        });
+
+        test('logical expressions with safe right side', () => {
+          const { code, metadata } = transform(`
+            import * as stylex from '@stylexjs/stylex';
+            export const styles = stylex.create({
+              root: (color) => ({
+                backgroundColor: 'red' || color,
+                color: 'black' || color,
```
nmn, line 2329: Kinda meaningless, but OK.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -2029,6 +2029,550 @@ describe('@stylexjs/babel-plugin', () => {
...
+            }
+          `);
+        });
+
+        test('conditional expressions with safe branches', () => {
+          const { code, metadata } = transform(`
+            import * as stylex from '@stylexjs/stylex';
+            export const styles = stylex.create({
+              root: (color, isDark) => ({
+                backgroundColor: isDark ? 'black' : 'white',
```
nmn, line 2435: Another possibility for future improvement. We could generate two static classNames at compile time for such cases.

packages/@stylexjs/babel-plugin/src/visitors/stylex-create.js
```diff
@@ -32,12 +32,45 @@ import { messages } from '../shared';
 import { evaluateStyleXCreateArg } from './parse-stylex-create-arg';
 import flatMapExpandedShorthands from '../shared/preprocess-rules';
 
-function isSafeToSkipNullCheck(expr: t.Expression) {
-  return (
-    (t.isBinaryExpression(expr) &&
-      ['+', '-', '*', '/', '**'].includes(expr.operator)) ||
-    (t.isUnaryExpression(expr) && ['-', '+'].includes(expr.operator))
-  );
+function isSafeToSkipNullCheck(expr: t.Expression): boolean {
```
nmn, line 35: This code looks fine.

nmn (approved):

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
mellyeliu: added in issue as follow up

PR #1170, opened 2025-07-29
[style-value-parser] process disjoint, intersecting, and subset ranges in media queries

mellyeliu:
Updated mergeIntervalsForAnd to properly merge and simplify ranges for min-width / max-width pairs:
- build numeric intervals for each dimension (width/height). it collects all the lower and upper bounds from the rules and determines the tightest range:
   - for min-width it takes the largest lower bound
   - for max-width it takes the smallest upper bound

Added handling for the special case of overlapping queries inside an outer range. If there's a gap in the valid intervals, the function now splits the outer range into valid disjoint intervals on either side of the hole.
 - if the hole fully overlaps the lower or upper side of the outer range, we drop the empty side
 - if both sides are valid, we return them as an or group so the final query represents both disjoint intervals
 - if only one side is valid, we just return that branch directly instead of wrapping it in or

Added tests for all combinations of ranges:
- disjoint ranges (completely separate, media query should be preserved)
- overlapping ranges (just decrease the bounds of the parent interval)
- subset ranges (split into two ranges, a, c becomes a, b and b, c)

### DIFF
--- a/packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
+++ b/packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
@@ -450,7 +450,7 @@ describe('Media Query Transformer', () => {
           default: '1 / 2',
           '@media (max-width: 1440px) and (min-width: 900px)': '1 / 4',
           '@media (max-width: 800px) and (min-width: 600px)': '1 / 3',
-          '@media (max-width: 500px)': '1 / 1',
+          '@media (max-width: 500px) and (min-width: 400px)': '1 / 1',
         },
       },
     };
@@ -461,7 +461,7 @@ describe('Media Query Transformer', () => {
           default: '1 / 2',
           '@media (min-width: 900px) and (max-width: 1440px)': '1 / 4',
           '@media (min-width: 600px) and (max-width: 800px)': '1 / 3',
-          '@media (max-width: 500px)': '1 / 1',
+          '@media (min-width: 400px) and (max-width: 500px)': '1 / 1',
         },
       },
     };
@@ -511,7 +511,7 @@ describe('Media Query Transformer', () => {
       foo: {
         gridColumn: {
           default: '1 / 2',
-          '@media (min-width: 900px) and (max-width: 999.99px) or (min-width: 1100.01px) and (max-width: 1440px)':
+          '@media ((min-width: 900px) and (max-width: 999.99px)) or ((min-width: 1100.01px) and (max-width: 1440px))':
             '1 / 4',
           '@media (min-width: 1000px) and (max-width: 1100px)': '1 / 3',
           '@media (min-width: 400px) and (max-width: 500px)': '1 / 1',
@@ -538,7 +538,7 @@ describe('Media Query Transformer', () => {
       foo: {
         gridColumn: {
           default: '1 / 2',
-          '@media (min-width: 900px) and (max-width: 999.99px) or (min-width: 1100.01px) and (max-width: 1440px)':
+          '@media ((min-width: 900px) and (max-width: 999.99px)) or ((min-width: 1100.01px) and (max-width: 1440px))':
             '1 / 4',
           '@media (min-width: 1000px) and (max-width: 1100px)': '1 / 3',
         },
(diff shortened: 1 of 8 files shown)


packages/style-value-parser/src/at-queries/media-query.js
```diff
@@ -514,14 +539,23 @@ export class MediaQuery {
           : `(not ${this.#toString(queries.rule)})`;
       case 'and':
         return queries.rules.map((rule) => this.#toString(rule)).join(' and ');
-      case 'or':
+      case 'or': {
+        const validRules = queries.rules.filter(
```
mellyeliu, line 543: ideally this edge case should be handled in the `normalize` function, but I had difficulty structuring the recursion in a way that handled all edge cases - open to improvements

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -418,7 +418,59 @@ describe('Media Query Transformer', () => {
...
+        gridColumn: {
+          default: '1 / 2',
+          '@media (max-width: 1440px) and (min-width: 900px)': '1 / 4',
+          '@media (max-width: 800px) and (min-width: 600px)': '1 / 3',
+          '@media (max-width: 500px) and (min-width: 400px)': '1 / 1',
+        },
+      },
+    };
+
+    const expectedStyles = {
```
mellyeliu, line 458: for disjoint ranges we expect the output to be the same (as the developer has already ensured they do not overlap)

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -443,4 +494,87 @@ describe('Media Query Transformer', () => {
...
+          '@media (max-width: 1050px) and (min-width: 1010px)': '1 / -1',
+        },
+      },
+    };
+
+    const expectedStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (min-width: 900px) and (max-width: 999.99px) or (min-width: 1100.01px) and (max-width: 1440px)':
```
mellyeliu: these ranges are narrowed to whatever new min/max restriction from the queries below from the intersecting parts

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -443,4 +494,87 @@ describe('Media Query Transformer', () => {
...
+          '@media (max-width: 1100px) and (min-width: 1000px)': '1 / 3',
+        },
+      },
+    };
+
+    const expectedStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (min-width: 900px) and (max-width: 999.99px) or (min-width: 1100.01px) and (max-width: 1440px)':
```
mellyeliu: these are now split into two ranges because the below query creates a "hole"

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
```diff
@@ -443,4 +494,87 @@ describe('Media Query Transformer', () => {
...
+          '@media (max-width: 500px) and (min-width: 400px)': '1 / 1',
+        },
+      },
+    };
+
+    const expectedStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (min-width: 900px) and (max-width: 999.99px) or (min-width: 1100.01px) and (max-width: 1440px)':
```
nmn:
```suggestion
          '@media ((min-width: 900px) and (max-width: 999.99px)) or ((min-width: 1100.01px) and (max-width: 1440px))':
```

The toString part of this has a bug.

packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
nmn: Same, needs brackets around the "and" ranges.

nmn (changes requested):
This is nearly ready. The simplification logic seems to be working well.

All that remains is the one small bug fix when stringifying disjoint ranges.

mellyeliu:
fixed parenthesis bug in ranges in `toString`

vincentriemer (approved):

PR #1178, opened 2025-07-31
[babel-plugin] Support for descendent/sibling selectors

nmn:

There's a long standing idea to add support for descendent and sibling selectors. We found a solution to do this without breaking any of the principles or guarantees that StyleX provides but we held off for a few reasons:
- We wanted to avoid expanding the API, and so wanted to see if it was actually needed.
  - Time and again, this keeps coming up as a feature request and developers have been using workarounds to achieve this.
- The actual API design was hard to get right and we couldn't really make a decision
  - It's probably better to put this out there and debate the exact naming now
- We kind of wanted to wait for `:has()` to be broadly available so the API felt better.
  - This was not a hard blocker, but the existence of `:has()` makes the API feel much more complete.
  
### How it works?

This PR introduces a set of new helper functions:
- `stylex.when.ancestor(pseudo: `:${string}`, customMarker?: TMarker)`
- `stylex.when.descendent(pseudo: `:${string}`, customMarker?: TMarker)`
- `stylex.when.siblingBefore(pseudo: `:${string}`, customMarker?: TMarker)`
- `stylex.when.siblingAfter(pseudo: `:${string}`, customMarker?: TMarker)`
- `stylex.when.sibling(pseudo: `:${string}`, customMarker?: TMarker)`

These are essentially just helper function that generate special pseudo classes similar to `:hover`.
- e.g, `when.ancestor(':hover)` returns `":where(.stylex-default-marker:hover *)"`

The important thing here is that this still is applied on the element being styled.

```tsx
import { when, create, props, defaultMarker } from '@stylexjs/stylex';

const styles = create({
    foo: {
      backgroundColor: {
        default: 'blue',
        [when.ancestor(':hover')]: 'red',
      },
    },
});

<div {...props(defaultMarker)}>
  <div {...props(styles.foo)}> Some Content </div>
</div>
```

`styles.foo` "observes any ancestor element with the `defaultMarker` and becomes active when that element is hovered.

You have to manually "mark" the element being observed and then observe it in your styles directionally.

We will provide a default marker for the common case and the ability to create custom markers for the uncommon case.

### How it avoid breaking any StyleX guarantees or principles

There are two important guarantees that are at risk when trying to implement such a feature:

1. "no styling at a distance"
  - This implementation does not introduce any styling at a distance. It enables new kinds of "observers", which work similarly to `:hover` and media queries, but enables *observing* some other element. So we "observe a distance". We never "style at a distance".
2. "Reliable merging"
  - This implementation builds on how we enable `:hover` etc, and so merging rules remain unchanged.
  
  
### What remains:

- [x] The APIs for
  - [x] `stylex.defaultMarker`
  - [x] `stylex.defineMarker()` 
    - (will be used similar to `defineConsts` and `defineVars`, but will be a single value with no args)  
- [ ] Updating `stylex.create` to ensure that these new selectors have a lower specificity than the usual pseudo classes
- [ ] Testing many more edge-cases.

## Linked PR/Issues

Numerous requests

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/when-functions-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/when-functions-test.js
@@ -47,7 +47,7 @@ describe('@stylexjs/babel-plugin', () => {
         "import { when, create } from '@stylexjs/stylex';
         const styles = {
           container: {
-            kWkggS: "x1t391ir x26ppzc",
+            kWkggS: "x1t391ir x148kuu",
             $$css: true
           }
         };
@@ -65,9 +65,9 @@ describe('@stylexjs/babel-plugin', () => {
             3000,
           ],
           [
-            "x26ppzc",
+            "x148kuu",
             {
-              "ltr": ".x26ppzc:where(.stylex-target:hover *){background-color:red}",
+              "ltr": ".x148kuu:where(.x-default-marker:hover *){background-color:red}",
               "rtl": null,
             },
             3040,
@@ -96,7 +96,7 @@ describe('@stylexjs/babel-plugin', () => {
         "import { when, create } from '@stylexjs/stylex';
         const styles = {
           container: {
-            kWkggS: "x1t391ir xeb64px",
+            kWkggS: "x1t391ir x1i6rnlt",
             $$css: true
           }
         };
@@ -114,9 +114,9 @@ describe('@stylexjs/babel-plugin', () => {
             3000,
           ],
           [
-            "xeb64px",
+            "x1i6rnlt",
             {
-              "ltr": ".xeb64px:where(.stylex-target:focus ~ *){background-color:red}",
+              "ltr": ".x1i6rnlt:where(.x-default-marker:focus ~ *){background-color:red}",
               "rtl": null,
             },
             3040,
@@ -146,7 +146,7 @@ describe('@stylexjs/babel-plugin', () => {
         "import * as stylex from '@stylexjs/stylex';
         const styles = {
           container: {
-            kWkggS: "x1t391ir x26ppzc xhhxj74",
+            kWkggS: "x1t391ir x148kuu xpijypl",
             $$css: true
           }
         };
@@ -164,17 +164,17 @@ describe('@stylexjs/babel-plugin', () => {
             3000,
           ],
           [
-            "x26ppzc",
+            "x148kuu",
             {
-              "ltr": ".x26ppzc:where(.stylex-target:hover *){background-color:red}",
+              "ltr": ".x148kuu:where(.x-default-marker:hover *){background-color:red}",
               "rtl": null,
             },
             3040,
           ],
           [
-            "xhhxj74",
+            "xpijypl",
             {
-              "ltr": ".xhhxj74:where(.stylex-target:focus ~ *){background-color:green}",
+              "ltr": ".xpijypl:where(.x-default-marker:focus ~ *){background-color:green}",
               "rtl": null,
             },
             3040,
@@ -204,7 +204,7 @@ describe('@stylexjs/babel-plugin', () => {
         "import { when as w, create } from '@stylexjs/stylex';
         const styles = {
           container: {
-            kWkggS: "x1t391ir x26ppzc xhhxj74",
+            kWkggS: "x1t391ir x148kuu xpijypl",
             $$css: true
           }
         };
@@ -222,17 +222,17 @@ describe('@stylexjs/babel-plugin', () => {
             3000,
           ],
           [
-            "x26ppzc",
+            "x148kuu",
             {
-              "ltr": ".x26ppzc:where(.stylex-target:hover *){background-color:red}",
+              "ltr": ".x148kuu:where(.x-default-marker:hover *){background-color:red}",
               "rtl": null,
             },
             3040,
           ],
           [
-            "xhhxj74",
+            "xpijypl",
             {
-              "ltr": ".xhhxj74:where(.stylex-target:focus ~ *){background-color:green}",
+              "ltr": ".xpijypl:where(.x-default-marker:focus ~ *){background-color:green}",
               "rtl": null,
             },
             3040,
--- a/packages/@stylexjs/babel-plugin/__tests__/when-functions-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/when-functions-test.js
@@ -33,9 +33,10 @@ describe('@stylexjs/babel-plugin', () => {
         
         const styles = create({
             container: {
-            [when.ancestor(':hover')]: {
-                backgroundColor: 'red',
-            },
+              backgroundColor: {
+                default: 'blue',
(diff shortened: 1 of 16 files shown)


necolas:
There is a potential performance issue with this approach. When a browser encounters a selector like `.class *`, it first has to find all elements on the page (right-to-left matching). Then, for each of those elements, it has to traverse up the DOM tree to see if it's a descendant of an element using `class`. On a large page with a deep DOM tree, this can be computationally very expensive and can lead to a noticeable increase in style recalculation time. We've seen cases at Meta where use of the `*` selector with StyleX is estimated to be responsible for 10% of recalculation times.

Expensive selectors like the universal selector (`*`) are the core problem. Even if the whole selector doesn't match anything on the page, the use of `*` on the right makes them expensive. My understanding is that `:where()` only changes the specificity and not the matching algorithm. The browser engine must still traverse the DOM to find matches. So the use of `:where()` doesn't make this part of the process faster.

The performance cost of pseudo-classes like `:hover` is another consideration. When an element is hovered, the browser has to re-evaluate styles, which can trigger a "restyle" event. Combining this with a descendant universal selector means the browser may need to re-evaluate the style for all children of the hovered element.

fabiospampinato:
> e.g, when.ancestor(':hover) returns ":where(.stylex-default-marker:hover *)"

Compiling to selectors that effectively end in "*", if I'm understanding the proposal right, seems wild to me. Presumably there should be a principle/guarantee that idiomatic usage of StyleX will always lead to sane selectors, otherwise any performance-conscious user of it would be walking on egg shells, but I wouldn't consider those kinds of selectors sane selectors.

nmn:
@necolas Regarding the `.class *` selector, this will only be used alongside another class selector always, so it won't have the same performance pitfall.

The actual CSS being generated will be `.child-class:where(.anc-class:hover *)` which is equivalent to `.anc-class:hover .child`. Sorry if I didn't make this clear in my write-up.

And if this implementation is still a problem, we can *actually* simplify the selector to literally being `.anc-class:hover .child` in the generated CSS. Happy to add that to the todo list if you think that's important. I didn't do it for small edge-case selector specificity reasons. (`:where` doesn't increase specificity)

cc: @fabiospampinato

packages/@stylexjs/babel-plugin/__tests__/when-functions-test.js
```diff
@@ -0,0 +1,280 @@
...
+            {
+              "ltr": ".x1t391ir{background-color:blue}",
+              "rtl": null,
+            },
+            3000,
+          ],
+          [
+            "x26ppzc",
+            {
+              "ltr": ".x26ppzc:where(.stylex-target:hover *){background-color:red}",
```
nmn: @necolas @fabiospampinato 

Example of the actual selector being generated so far.

packages/@stylexjs/babel-plugin/flow_modules/@babel/traverse/index.js.flow
```diff
@@ -632,10 +632,10 @@ declare export class NodePath<+T: Node = Node> {
   getCompletionRecords(): Array<NodePath<>>;
 
   getSibling(key: string | number): NodePath<>;
-  getPrevSibling(): NodePath<>;
-  getNextSibling(): NodePath<>;
-  getAllPrevSiblings(): Array<NodePath<>>;
-  getAllNextSiblings(): Array<NodePath<>>;
+  getsiblingBefore(): NodePath<>;
```
mellyeliu: what's up with the casing here?
```suggestion
  getSiblingBefore(): NodePath<>;
```

packages/@stylexjs/babel-plugin/__tests__/when-functions-test.js
```diff
@@ -0,0 +1,280 @@
...
+              "rtl": null,
+            },
+            3040,
+          ],
+        ]
+      `);
+    });
+
+    test('when.siblingBefore function', () => {
```
mellyeliu, line 79: can we add a test for each selector here?

nmn:
Almost there, a few more fixes for the initial implementation.

necolas:
What's the expected behavior when nesting components? For example, can hovering an element far up the tree change the appearance of children? https://codepen.io/necolas/pen/pvjVXrz

The ancestor one is going to be most commonly used. But given the use of `*` selectors it's not out of the question that CSS could have worse perf than JS in scenarios where the scoped subtree is very large. Less concerned with the other selectors. It would be useful to have performance research carried out

packages/@stylexjs/babel-plugin/flow_modules/@babel/traverse/index.js.flow
nmn: This is likely wrong, but this copied from a TS type file. Will double check this at some other time.

packages/@stylexjs/babel-plugin/__tests__/when-functions-test.js
```diff
@@ -0,0 +1,337 @@
...
+        const styles = create({
+            container: {
+              backgroundColor: {
+                default: 'blue',
+                [when.ancestor(':hover')]: 'red',
+              },
+            },
+        });
+
+        console.log(styles.container);
```
mellyeliu, line 43: remove

packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js
```diff
@@ -58,19 +58,49 @@ export default function transformStylexProps(
...
+      case 'ObjectExpression':
+      case 'Identifier':
       case 'MemberExpression': {
-        const resolved = parseNullableStyle(arg, state);
+        const resolved = parseNullableStyle(
+          argPath,
+          state,
+          evaluatePathFnConfig,
+        );
+        console.log('resolved', resolved);
```
mellyeliu: remove

nmn:
While chasing down why `stylex.props` wasn't pre-computing the new selectors, I found out that it was missing many other expressions as well so this PR now also has a noticeable improvement to `stylex.prop` that will be useful outside of just the new feature.

packages/@stylexjs/babel-plugin/src/shared/when/when.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Creates selector that observes if the given pseudo-class is
+ * active on a descendant with the "defaultMarker"
+ *
+ * @param pseudo - The pseudo selector (e.g., ':hover', ':focus')
+ * @returns A :has() clause for the descendant observer
+ */
+export function descendant(pseudo: string): string {
+  validatePseudoSelector(pseudo);
+  return `:has(.x-default-marker${pseudo})`;
```
mellyeliu: shouldn't we pass `options.classNamePrefix` here to align with the default marker impl?

mellyeliu:
lgtm aside from aligning `when` to use the same prefix as `defaultMarker` (though I don't know if users actually set this commonly). And some failing tests

mellyeliu (approved):
lggg

PR #1180, opened 2025-08-01
[legacy][babel-plugin] Expand listStyle

necolas:
Prevents style resolving bug involving listStyle, listStyleImage, listStylePosition, and listStyleType. Although the legacy strategy doesn't expand all shorthands, this one in particular is a problem because we commonly use listStyle in base components and listStyleType in product code that is meant to override the base component styles.

StyleX does not prevent unpredictable mixing of `listStyle` with `listStyleType` etc longform properties. This was the root cause of an internal bug.

## Additional Context

This change reduced the total amount of CSS in the majority of internal style sheets.


packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/legacy-expand-shorthands.js
```diff
@@ -83,6 +83,15 @@ import splitValue from '../utils/split-css-value';
...
 type TReturn = $ReadOnlyArray<[string, TStyleValue]>;
 
+const listStyleGlobalValues = new Set<string>([
+  'inherit',
+  'initial',
+  'revert',
+  'unset',
+]);
+const listStylePositionValues = new Set<string>(['inside', 'outside']);
+const listStyleTypeRegex = /^[a-z-]+$/;
```
nmn: list-style-type also accepts strings, so: `"\1F44D"` is valid and so is `'\1F44D'`

nmn (approved):
One of the reasons this shorthand wasn't expanded before is that using the shorthand in all scenarios works well. It's highly unlikely that you'd ever need both `list-style-type` and `list-style-image` on an element.

So while this might work as short term fix, I think the better fix for the code internally might be to disallow the constituent longhands of `listStyle` and enforce the usage of the shorthand `listStyle` in all scenarios.

That will lead to overall best performance as there will be a single class instead of 3 on HTML elements and the CSS size would be similar or slightly smaller.

---

Also, before merging this, you'd need to verify if a CSS variable is being used as the value for `listStyle` and if the variable happens to contain anything but an image value. If so, this would cause regressions.

---

So, overall, tread carefully, but the expansion code itself is OK, outside of the one edge case I commented about. (and the failing flow types)

necolas:
>  I think the better fix for the code internally might be to disallow the constituent longhands of listStyle and enforce the usage of the shorthand listStyle in all scenarios.

Not sure it makes sense to introduce an exception where longform properties are banned and the shortform is required. We have code that mixes position and type styles too.

> That will lead to overall best performance as there will be a single class instead of 3 on HTML elements and the CSS size would be similar or slightly smaller.

That ship has already sailed for the legacy strategy. By far the biggest contributors to classList length are the padding, margin, border, etc., expansions.

> Also, before merging this, you'd need to verify if a CSS variable is being used as the value for listStyle

There isn't a single internal case of using a variable in `listStyle` that I could find. So it's probably best to error on use of CSS variables since we can't know which property the value will represent.

packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/legacy-expand-shorthands.js
necolas: We don't appear to use strings with escape sequences for list-style-type, so will punt on that edge case

necolas:
Circling back to confirm that this patch appears to resolve the issue we encountered internally, without requiring changes to product code.

PR #1183, opened 2025-08-01
[babel-plugin] Reduce chances of dynamic variable name collisions

necolas:
This is a simple fix for a bug that Figma encountered. The CSS custom properties that StyleX generates for dynamic styles can easily collide with existing CSS vars during a code migration. Prefixing them with `x-` makes that much less likely.


dwei-figma:
💖

nmn (approved):

nmn:
Lgtm. Surprising, that the additional `--` wasn't enough, but this makes sense.

necolas:
@nmn I think they have some existing vars like `--color` that collide with the ones we generate, e.g.,

image

PR #1184, opened 2025-08-01
[release] v0.15.1

mellyeliu:
- Hoist stylex.create and static className objects to the top level for support inside functions
- Mark `style-value-parser` as public

Will be easier/more maintainable than inlining across packages

### DIFF
--- a/packages/style-value-parser/package.json
+++ b/packages/style-value-parser/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "style-value-parser",
-  "version": "0.15.0",
+  "version": "0.15.1",
   "type": "module",
   "main": "lib/index.js",
   "license": "MIT",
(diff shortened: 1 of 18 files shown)


mellyeliu:
publishing to unblock users, will move to @stylexjs path in later follow up

necolas:
Please can we include #1183 to unblock Figma

packages/style-value-parser/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
```
necolas, line 2: Where is the part that marks "style-value-parser as public"? We shouldn't have to make it public either, as it can be inlined in the build without needing to maintain another public API.

mellyeliu: Thanks for the catch! I think the change got lost somewhere in the push: https://github.com/facebook/stylex/pull/1185

PR #1185, opened 2025-08-01
[style-value-parser] mark package as public

mellyeliu:


necolas:
We can't simply mark this package as public, since there is already an npm package called `style-value-parser`...which contained malicious code https://www.npmjs.com/package/style-value-parser

necolas:
See if #1186 works better. It's a less invasive change and doesn't force us to publish the parser to npm

CI is failing because the import paths in the babel plugin have not been updated. Nor has the workspace path to the parser in the root `package.json`, or the path in the flowconfig.

```
import { lastMediaQueryWinsTransform } from 'style-value-parser';
```

image

mellyeliu:
closing for https://github.com/facebook/stylex/pull/1186, can revisit this as needed

PR #1186, opened 2025-08-01
[babel-plugin] Exclude private dependencies

necolas:


mellyeliu (approved):
Thanks! Let's go with this, it's the better fix rn!

mellyeliu:
tested on `example-nextjs` and seems to be working fine with the inline build. feel free to make a patch with this and the figma fix

necolas:
FWIW, the parser was getting inlined from the moment it was imported into the babel-plugin, because the rollup build inlines any import that isn't explicitly marked as an external. But when the parser was added to the package deps, it created an install error when used outside the monorepo. Inlining private, shared packages (not declared in package deps) is something we've been doing with success in the React repo for a long time, as a way to both improve performance and avoid publishing what amounts to private APIs that might churn a lot more than the main library that uses them.

necolas:
@mellyeliu 0.15.2 PR https://github.com/facebook/stylex/pull/1187

Issue #1191, opened 2025-08-06
Dynamic styles which includes media/conditional styles are incorrectly compiled since 0.14.3


pawelblaszczyk5:
### Describe the issue

Hi 👋

I've tried to bump StyleX at work project today and found out that there was some regression which I bisected to 0.14.3 version. When compiling dynamic styles which includes media queries/pseudo-classes/pseudoe-lements/:is etc. it concatenates classes for these without spaces.

Example, for this input (for the sake of example I created dynamic styles which doesn't accept arguments and aren't truly dynamic but it doesn't change the behavior and it's minimal version to reproduce):

```tsx
const example = stylex.create({
  static: {
    color: {
      "@media (min-width:1024px)": "green",
      default: "red",
    },
  },
  dynamicMedia: () => ({
    backgroundColor: {
      "@media (min-width:1024px)": "purple",
      default: "gray",
    },
  }),
  dynamicConditional: () => ({
    fontSize: {
      ":is([data-hovered])": 45,
      default: 35,
    },
  }),
});
```

We're getting this output:

```tsx
const example = {
  static: {
    kMwMTN: "x1319odh x1e2nbdu",
    $$css: true
  },
  dynamicMedia: () => [{
    kWkggS: "xcu4pi3x1cbl2y9",
    $$css: true
  }, {}],
  dynamicConditional: () => [{
    kGuDYH: "x35k2qlx1gwigmq",
    $$css: true
  }, {}]
};
```

You can see that for dynamic styles there're missing spaces in classes - `x35k2qlx1gwigmq ` instead of `x35k2ql x1gwigmq ` this causes styles to not be properly applied - stylesheet is generated correctly. 

```css
@layer priority1;
@layer priority1{
.x1cbl2y9{background-color:gray}
.x1e2nbdu{color:red}
.x1gwigmq{font-size:35px}
.x35k2ql:is([data-hovered]){font-size:45px}
@media (min-width:1024px){.xcu4pi3.xcu4pi3{background-color:purple}}
@media (min-width:1024px){.x1319odh.x1319odh{color:green}}
}
```

### Expected behavior

The output includes spaces between classes correctly

### Steps to reproduce

1. Use StyleX >= 0.14.3
2. Use conditional styles inside of dynamic styles
3. Observe styles not being applied correctly

Repo with reproduction: https://github.com/pawelblaszczyk5/stylex-reproduction

### Test case

```tsx
const example = stylex.create({
  dynamicMedia: () => ({
[...]

PR #1193, opened 2025-08-07
[babel-plugin] fix classname construction for dynamic contextual styles

mellyeliu:
Bug introduced in https://github.com/facebook/stylex/pull/1153, missing space concat for contextual styles. When we are constructing a key with multiple classnames (contextual style) we need to add a space suffix to each classname except the last. Previous logic included this `const suffix = index === classList.length - 1 ? '' : ' ';` but was not ported over.

Fixes https://github.com/facebook/stylex/issues/1191


pawelblaszczyk5:
Would it be possible to make some small patch release with it? Or is there some bigger release coming soonish anyway? Thanks in advance 😄

mellyeliu:
of course! putting a patch up here https://github.com/facebook/stylex/pull/1198

pawelblaszczyk5:
Thank you so much!

PR #1194, opened 2025-08-07
[babel-plugin] Optimize precomputed props call in JSX

necolas:
Optimizes simple cases of `{...stylex.props()}` in JSX to avoid the overhead of extra object creation that the Babel jsx transform introduces.

This simple input:

```
<div {...stylex.props(styles.text)}>
  hi
</div>;
```

was being transformed into:

```
React.jsx(
  "div",
  babelHelpers["extends"](
    {className: "xw7yly9"},
    {children: "hi"}
  )
);
```

To prevent creating temporary objects, and dynamically iterating and merging props, this patch transforms the input to static object

```
React.jsx(
  "div",
  {
    className: "xw7yly9",
    children: "hi"
  }
);
```

It also adds a warning in the case where a `className` prop is found after the `{...stylex.props}`.

At the moment we only have 2 or 3 props that `stylex.props` returns, but the code is written in a generalized way so it should still work if that changes in the future. This patch does not target the inlining of `stylex.props` calls that contain conditions.

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js
+++ b/packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js
@@ -202,6 +202,48 @@ export default function transformStylexProps(
     // convert resolvedStyles to a string + ternary expressions
     // We no longer need the keys, so we can just use the values.
     const stringExpression = makeStringExpression(resolvedArgs);
+
+    // Check if this is used as a JSX spread attribute and optimize
+    // the output to avoid object creation and Babel helper
+    if (path.parentPath.node.type === 'JSXSpreadAttribute') {
+      if (
+        t.isObjectExpression(stringExpression) &&
+        stringExpression.properties.length > 0 &&
+        stringExpression.properties.every(
+          (prop) =>
+            t.isObjectProperty(prop) &&
+            (t.isIdentifier(prop.key) || t.isStringLiteral(prop.key)) &&
+            !prop.computed,
+        )
+      ) {
+        // Convert each property to a JSX attribute
+        const jsxAttributes = stringExpression.properties
+          .filter((prop) => t.isObjectProperty(prop))
+          .map((prop) => {
+            const objectProp = prop;
+            const key = objectProp.key;
+            let attrName = '';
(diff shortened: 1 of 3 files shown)


packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js
```diff
@@ -202,6 +202,113 @@ export default function transformStylexProps(
...
+          // Find existing attributes with same names
+          const attributesToRemove: number[] = [];
+          const attributesToSkip = new Set<string>();
+
+          jsxOpeningElement.attributes.forEach((attr, index) => {
+            if (t.isJSXAttribute(attr) && t.isJSXIdentifier(attr.name)) {
+              const attrName = attr.name.name;
+              if (stylexAttributeNames.has(attrName)) {
+                if (index < spreadAttributeIndex) {
+                  // Attribute before spread can be removed
```
kassens: technically, it could have side effects:

```
{
  foo: alert('hi'),
  ...{ foo: 1 },
}
```
but probably okay to ignore?

packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js
```diff
@@ -202,6 +202,113 @@ export default function transformStylexProps(
...
+                  attributesToRemove.push(index);
+                } else if (index > spreadAttributeIndex) {
+                  // Attribute after spread can be warned about
+                  attributesToSkip.add(attrName);
+                  console.warn(
```
kassens: Are you using warnings in other places? I'm not sure how that's typically surfaced (in OSS as well as internally)

packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js
kassens: The alternative would be to just bail on duplicate arguments and not apply this optimization if you don't know what to safely do.

necolas:
Changes: don't try to dedupe jsx attributes, leave that edge case to the React runtime. Remove warning log

vincentriemer (approved):

PR #1202, opened 2025-08-20
[babel plugin] avoid mutating rules in processStylexRules

henryqdineen:

We are evaluating StyleX and I ran into a bug in our WIP bundler integration that was caused by `processStylexRules` mutating rule objects. We were caching rules and it caused broken behavior for incremental compilations.

My fix in this PR is to remove the mutation of the rules. I have considered that consumers could just clone the rules before they pass it to `processStylexRules()` but I think it is better to remove the mutation so others don't run into similar hard to track down bugs. 

My fix is fairly naive in that it always shallow clones the `styleObj`. I was favoring simplicity here but if we want to be more perf conscious I have a version that [only clones when necessary](https://github.com/henryqdineen/stylex/commit/4da3e7b280763d234cae7227d80080914c0007f2).

I don't love the unit test but it serves it's purpose. If you would prefer I could add a unit test that more closely recreates the issue I was seeing. 

Thanks!

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -291,5 +291,33 @@ describe('@stylexjs/babel-plugin', () => {
         }"
       `);
     });
+
+    test('no mutation of rules', () => {
+      const { metadata } = transform(fixture);
+
+      // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze#deep_freezing
+      function deepFreeze(object) {
+        const propNames = Reflect.ownKeys(object);
+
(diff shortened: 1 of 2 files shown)


necolas:
Did you try creating a deep clone of the rules at the very top of the `processStylexRules` function? I believe the metadata can be serialized, so `JSON.parse(JSON.stringify(rules))` might be all that's needed. In a large codebase the rules object can get very large, so it might be really slow.

henryqdineen:
> Did you try creating a deep clone of the rules at the very top of the `processStylexRules` function?

No but I would expect that you work but be less performant. I find the `JSON.stringify()` can be especially slow with large object. Are you suggesting me to implement that in this PR?

henryqdineen:
Thanks @nmn! I just rebased and fixed the conflict.

PR #1212, opened 2025-09-02
[babel-plugin] support configurations within `processStylexRules`

mellyeliu:
To enable us to use `processStylexRules` internally, let's port some of the special handling we do in Haste to the OSS implementation. We only need to apply this to the legacy style resolution, which is only used inside of Meta. It's more performant to move this logic here than to do additional post processing again after grabbing the outputted rules

This includes:
- Adding generalizable config to `processStylexRules` to replace the previous useLayers param (backwards compatible via boolean check) 
- Adding `/* @ltr begin */` comments
    - The logical style polyfills that generate these comments are being removed, but we need to keep inline `float` polyfills for a while longer, and for legacy browsers
- Adding `theme` specificity bumps globally. We do this to bump the specificity of `createTheme` above `defineVars` with multiple sheets, but we can also do it everywhere (see: https://www.internalfb.com/diff/D66392781) 

### Testing:

- We expect no changes to OSS `styleResolution`s. The use of `processStylexRules` will be gated.
- No visual regressions to themes and LTR/RTL rendering

image

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/index.js
+++ b/packages/@stylexjs/babel-plugin/src/index.js
@@ -358,8 +358,18 @@ export type Rule = [
 ];
 function processStylexRules(
   rules: Array<Rule>,
-  useLayers: boolean = false,
+  config?:
+    | boolean
+    | {
+        useLayers?: boolean,
+        enableLTRRTLComments?: boolean,
+        ...
+      },
 ): string {
+  const {
+    useLayers = false,
+    enableLTRRTLComments: _enableLTRRTLComments = false,
+  } = typeof config === 'boolean' ? { useLayers: config } : config ?? {};
   if (rules.length === 0) {
     return '';
   }
--- a/packages/@stylexjs/babel-plugin/src/index.js
+++ b/packages/@stylexjs/babel-plugin/src/index.js
@@ -368,7 +368,7 @@ function processStylexRules(
 ): string {
   const {
     useLayers = false,
-    enableLTRRTLComments: _enableLTRRTLComments = false,
+    enableLTRRTLComments = false,
   } = typeof config === 'boolean' ? { useLayers: config } : config ?? {};
   if (rules.length === 0) {
     return '';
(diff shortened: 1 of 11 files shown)


packages/@stylexjs/babel-plugin/src/index.js
```diff
@@ -478,11 +482,33 @@ function processStylexRules(
...
+          }
+
           return rtlRule
-            ? [
-                addAncestorSelector(ltrRule, "html:not([dir='rtl'])"),
-                addAncestorSelector(rtlRule, "html[dir='rtl']"),
-              ]
+            ? styleResolution === 'legacy-expand-shorthands'
+              ? [
+                  `/* @ltr begin */${ltrRule}/* @ltr end */`,
```
mellyeliu, line 507: the LTR/RTL comments go around the whole rule, so we don't need to do extra handling for media queries

packages/@stylexjs/babel-plugin/src/index.js
```diff
@@ -359,6 +359,10 @@ export type Rule = [
 function processStylexRules(
   rules: Array<Rule>,
   useLayers: boolean = false,
+  styleResolution?:
+    | 'application-order'
+    | 'property-specificity'
+    | 'legacy-expand-shorthands',
```
nmn, line 365:
```suggestion
  config?: boolean | {
    useLayers?: boolean,
    rtlHandling?: "inline-comment" | null,
    ...
  }
```

packages/@stylexjs/babel-plugin/src/index.js
```diff
@@ -359,6 +359,10 @@ export type Rule = [
...
 ): string {
   if (rules.length === 0) {
```
nmn:
```suggestion
  config?: boolean | {
    useLayers?: boolean,
    rtlHandling?: "inline-comment" | null,
    ...
  }
): string {
  const {useLayers = false, rtlHandling = null} = typeof config === 'boolean'
    ? { useLayers: config }
    : config ?? {};
  if (rules.length === 0) {
```

nmn (approved):

PR #1218, opened 2025-09-07
[feat] Simple `sx={}` syntax as an alternative to stylex.props

nmn:
This adds a small syntactic sugar to allow using a `sx=` prop on lower-case html JSX elements as an alternative to `{...stylex.props}`

So you can write:

```tsx
<div sx={[styles.foo, styles.bar]} />
```

instead of 

```tsx
<div {...stylex.props(styles.foo, styles.bar)} />
```

if you want.

This addresses a very common complaint.

## Additional Context

A single test has been added.

This PR is not quite ready to merge yet. The transformation is currently hardcoded.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-props-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-props-test.js
@@ -103,6 +103,43 @@ describe('@stylexjs/babel-plugin', () => {
         `);
       });
 
+      test('Test that sx attribute can be used instead of ...stylex.props', () => {
+        expect(
+          transform(
+            `
+            import stylex from 'stylex';
+            const styles = stylex.create({
+              red: {
+                color: 'red',
+              }
+            });
+            function Foo() {
+              return (
+                <>
+                  <div id="test" sx={styles.red}>Hello World</div>
+                  <div className="test" sx={styles.red} id="test">Hello World</div>
+                  <div id="test" sx={styles.red} className="test">Hello World</div>
+                </>
+              );
+            }
+          `,
+            options,
+          ),
+        ).toMatchInlineSnapshot(`
+          "import _inject from "@stylexjs/stylex/lib/stylex-inject";
+          var _inject2 = _inject;
+          import stylex from 'stylex';
+          _inject2(".color-x1e2nbdu{color:red}", 3000);
+          function Foo() {
+            return <>
+                            <div id="test" className="color-x1e2nbdu" data-style-src="npm-package:components/Foo.react.js:4">Hello World</div>
+                            <div className="test" className="color-x1e2nbdu" data-style-src="npm-package:components/Foo.react.js:4" id="test">Hello World</div>
+                            <div id="test" className="color-x1e2nbdu" data-style-src="npm-package:components/Foo.react.js:4" className="test">Hello World</div>
+                          </>;
+          }"
+        `);
+      });
+
       test('local dynamic styles', () => {
         expect(
           transform(
--- a/packages/@stylexjs/babel-plugin/src/index.js
+++ b/packages/@stylexjs/babel-plugin/src/index.js
@@ -278,6 +278,40 @@ function styleXTransform(): PluginObj<> {
         },
       },
 
+      JSXOpeningElement(path: NodePath<t.JSXOpeningElement>) {
+        const node = path.node;
+        if (
+          node.name.type !== 'JSXIdentifier' ||
+          typeof node.name.name !== 'string' ||
+          node.name.name[0] !== node.name.name[0].toLowerCase()
+        ) {
+          return;
+        }
+        // console.log(path.node.attributes);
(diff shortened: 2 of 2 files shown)


xrzhuang (commented):
omg i need this now !!!! 

ship it 🚀

mellyeliu:
not sure if we could use this internally with Flow etc, but will try to collect feedback on naming regardless

matclayton:
This would be amazing, if you could get it working on flow especially. We're migrating to stylex at the moment and this is one of my personal pet peeves. We're also on flow (and relay), could you get the flow team to add a custom config like relay has `relay_integration=true` so it becomes stylex aware, and handles it accordingly?

zaydek:
This would be a big help. Would love to see this so I don't have to alternate between `{...stylex.props(…)}` and `sx={…}` depending on if it's an intrinsic element or a component.

mellyeliu:
Will sync with the Flow team on support for this

matclayton:
@mellyeliu did you hear from the flow team. Super interested as we’re starting a large styled components to stylex conversion on a flow code base. This was one of the major sticking points internally and it would be great to know the direction you guys are thinking about.

refirst11:
```jsx
<div stylex={[styles.foo, styles.bar]} />
```
It's a naming problem, but in the case of `sx={}`, I feel like it's easy to confuse it with the method of directly describing the object of the **component library**. Intuitively, I thought of `stylex={}` as an upgraded version of React's `style={}`.  
That said, I understand the ergonomics argument. If typos are a concern, they could be caught with ESLint rules. 

Just wanted to share this perspective.

mellyeliu:
@matclayton will prioritize this soon! hoping to get some broader Flow integration help from the team

henryqdineen:
So with TypeScript would consumers have to do something like this or would the StyleX typedefs set it?

```
declare module 'react' {
  interface HTMLAttributes<T> {
    sx?: StyleXStyles 
  }  
}
```

predaytor:
may I ask what progress has been made on this? really neat addition!

skovhus:
We are also curious how settled the API is? Would really love to start using this, and are okay with patching if the API is fairly stable.

predaytor:
> ```js
> <div stylex={[styles.foo, styles.bar]} />
> ```
> 
> It's a naming problem, but in the case of `sx={}`, I feel like it's easy to confuse it with the method of directly describing the object of the **component library**. Intuitively, I thought of `stylex={}` as an upgraded version of React's `style={}`. That said, I understand the ergonomics argument. If typos are a concern, they could be caught with ESLint rules.
> 
> Just wanted to share this perspective.

I too share this concern. `sx={}` seems a bit short. `stylex={}` or being able to specify your own keyword in the babel config would be a plus.

mellyeliu:
I haven't forgotten about this. Will share an update by next week, thank you all for the patience!

mellyeliu:
The Flow team is investigating adding a flowconfig or something similar for this, cc: @panagosg7

skovhus:
@mellyeliu thanks for giving an update. As this seems to be part the roadmap, would it be possible to release this as an experimental feature without flow support? Or at least confirm if the API is settled?

mellyeliu:
We'll merge this soon! The `sx` name will be configurable.

skovhus:
Yay! Our test harness in [styled-components-to-stylex-codemod](https://github.com/skovhus/styled-components-to-stylex-codemod) found two cases where this breaks:
- Polymorphic <Component> tags — when a wrapper uses as prop, the JSX tag is a variable like Component (uppercase). The babel plugin only processes lowercase tags, so sx passes through as a regular DOM prop.
- Member expression tags like <animated.div> — the babel plugin checks for JSXIdentifier type, but member expressions are JSXMemberExpression, so they're skipped.

Are these known limitations?

martpie:
I have also been playing with it and it seems the sx prop does not work when passed to external (I think?) React components?

Here's an example

```tsx
export default function Icon(props: Props) {
  const { name } = props;
  const IconImpl = icons[name];

  if (IconImpl === undefined) {
    return (
      <span sx={styles.icon}>
        ?
      </span>
    );
  }

  return (
    <IconImpl // this is a React Componment generated via vite+svgr, but this happens with anything, like a react-router link.
      // only this works, not sx, using `sx` will make the whole component render nothing!
      {...stylex.props(styles.icon)}
    />
  );
}
```

It's not a big problem but when trying to create typescript types for it, that gets more complicated as something like the following:

```ts
import type { CompiledStyles, StyleXArray } from '@stylexjs/stylex';

declare module 'react' {
  interface DOMAttributes<T> {
    sx?: StyleXArray<null | undefined | CompiledStyles | boolean>;
  }
}
```

will allow the `sx` prop on any component, regardless if it's "external" or not. So it's basically impossible to detect when a component can use `sx` or not (or simply, when it will render nothing because of the sx prop mis-usage). Probably related to #1360 as well.

Happy to help test workarounds or fixes if that can help.

PR #1219, opened 2025-09-07
[babel-plugin] support `defineConsts` var() overrides

mellyeliu:
Adds support for overriding `defineConsts` variables that are set to `var(--*)` references when used in `stylex.create` to support Figma migration to StyleX.

- Previously, constants defined as CSS variable indirections could not be overridden in component create calls because the key reference was not changed. 
- We now map the `--consthash` value to the CSS variable name (`--background-color` in the example)
   - We replace `var(--hash)` with the `constValue` as always
   - We replace `--hash:` with the extracted `--varName:` for both simple variables `var(--varName)` and fallback variables `var(--varName, fallback)`. The `:` is to ensure we're replacing keys and not other edge cases that interfere with debug classnames 
- Constants now behave more like passthroughs: if the referenced CSS variable has an override, the override will be applied

We generally discourage this level of abstraction for maintenance reasons but want to provide optionality for users

```tsx
const colors = stylex.defineConsts({
  bg: "var(--background-color)",
  accent: "var(--accent, lightblue)"
});
```

Can now be used like so, similar to `defineVars`

```tsx
  root: {
    [colors.bg]: "red",
    [colors.accent]: "darkblue"
  }
});
```

Tested on NextJS app for the following cases:
- Handling of regular values
- Handling of simple variables like `var(--background-color)`
- Handling of fallback vars `var(--background-color, fallback)`, testing with dynamic variables, etc

As a follow up,`transform-process` tests should to be rewritten to properly `defineConsts` transforms

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/index.js
+++ b/packages/@stylexjs/babel-plugin/src/index.js
@@ -446,11 +446,26 @@ function processStylexRules(
       let original = styleObj[dir];
 
       for (const [varRef, constValue] of constsMap.entries()) {
-        if (typeof original === 'string') {
-          const replacement = String(constValue);
-          original = original.replaceAll(varRef, replacement);
-          styleObj[dir] = original;
+        if (typeof original !== 'string') continue;
+
+        const replacement = String(constValue);
+
+        original = original.replaceAll(varRef, replacement);
+
+        // When the replacement is a variable, we need to replace the key to allow variable overrides
+        if (replacement.startsWith('var(') && replacement.endsWith(')')) {
+          const inside = replacement.slice(4, -1).trim();
+          // Account for fallback variables
+          const commaIdx = inside.indexOf(',');
+          const targetName = (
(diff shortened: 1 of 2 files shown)


packages/@stylexjs/babel-plugin/src/index.js
```diff
@@ -448,7 +448,19 @@ function processStylexRules(
...
           const replacement = String(constValue);
-          original = original.replaceAll(varRef, replacement);
+          if (
+            replacement.startsWith('var(--') &&
+            replacement.endsWith(')') &&
+            !replacement.includes(',')
+          ) {
+            const varName = replacement.slice(4, -1);
+            const constName = varRef.slice(4, -1);
+            original = original.replace(/\{([^}]*)\}/g, (match, inner) => {
```
nmn: Might want to tweak this logic a bit.

I'd go with:

```
.replaceAll(constName + ':', varName + ':')
.replaceAll('var(' + constName, 'var(' + varName)
```

vincentriemer (approved):

Issue #1042, opened 2025-05-02
Recommended replacement for the deprecated webpack plugin


hauxir:
I'm currently using webpack in combination with esbuild-loader, and have been relying on the @stylexjs/webpack-plugin for Stylex integration.
Since this plugin is now deprecated, what is the recommended replacement or migration path?

hauxir:
anyone?

necolas:
You can use the postcss plugin - https://stylexjs.com/docs/learn/installation/

necolas:
I renamed this issue because we still need to update the webpack example to use an officially supported solution.

SukkaW:
Just a head up: If you do not want to use Babel and just want to use StyleX with esbuild/swc/oxc, etc., you can give my third-party webpack solutions [`stylex-webpack`](https://github.com/SukkaW/stylex-webpack) a shot.

It also works perfectly with Next.js, both Pages Router and App Router, w/o disabling SWC, the only one that can do so currently.

PR #1228, opened 2025-09-14
[example] add webpack example

RavenColEvol:
This PR adds webpack example app with stylex setup.

## Linked PR/Issues

Fixes https://github.com/facebook/stylex/issues/1042

## Additional Context
Screenshot 2025-09-14 at 7 54 37 PM

Tests:
- Made sure important scripts are present like dev, serve, build
- Made sure build is created properly with a separate css file with stylex styles

## Pre-flight checklist


examples/example-webpack/src/app.css
```diff
@@ -0,0 +1,29 @@
...
+:root {
+  --fg1: hsl(0, 0%, 0%);
+  --bg1: hsl(276, 17%, 96%);
+  --pink-h: 295;
+  --pink-s: 62%;
+  --pink-l: 60%;
+  --cyan-h: 202;
+  --cyan-s: 100%;
+  --cyan-l: 50%;
+}
```
nmn: Can we move these to a StyleX `defineVars` call to also show that theming is working as expected?

RavenColEvol: I thought of it @nmn but I was getting linting error because of hsl color. Let me see what i can do.

RavenColEvol: Added @nmn 
Screenshot 2025-09-15 at 10 43 42 AM
Screenshot 2025-09-15 at 10 43 57 AM
Screenshot 2025-09-15 at 10 44 05 AM

examples/example-webpack/package.json
```diff
@@ -5,17 +5,36 @@
   "description": "A simple webpack example to test stylexjs/webpack-plugin",
   "main": "index.js",
   "scripts": {
-    "example:build": "webpack",
-    "example:dev": "webpack --watch"
```
RavenColEvol, line 9: @necolas I felt having `example:` prefix is bit uncommon for me as a developer so I changed it to dev, serve and build. Is there any intentional purpose behind the prefix?

examples/example-webpack/src/app.css
RavenColEvol: @nmn @necolas  this is unrelated to this PR but I think we should have same example app for all examples. I see all of them differ. If you and other team members think we can have this ( webpack example app ) common to all we can do that. This will keep consistency and we can share modules between examples app so there is no duplication and future changes will be easy to adapt across examples.

mellyeliu (approved):
i'm a fan of the homescreen!

examples/example-webpack/README.md
```diff
@@ -1,3 +1,124 @@
...
+  rules: {
+    '@stylexjs/valid-styles': 'error',
+    '@stylexjs/no-unused': 'error', 
+    '@stylexjs/valid-shorthands': 'warn',
+    '@stylexjs/sort-keys': 'warn'
+  },
+};
+```
+
+### 6. CSS Entry Point (`src/app.css`)
```
mellyeliu: nit: can we sentence case these?

examples/example-webpack/package.json
necolas: So we don't build the examples in CI: https://github.com/facebook/stylex/blob/main/package.json#L9

RavenColEvol: Will revert it back with example: prefix then

examples/example-webpack/README.md
RavenColEvol: Changed @mellyeliu does it work

mellyeliu: meaning first word capitalization, but it's inconsistent throughout the app, we can fix this in a patch sometime

PR #1238, opened 2025-09-23
[feat] Add defineMarker() for custom markers for selector combinators

nmn:

The previous PR adds support for descendent/sibling selectors but only with a defaultMarker. That works for many common cases, but sometimes more that one marker is needed to differentiate between two ancestors, or two siblings etc.

This PR adds an API for defining custom markers with the `defineMarker` API.

## Additional Context

I updated the next.js example to use the new feature and it works as expected.
There are also some unit tests to verify that everything produces the code we expect.

## Pre-flight checklist

### DIFF
--- a/examples/example-nextjs/components/Card.tsx
+++ b/examples/example-nextjs/components/Card.tsx
@@ -8,8 +8,13 @@
  */
 
 import * as stylex from '@stylexjs/stylex';
-import { globalTokens as $, spacing, text, colors } from '@/app/globalTokens.stylex';
-import { tokens } from '@/app/CardTokens.stylex';
+import {
+  globalTokens as $,
+  spacing,
+  text,
+  colors,
+} from '@/app/globalTokens.stylex';
+import { cardMarker, headingMarker } from './CardTokens.stylex';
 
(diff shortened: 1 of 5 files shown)


nmn:
This feature works as expected with one exception: It has the same flaw as media queries. Where, if you use `ancestor:hover` with two separate custom markers, the generated CSS is always ordered alphabetically. This may or may not be what you intended.

We can solve this with negation the same way as we intend to solve media queries. We can simply chain `:not()` selectors at the end.

---

Other than that one note, this feature is ready for review.

mellyeliu:
Didn't we agree the `:not()` selector approach for pseudos would balloon CSS size a bit too much? Perhaps it's fine here because this will be used a lot less than pseudos in general

nmn:
> Didn't we agree the :not() selector approach for pseudos would balloon CSS size a bit too much? Perhaps it's fine here because this will be used a lot less than pseudos in general

This is true and in general we don't want to do this. I think we would need to be selective and only use `:not` when there are two or more `when` selectors in the same direction with custom markers.

PR #1240, opened 2025-09-23
fix: Allow `defineConsts` to take numbers as values with TypeScript

dwei-figma:

There seems to be a divergence between the type of `TTokens` in Flow vs. in TypeScript. This causes `defineConsts` to reject number values in TypeScript, like so:
```ts
export const zIndex = stylex.defineConsts({
  // Type 'number' is not assignable to type 'string | CSSType<string | number | null> | { [key: string]: string; }'.ts(2322)
  connector: 1,
  bifurcatedPicker: 2,
})
```

Sync the type of `TTokens` in `packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts` against that of `packages/@stylexjs/stylex/src/types/StyleXTypes.js` to fix this.

## Additional Context

Test case added to `packages/typescript-tests/src/typetests.ts`.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
+++ b/packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
@@ -248,9 +248,9 @@ export type IDFromVarGroup<T extends VarGroup<{}>> = T['__opaqueId'];
 
 type TTokens = Readonly<{
   [key: string]:
+    | NestedVarObject<null | string | number>
+    | StyleXVar<null | string | number>
     | CSSType<null | string | number>
-    | string
-    | { [key: string]: string };
 }>;
 
 type UnwrapVars<T> = T extends StyleXVar<infer U> ? U : T;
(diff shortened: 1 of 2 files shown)


packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
```diff
@@ -264,7 +264,7 @@ type NestedVarObject<T> =
   | T
   | Readonly<{
       default: NestedVarObject<T>;
-      [key: `@${string}`]: NestedVarObject<T>;
+      [key: AtRuleStr]: NestedVarObject<T>;
```
dwei-figma, line 267: `AtRuleStr` is an alias for the template string type ``` `@${string}` ``` earlier in the file, so this should have no effect. 

Was looking to understand why this restriction doesn't seem to currently apply to `defineVars` and `defineConsts`. Seems like it's because these two take a generic that extends `TTokens` instead of `TTokens` directly. 

Not sure if our intention is to actually clamp down, but if we do, I can look into how to make this work.

PR #1242, opened 2025-09-25
Feat/descendent sibling

nmn:

Fix bugs when importing values from `.stylex.js` filed.


mellyeliu:
@nmn looks like there's still a few failures. will get you a repro soon

mellyeliu:
will attempt a resync soon

Issue #1201, opened 2025-08-19
[eslint-plugin] Split up `valid-styles`


mellyeliu:
Constant ask from ESLint folks. Most common complaint with linting is that valid-styles as a rule is not descriptive and difficult to understand. It handles a lot of things: property validation, type checking, object structure validation, pseudo-selectors and at-rules, border shorthand expansion, and converts nonstandard properties to current ones.

There's also some duplication with `valid-shorthands`. 

Let's consolidate and split up logic where it makes sense

jcperez-ch:
This got closed when landing the first merge, this issue is still open and in development.

mellyeliu:
i think the git ui scans for "Fixes XXX" and closes regardless :D

PR #1246, opened 2025-09-26
[eslint] Split nonstandard rules into stylex-no-nonstandard-styles.js

jcperez-ch:

We wan to split fixable rules related to non-standard properties and values on a separate eslint rule, out of stylex-valid-styles.
Why? Because having fixable rules will serve as an education leverage for users to understand the pool of properties and values that are non-standard and its equivalent standard alternatives.  It will alleviate the extreme burden of checking for way too many rules with different purposes on a single eslint rule.

It also adds some cases for `require('@stylexjs/stylex')` to comply the test coverage threashold.

## Linked PR/Issues

Nothing so far, the issue in question is in progress (issue [1201](https://github.com/facebook/stylex/issues/1201))

## Additional Context

Tests for the rule have also been ported from one test file to a new one

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-no-nonstandard-styles-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-no-nonstandard-styles-test.js
@@ -1,0 +1,874 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+jest.disableAutomock();
+
+const { RuleTester: ESLintTester } = require('eslint');
+const rule = require('../src/stylex-no-nonstandard-styles');
+
+const eslintTester = new ESLintTester({
+  parser: require.resolve('hermes-eslint'),
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+});
+
+eslintTester.run('stylex-no-nonstandard-styles', rule.default, {
+  valid: [
+    `
+      import * as stylex from '@stylexjs/stylex';
+      const styles = stylex.create({
+        validStyle: {
+          marginInlineStart: "10px",
+          marginInlineEnd: "5px",
+          marginInline: "15px",
+          marginBlock: "20px",
+          paddingInlineStart: "8px",
+          paddingInlineEnd: "12px",
+          paddingInline: "10px",
+          paddingBlock: "16px",
+        },
+      });
+    `,
+    `
+      const stylex = require('@stylexjs/stylex');
+      
+      const styles = stylex.create({
+        validStyle: {
+          marginInlineStart: "10px",
+          marginInlineEnd: "5px",
+          marginInline: "15px",
+          marginBlock: "20px",
+          paddingInlineStart: "8px",
+          paddingInlineEnd: "12px",
+          paddingInline: "10px",
+          paddingBlock: "16px",
+        },
+      });
+    `,
+    `
+      import * as stylex from '@stylexjs/stylex';
+      const start = 'start';
+      const grayscale = 'grayscale';
+      const styles = stylex.create({
+        default: {
+          textAlign: start,
+          MozOsxFontSmoothing: grayscale,
+          WebkitFontSmoothing: 'antialiased',
+          transitionProperty: 'opacity, transform',
+          transitionDuration: '0.3s',
+          transitionTimingFunction: 'ease',
+        }
+      });
+    `,
+    `
+      import * as stylex from '@stylexjs/stylex';
+      const bounce = stylex.keyframes({
+        '0%': {
+          transform: 'translateY(0)',
+        },
+        '50%': {
+          transform: 'translateY(-10px)',
+        },
+        '100%': {
+          transform: 'translateY(0)',
+        },
+      });
+      const styles = stylex.create({
+        default: {
+          animationName: bounce,
+          animationDuration: '1s',
+          animationIterationCount: 'infinite',
+        }
+      });
+    `,
+    `
+      import * as stylex from '@stylexjs/stylex';
+      const styles = stylex.create({
+        default: {
+          animationName: stylex.keyframes({
+            '0%': {
+              transform: 'translateY(0)',
+            },
+            '50%': {
+              transform: 'translateY(-10px)',
+            },
+            '100%': {
+              transform: 'translateY(0)',
+            },
+          }),
+          animationDuration: '1s',
+          animationIterationCount: 'infinite',
+        }
+      });
+    `,
+    `
+      import * as stylex from '@stylexjs/stylex';
+      const bounce = stylex.keyframes({
+        '0%': {
+          transform: 'translateY(0)',
+        },
+        '50%': {
+          transform: 'translateY(-10px)',
+        },
+        '100%': {
+          transform: 'translateY(0)',
+        },
+      });
+      const shimmy = stylex.keyframes({
+        '0%': {
+          backgroundPosition: '-468px 0',
+        },
+        '100%': {
+          backgroundPosition: '468px 0',
+        },
+      });
+      const styles = stylex.create({
+        default: {
+          animationName: \`\${bounce}, \${shimmy}\`,
+          animationDuration: '1s',
+          animationIterationCount: 'infinite',
+        }
+      });
+    `,
+    // test for positive numbers
+    "import * as stylex from '@stylexjs/stylex'; stylex.create({default: {marginInlineStart: 5}});",
+    // test for literals as namespaces
+    'import * as stylex from \'@stylexjs/stylex\'; stylex.create({"default-1": {marginInlineStart: 5}});',
+    'import * as stylex from \'@stylexjs/stylex\'; stylex.create({["default-1"]: {marginInlineStart: 5}});',
+    // test for numbers as namespaces
+    "import * as stylex from '@stylexjs/stylex'; stylex.create({0: {marginInlineStart: 5}});",
+    // test for computed numbers as namespaces
+    "import * as stylex from '@stylexjs/stylex'; stylex.create({[0]: {marginInlineStart: 5}});",
+    // test for negative values.
+    "import * as stylex from '@stylexjs/stylex'; stylex.create({default: {marginInlineStart: -5}});",
+    // test for unitless length value 0
+    "import * as stylex from '@stylexjs/stylex'; stylex.create({default: {margin: 0}});",
+    "import * as stylex from '@stylexjs/stylex'; stylex.create({default: {padding: '0'}});",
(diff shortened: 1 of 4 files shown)


packages/@stylexjs/eslint-plugin/__tests__/stylex-no-nonstandard-styles-test.js
```diff
@@ -0,0 +1,874 @@
...
+         marginInlineEnd: \`\${Math.floor(x)}px\`,
+         paddingInlineStart: Math.ceil(-x),
+         paddingInlineEnd: Math.round(x / 2),
+       },
+     })`,
+    // test for Search
+    `import * as stylex from '@stylexjs/stylex';
+     stylex.create({
+       default: {
+         'WebkitAppearance': 'textfield',
```
mellyeliu: do these webkit tests belong here?

packages/@stylexjs/eslint-plugin/__tests__/stylex-no-nonstandard-styles-test.js
```diff
@@ -0,0 +1,874 @@
...
+        fieldSizing: 'content',
+      },
+    })`,
+    // test for stylex create vars tokens
+    `
+    import * as stylex from '@stylexjs/stylex';
+    import {TextTypeTokens as TextType, ColorTokens} from 'DspSharedTextTokens.stylex';
+    stylex.create({
+      root: {
+        fontSize: TextType.fontSize,
```
mellyeliu: same with these `defineVars` tests

packages/@stylexjs/eslint-plugin/__tests__/stylex-no-nonstandard-styles-test.js
```diff
@@ -0,0 +1,874 @@
...
+    import { tokens } from 'tokens.stylex';
+    stylex.create({
+      root: (position) => ({
+        [tokens.position]: \`\${position}px\`,
+      })
+    })
+    `,
+    // test importing vars from paths including code file extension
+    `
+    import * as stylex from '@stylexjs/stylex';
```
mellyeliu: and these as well

packages/@stylexjs/eslint-plugin/__tests__/stylex-no-nonstandard-styles-test.js
```diff
@@ -0,0 +1,874 @@
...
+    `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = {default: {width: '30pt'}};
+        stylex.create(styles);
+      `,
+    `import * as stylex from '@stylexjs/stylex';
+    import { FOO } from 'foo';
+     stylex.create({
+       default: {
+         scrollMarginTop: FOO + 5,
```
mellyeliu: this also is specific to the `valid-styles` test, not this rule here

mellyeliu (changes requested):
Thanks so much for taking a stab at this! 💯 

I think we should keep this rule scoped to handling just the nonstandard properties and values we support internally. Adding in checks for other scenarios like var imports with `.stylex.js` muddies the "educational" purpose of keeping rules isolated. Ideally we can check for all the nonstandard properties like `marginStart`, `borderVerticalWidth`, `float: start`, and keep everything else within `valid-styles` for now.

These nonstandard properties are pretty much specific to internal use as well, so we can probably keep this off by default.

mellyeliu:
Also, when adding new rules, we'll need to add it to the `packages/eslint-plugin/src/index.js` file as well, see https://github.com/facebook/stylex/pull/966 as an example!

packages/@stylexjs/eslint-plugin/__tests__/stylex-no-nonstandard-styles-test.js
jcperez-ch: yikes, maybe not, I should remove this

jcperez-ch:
@mellyeliu  I will address these comments, thanks for the review, I'll leave the PR open, knowing that is not ready yet.

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -619,6 +635,74 @@ eslintTester.run('stylex-valid-styles', rule.default, {
       },
     });
     `,
+    // tests what is covered by stylex-no-nonstandard-styles
```
mellyeliu: do we need this block here? 

thinking this could be confusing as these are still invalid styles, even if they're not tested by this rule anymore.

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -52,6 +52,22 @@ eslintTester.run('stylex-valid-styles', rule.default, {
         },
       });
     `,
+    `
+      const stylex = require('@stylexjs/stylex');
+      
+      const styles = stylex.create({
+        validStyle: {
+          marginInlineStart: "10px",
```
mellyeliu, line 60: also do we need this? seems like a repeat of the block above with commonjs

packages/@stylexjs/eslint-plugin/src/stylex-no-nonstandard-styles.js
```diff
@@ -0,0 +1,577 @@
...
+    const dynamicStyleVariables = new Set<string>();
+
+    type PropLimits = {
+      [string]: {
+        limit: null | string | number | Array<string | number>,
+        reason: string,
+      },
+    };
+
+    const legacyReason =
```
mellyeliu: do we use this anywhere? I think this belongs in a separate file because ideally we want to keep this rule for the nonstandard RN properties we adopted (invalid CSS), instead of properties we don't support for legacy reasons (properties that are valid CSS that we want to eventually support)

packages/@stylexjs/eslint-plugin/src/stylex-no-nonstandard-styles.js
```diff
@@ -0,0 +1,577 @@
...
+    function isStylexCreateDeclaration(node: Node) {
+      return (
+        node &&
+        node.type === 'CallExpression' &&
+        isStylexCreateCallee(node.callee) &&
+        node.arguments.length === 1
+      );
+    }
+
+    function checkStyleProperty(
```
mellyeliu, line 155: a good chunk of this logic is duplicated with the `valid-styles` rule. Can we try to keep what we can shared, then fork out any separate logic only?

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
jcperez-ch: One thing I realized is, that if I don't add certain rules that go beyond in the code: (like deconstructing the import or using `require` or passing propLimits, the coverage fails to reach the 75% of lines covered. For some cases where it's invalid, we can add some comments, but yeah, that is the reason I added those tests.

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
jcperez-ch: Same as above, this helps to increase the coverage percentage

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
mellyeliu: makes sense, just saw the PR description!

let's see what the coverage looks like when we simplify the code in `no-nonstandard-styles`

packages/@stylexjs/eslint-plugin/src/stylex-no-nonstandard-styles.js
jcperez-ch: You're right, this is not used at all in this rule, taking it out

packages/@stylexjs/eslint-plugin/src/stylex-no-nonstandard-styles.js
```diff
@@ -0,0 +1,467 @@
...
+export type RuleResponse = void | {
+  message: string,
+  distance?: number,
+  suggest?: {
+    fix: Rule.ReportFixer,
+    desc: string,
+  },
+};
+
+const stylexValidStyles = {
```
mellyeliu: rename? 
```suggestion
const stylexNoNonstandardStyles = {
```

packages/@stylexjs/eslint-plugin/src/stylex-no-nonstandard-styles.js
```diff
@@ -0,0 +1,467 @@
...
+      const key =
+        propName ??
+        (styleKey.type === 'Identifier' ? styleKey.name : styleKey.value);
+      if (typeof key !== 'string') {
+        return;
+      }
+
+      const ruleChecker = CSSPropertiesWithOverrides[key];
+      if (ruleChecker == null) {
+        const closestKey = CSSPropertyKeys.find((cssProp) => {
```
mellyeliu: I also don't think we need to check for this logic, because everything should be a deterministic autofix mapping, we wouldn't be suggesting a closest match.

mellyeliu (commented):
Just one last round of fixes 😁 We shouldn't need the blocks pulling in `CSSProperties` or suggesting a `closestKey`, since everything will be a deterministic mapping to its standard equivalent in the case of the ~20 or so properties that trigger this rule.

packages/@stylexjs/eslint-plugin/src/stylex-no-nonstandard-styles.js
```diff
@@ -0,0 +1,429 @@
...
+            as: string,
+          },
+      >,
+    };
+    const {
+      validImports: importsToLookFor = ['stylex', '@stylexjs/stylex'],
+    }: Schema = context.options[0] || {};
+    const importTracker = createImportTracker(importsToLookFor);
+    const variables = new Map<string, Expression | 'ARG'>();
+    const dynamicStyleVariables = new Set<string>();
```
mellyeliu, line 110: I don't think we populated this, with this implementation we're skipping dynamic styles

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
mellyeliu: I do still think we should remove this block, is the test coverage complaining on this? this code isn't handled by valid-styles anymore and it's confusing to label it as "valid" because it's technically not valid, so I would want to remove it. The tests sometimes function as a "this is correct code" and "this is not correct code" check for us + users

packages/@stylexjs/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -491,62 +491,47 @@ const stylexValidStyles = {
                 ? convertToStandardProperties[style.key.value]
```
mellyeliu: and this is the map that we moved to `no-nonstandard-styles` right? we shouldn't need it in `valid-styles` anymore

PR #1254, opened 2025-09-28
[fix] Enforce ordering for .when selectors lower than regular pseudos

### LINKED ISSUE #1251: `when` does not appear to respect user-authored style order (e.g. :hover/:active)
### Describe the issue

Given the following code:

```tsx
import * as React from "react";

export function App(): React.ReactNode {
  return (
    <div {...stylex.props(styles.center)}>
      <div {...stylex.props(styles.column)}>
        {/* Button that does not use when */}
        {/* This works correctly */}
        <button {...stylex.props(styles.buttonWithoutWhen)}>Button</button>
        {/* Button that uses when */}
        {/* This does not work correctly */}
        <div {...stylex.props(stylex.defaultMarker())}>
          <button {...stylex.props(styles.buttonWithWhen)}>Button</button>
        </div>
      </div>
    </div>
  );
}

  // Misc.
  center: {
    alignItems: "center",
    display: "flex",
    justifyContent: "center",
    minBlockSize: "100vh",
    paddingBlock: 128,
    paddingInline: 16,
  },
  column: { display: "flex", flexDirection: "column", gap: 16 },

  // Button
  buttonWithoutWhen: {
    alignItems: "center",
    backgroundColor: {
      // This works correctly
      default: "red",
      ":hover": "orange",
      ":active": "blue",
    },
    blockSize: 36,
    borderRadius: 1e3,
    display: "flex",
    justifyContent: "center",
    paddingInline: 18,
  },
  buttonWithWhen: {
    alignItems: "center",
    backgroundColor: {
      // This does not work correctly
      default: "red",
      [stylex.when.ancestor(":hover")]: "orange",
      [stylex.when.ancestor(":active")]: "blue",
    },
    blockSize: 36,
    borderRadius: 1e3,
    display: "flex",
    justifyContent: "center",
    paddingInline: 18,
  },
});
```

It emits the following CSS (I ran it through Prettier)

```css
@layer stylex {
  @layer priority1, priority2, priority3;
  @layer priority1 {
    .borderRadius-xzl92wl { border-radius: 1000px; }
    .gap-xou54vl { gap: 16px; }
    .paddingBlock-x1qvrp8 { padding-block: 128px; }
    .paddingInline-x1hr3lfm { padding-inline: 16px; }
    .paddingInline-x12j899k { padding-inline: 18px; }
  }
  @layer priority2 {
    .alignItems-x6s0dn4 { align-items: center; }
    .backgroundColor-xrkmrrc { background-color: red; }
    .display-x78zum5 { display: flex; }
    .flexDirection-xdt5ytf { flex-direction: column; }
    .justifyContent-xl56j7k { justify-content: center; }
    .backgroundColor-x17h08bh:where(.x-default-marker:active *) { background-color: blue; }
    .backgroundColor-x1l9x0em:where(.x-default-marker:hover *) { background-color: orange; }
    .backgroundColor-xeirn3a:hover { background-color: orange; }
    .backgroundColor-x12plbpf:active { background-color: blue; }
  }
  @layer priority3 {
    .height-xc9qbxq { height: 36px; }
    .minHeight-xg6iff7 { min-height: 100vh; }
  }
}
```

And here's a quick video to show the issue:

https://github.com/user-attachments/assets/43bfd7a6-49f5-42b1-a97c-04cc1215b44a

I think what's happening is the order of these lines does not respect the order in which I specify the `when(:hover)` and `when(:active)` styles:

```css
    .backgroundColor-xeirn3a:hover { background-color: orange; }
    .backgroundColor-x12plbpf:active { background-color: blue; }
    /* ^ Correct order; respects user-authored style order */

    .backgroundColor-x17h08bh:where(.x-default-marker:active *) { background-color: blue; }
    .backgroundColor-x1l9x0em:where(.x-default-marker:hover *) { background-color: orange; }
    /* ^ Incorrect order; does not respect user-authored style order */
```

Is this a bug with `when`?

Here's a simple repro if useful: https://github.com/zaydek/repro-stylex-when. This was initialized using https://github.com/facebook/stylex/discussions/825#discussioncomment-11720866 for reference.

Thanks.

### Expected behavior

Button using `when` should be `blue` on `:active`.

### Steps to reproduce

- Install https://github.com/zaydek/repro-stylex-when
- npm run dev

### Test case

_No response_

### Additional comments

_No response_

nmn:

The new `.when.*` utility based selectors are currently implemented to all have the same low priority. This correctly ensures that `:when` always loses to pseudo classes on the element itself, however, when using multiple `.when` selectors, their order is currently alphabetical.

This PR fixes that to ensure that `.when` selectors have the same ordering as other pseudo classes. (e.g. `:hover` will lose to `:active`)

## Linked PR/Issues

Fixes #1251

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/shared/utils/property-priorities.js
+++ b/packages/@stylexjs/babel-plugin/src/shared/utils/property-priorities.js
@@ -715,6 +715,15 @@ export const AT_RULE_PRIORITIES: $ReadOnly<AtRulePriorities> = {
 
 export const PSEUDO_ELEMENT_PRIORITY: number = 5000;
 
+const ANCESTOR_SELECTOR = /^:where\(\.[0-9a-zA-Z_-]+(:[a-zA-Z-]+)\s+\*\)$/;
+const DESCENDANT_SELECTOR = /^:where\(:has\(\.[0-9a-zA-Z_-]+(:[a-zA-Z-]+)\)\)$/;
+const SIBLING_BEFORE_SELECTOR =
+  /^:where\(\.[0-9a-zA-Z_-]+(:[a-zA-Z-]+)\s+~\s+\*\)$/;
+const SIBLING_AFTER_SELECTOR =
+  /^:where\(:has\(~\s\.[0-9a-zA-Z_-]+(:[a-zA-Z-]+)\)\)$/;
+const ANY_SIBLING_SELECTOR =
+  /^:where\(\.[0-9a-zA-Z_-]+(:[a-zA-Z-]+)\s+~\s+\*,\s+:has\(~\s\.[0-9a-zA-Z_-]+(:[a-zA-Z-]+)\)\)$/;
+
 export default function getPriority(key: string): number {
   if (key.startsWith('--')) {
     return 1;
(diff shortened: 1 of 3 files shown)


packages/@stylexjs/babel-plugin/src/shared/utils/property-priorities.js
```diff
@@ -736,6 +745,48 @@ export default function getPriority(key: string): number {
...
+    const [_all, pseudo] = siblingBeforeMatch;
+    const basePseudoPriority = PSEUDO_CLASS_PRIORITIES[pseudo] ?? 40;
+    return 20 + basePseudoPriority / 100;
+  }
+
+  const siblingAfterMatch = SIBLING_AFTER_SELECTOR.exec(key);
+  if (siblingAfterMatch) {
+    const [_all, pseudo] = siblingAfterMatch;
+    const basePseudoPriority = PSEUDO_CLASS_PRIORITIES[pseudo] ?? 40;
+    return 30 + basePseudoPriority / 100;
```
mellyeliu, line 775: what's the rationale behind having sibling after > sibling before > descendant > ancestor?

mellyeliu (approved):
Let's start with this for now as a fix, we can refine/add sorting in follow up

Issue #1257, opened 2025-09-29
[eslint-plugin] Create `no-lookahead-selector` ESLint rule


mellyeliu:
We probably want to introduce a lint rule to conditionally allow lookahead selectors (for `siblingAfter`, `ancestor`, and `anySibling`. Users can enable or disable the rule based on their supported browsers.

See browser support for `:has()` selector: https://developer.mozilla.org/en-US/docs/Web/CSS/:has

We would need this for internal adoption of descendant selectors

PR #1258, opened 2025-10-01
[eslint-plugin] create `no-lookahead-selectors` rule

mellyeliu:
Add ESLint rule to disallow descendant and sibling selectors with limited browser support
- `no-lookahead-selectors` detects usage of `stylex.when.anySibling`, `stylex.when.descendant`, and `stylex.when.nextSibling`
- These selectors rely on CSS `:has()` which has limited browser support
- See: https://caniuse.com/css-has

We'll need this for internal adoption because we don't support these browsers on all surfaces. Teams can toggle the rule on and off by product

Fixes https://github.com/facebook/stylex/issues/1257, will add docs in follow up

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-no-lookahead-selectors-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-no-lookahead-selectors-test.js
@@ -240,7 +240,7 @@ eslintTester.run('stylex-no-lookahead-selectors', rule.default, {
           button: {
             color: {
               default: 'black',
-              [stylex.when.descendant(':dark')]: 'red',
+              [stylex.when.descendant(':hover')]: 'red',
             },
           },
         });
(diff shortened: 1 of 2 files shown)


packages/@stylexjs/eslint-plugin/__tests__/stylex-no-lookahead-selectors-test.js
```diff
@@ -0,0 +1,324 @@
...
+          },
+        });
+      `,
+    },
+  ],
+  invalid: [
+    {
+      // stylex.when.anySibling usage
+      code: `
+        import * as stylex from '@stylexjs/stylex';
```
Jta26, line 121: Do you need to add an invalid test here in case someone destructures the stylex import? Or is this too burdensome to support in the rule?

example: `import {when} from '@stylexjs/stylex'`

PR #1270, opened 2025-10-08
[eslint-plugin] handle descendant selector styles in `valid-styles`

mellyeliu:
We'll need to allowlist `stylex.when` keys within `valid-styles` to avoid a computed keys error. Similar to what was done for `.stylex.js` variables (https://github.com/facebook/stylex/pull/393) and variables coming in from `.transformed` (https://github.com/facebook/stylex/pull/1208)


packages/@stylexjs/eslint-plugin/src/stylex-valid-styles.js
```diff
@@ -417,32 +418,79 @@ const stylexValidStyles = {
         ) {
           return undefined;
```
mellyeliu, line 419: this early return also results in missing validation for computed keys from `.stylex.js` exports, to address at some point

PR #1273, opened 2025-10-11
[eslint-plugin] add a config option to change the ordering of properties in the sort-keys rule

yjoer:

This PR continues the work from #1143, which appeared to have stalled. The changes introduce a configuration option to guard the new ordering rules, ensuring that the default ordering behavior remains ascending. This approach preserves compatibility with existing projects for users who do not enable the new configuration. To avoid disrupting existing tests, new test cases for [clean-order](https://github.com/kutsan/stylelint-config-clean-order) have been added instead of modifying the existing ones.

Additionally, adjustments have been made to ensure that logical properties appear before physical properties. Any discrepancies against the upstream have been marked with // after the corresponding property.

Support for [recess-order](https://github.com/stormwarning/stylelint-config-recess-order) has also been added. Among the configs listed in [stylelint-order](https://github.com/hudochenkov/stylelint-order?tab=readme-ov-file#example-configs), only clean-order and recess-order have at least a commit within the past year and exceed [100k weekly downloads](https://npmtrends.com/stylelint-config-clean-order-vs-stylelint-config-recess-order). Since recess-order shares the same code flow as clean-order, only a few additional test cases were added.

## Linked PR/Issues

Fixes #1143

## Pre-flight checklist

- [x ] I have read the contributing guidelines
- [x ] Performed a self-review of my code

### DIFF
--- a/packages/@stylexjs/eslint-plugin/src/utils/getPropertyPriorityAndType.js
+++ b/packages/@stylexjs/eslint-plugin/src/utils/getPropertyPriorityAndType.js
@@ -1011,8 +1011,8 @@ const RECESS_ORDER_PRIORITIES = [
 
 const ORDER_PRIORITIES = {
   default: undefined,
-  'clean': CLEAN_ORDER_PRIORITIES,
-  'recess': RECESS_ORDER_PRIORITIES,
+  clean: CLEAN_ORDER_PRIORITIES,
+  recess: RECESS_ORDER_PRIORITIES,
 };
 
 export default function getPropertyPriorityAndType(


packages/@stylexjs/eslint-plugin/src/utils/getPropertyPriorityAndType.js
```diff
@@ -11,18 +11,1024 @@
...
+
+const ORDER_PRIORITIES = {
+  asc: undefined,
+  'clean-order': CLEAN_ORDER_PRIORITIES,
+  'recess-order': RECESS_ORDER_PRIORITIES,
+};
 
 export default function getPropertyPriorityAndType(
   key: string,
+  order: 'asc' | 'clean-order' | 'recess-order',
```
mellyeliu: Let's call this `default` or `alphanumeric`. A bit difficult to get right as it's not purely lexicographic. 

```suggestion
  order: 'ascending' | 'clean-order' | 'recess-order',
```

yjoer: Let's call it `default` to avoid confusion, since it's not entirely in alphabetical order.

nmn: Perhaps: `"default" | "clean" | "recess"` is good? The `-order` suffix is redundant IMO. The only hesitation I have is calling anything "clean".

yjoer: I removed the suffixes. I believe your concern can be addressed by clarifying that "clean" refers to "stylelint-config-clean-order" in the docs.

mellyeliu (approved):
Thanks for making this change! Looks great, we can iterate on the naming + docs in a later patch

PR #1274, opened 2025-10-13
[babel-plugin] refine createTheme and defineVars priorities

mellyeliu:
Pushed up a bad revision in https://github.com/facebook/stylex/pull/1272, re-tweaking `priorityForAtRule`

Also simplifying some of the logic in `createTheme` as the `default` at-rule behaviour was previously forked

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/shared/stylex-create-theme.js
+++ b/packages/@stylexjs/babel-plugin/src/shared/stylex-create-theme.js
@@ -58,7 +58,6 @@ export default function styleXCreateTheme(
     .map((atRule) => wrapWithAtRules(rulesByAtRule[atRule].join(''), atRule))
     .join('');
 
-  // Create a class name hash
   const overrideClassName = classNamePrefix + createHash(atRulesStringForHash);
 
   const stylesToInject: { [string]: InjectableStyle } = {};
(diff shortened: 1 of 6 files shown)


packages/@stylexjs/babel-plugin/src/shared/stylex-define-vars.js
```diff
@@ -125,7 +125,7 @@ function constructCssVariablesString(
     result[varGroupHash + suffix] = {
       ltr,
       rtl: null,
-      priority: 0.1 + priorityForAtRule(atRule) * 0.1,
+      priority: priorityForAtRule(atRule) * 0.1,
```
nmn:
```suggestion
      priority: priorityForAtRule(atRule) / 10,
```

packages/@stylexjs/babel-plugin/src/shared/stylex-create-theme.js
```diff
@@ -67,19 +66,14 @@ export default function styleXCreateTheme(
...
-        rtl: null,
-      };
-    } else {
-      stylesToInject[overrideClassName + '-' + createHash(atRule)] = {
-        ltr: wrapWithAtRules(rule, atRule),
-        priority: 0.5 + 0.1 * priorityForAtRule(atRule),
-        rtl: null,
-      };
-    }
+    const priority = 0.4 + 0.1 * priorityForAtRule(atRule);
```
nmn:
```suggestion
    const priority = 0.4 + priorityForAtRule(atRule) / 10;
```

PR #1276, opened 2025-10-13
[types] Flow types fix for stylex.when.*

nmn:

Using computed keys within styles conditional styles can have type errors with Flow.

This attempts to fix it using the `StringPrefix` type.

### DIFF
--- a/packages/@stylexjs/stylex/src/types/StyleXTypes.js
+++ b/packages/@stylexjs/stylex/src/types/StyleXTypes.js
@@ -145,7 +145,7 @@ export type MapNamespaces<+S: { +[string]: mixed }> = $ReadOnly<{
     ? (...args: Args) => $ReadOnly<[MapNamespace<Obj>, InlineStyles]>
     : MapNamespace<S[Key]>,
 }>;
-export type StyleX$Create = <S: { +[string]: { ... } }>(
+export type StyleX$Create = <const S: { +[string]: { ... } }>(
   styles: S,
 ) => MapNamespaces<S>;
 
@@ -226,7 +226,7 @@ export type StyleX$DefineVars = <DefaultTokens: TTokens, ID: string = string>(
 ) => VarGroup<FlattenTokens<DefaultTokens>, ID>;
 
 export type StyleX$DefineConsts = <
-  DefaultTokens: { +[string]: number | string },
+  const DefaultTokens: { +[string]: number | string },
 >(
   tokens: DefaultTokens,
 ) => DefaultTokens;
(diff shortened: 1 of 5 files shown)


mellyeliu:
This requires a bit more investigation. [Flow](https://flow.org/try/#1N4Igxg9gdgZglgcxALlAIwIZoKYBsD6uEEAztvhgE6UYCe+JADpdhgCYowa5kA0I2KAFcAtiRQAXSkOz9sADwxgJ+NPTbYuQ3BMnTZA+Y2yU4IwRO4A6SFBIrGVDGM7c+h46fNRLuKxJIGWh8MeT0ZfhYlCStpHzNsFBAMIQkIEQwJODAQfiEyfBE4eWw2fDgofDBMsAALfAA3KjgsXGxxZC4eAw0G-GhcWn9aY3wWZldu-g1mbGqJUoBaCRHEzrcDEgBrbAk62kXhXFxJ923d-cPRHEpTgyEoMDaqZdW7vKgoOfaSKgOKpqmDA+d4gB5fMA-P6LCCMLLQbiLOoYCqgh6-GDYRYIXYLSgkRZkCR4jpddwPfJLZjpOBkUEKTwJEJ+DAkMiUFSwkyZCC3dbdAC+-EgGiSK2MAAIAEJEMBbAA8ADUAHwSgC8EuAAB0oBKJRotDoAPzICWK3g6vUAbXspigCAAuqbzZaJVZ3TqBQBuHU6jRPKjYCWAiX2WhteSm7W6iUAd1qgijrr1wMh9l5poAFIwyEI2BATaGpBUEABKdWqgDKxftAAUWPB5PKAOTIeMmbDN5UWmNC11gKILLNMOZRiUAagylAQFULMogcvlwhEN27bvdEoF5bVqqKJTYPe9vrsKwjNkH2Ez0b1U5nUCTMb1BpSOlNAAYe3rrWGz+2oFZU3aNJKEzVtaggBoTGbUsnQlD9XT7KAtx9KBchASD8TgaAkgaN8rAAJgADgAdisN8QAFIA) isn't too happy with this and our other use case (`defineVars` keys within `create` calls) generates the same issue. 

The Flow typechecker only seems to accept placeholder string literals as definitively not conflicting with the `default` key. Returning a refined type like `StringPrefix` still triggers the computed key collision error. Reaching out to the Flow team to get some support.

mellyeliu:
Reached out to Flow team, they're adding a fix for `StringPrefix` and `StringSuffix` to be used in object initializers with non-overlapping keys

packages/@stylexjs/stylex/src/types/StyleXTypes.js
```diff
@@ -254,3 +254,11 @@ export type StyleX$CreateTheme = <
   baseTokens: BaseTokens,
   overrides: OverridesForTokenType<TokensFromVarGroup<BaseTokens>>,
 ) => Theme<BaseTokens, ID>;
+
+export type StyleX$When = {
+  ancestor: (_pseudo?: StringPrefix<':'>) => ':where-ancestor',
```
mellyeliu, line 259: This is what I did internally, but I think it only works because Flow's missing checks for conflicts between computed keys themselves, and only checks string literal and computed keys mixed together

PR #1284, opened 2025-10-17
[eslint-plugin] Fix `valid-styles` handling of `when` + vars/consts

dwei-figma:

`valid-styles` is misreporting vars/consts used as values for a style variant declared with `stylex.when` to be invalid. 

The cause seems to be a short circuit that we added for the case when the key is `stylex.when` based, which disables later logic that accepts vars/consts alternatively to passing the checker for literal/resolved values. I inspected the code that comes later and it seems like everything should work as long as `propName` is passed. The latter is implicitly guaranteed because it seems like attempts to use `stylex.when` in the outer layer will fail early with the "Keys must be strings" error.

## Additional Context

Added a few test cases to validate the behavior more thoroughly.

Also I presumed that we are fine with `when` not being allowed with the legacy outer syntax - but lemme know if not and I can create a follow-up.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -2514,7 +2514,7 @@ revert`,
         },
       ],
     },
-    // test for trying to use `stylex.when` as outer key which is not supported
+    // test for trying to use `stylex.when` as outer key when `allowOuterPseudoAndMedia` is false
     {
       code: `
         import * as stylex from '@stylexjs/stylex';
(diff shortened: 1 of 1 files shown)


packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -2459,5 +2514,31 @@ revert`,
         },
       ],
     },
+    // test for trying to use `stylex.when` as outer key which is not supported
```
mellyeliu: Let's set `allowOuterPseudoAndMedia` to false for this test for the time being for clarity

mellyeliu (approved):
Thanks for catching this early return from https://github.com/facebook/stylex/pull/1270, appreciate the cleanup! 

The outer pseudos pattern is discouraged to resolve styles correctly but not outright banned and hasn't been thoroughly tested with the new APIs. If we decide to disallow `stylex.when` selectors as outer keys, we likely want to rename `allowOuterPseudoAndMedia` config to avoid user confusion. But it's not a priority and worth thinking through a bit more. The same thing applies to shareable media queries defined with `defineConsts`.

PR #1287, opened 2025-10-18
[eslint-plugin] order pseudo-classes and stylex.when selectors according to the priorities

### LINKED ISSUE #1285: [eslint-plugin] Have `sort-keys` respect priorities
We enforces a canonical ordering of pseudo-classes and `.when` selectors  rather than respecting user authored order. Defining negations of pseudo classes like `:hover:not(:focus)` as was done for media queries would be too costly to CSS size because of combinatorial explosion. 

Instead, the `sort-keys` plugin + `getPropertyPriorityAndType` util should sort this by the preset priorities so the behaviour is intuitive to users: 
- `stylex.when` selectors
- Other pseudos as defined in `property-priorities`

yjoer:

I extracted property-priorities into a public package because the existing package doesn't seem to bundle its dependencies. Keeping the shared package private would result in resolution failures in downstream projects since the code wouldn't be accessible. To support stylex.when selectors, the first call argument is extracted as the property name, and the subsequent processing follows the same flow as for identifier and literal properties.

## Linked PR/Issues

Fixes #1285

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
@@ -1312,7 +1312,8 @@ eslintTester.run('stylex-sort-keys', rule.default, {
       `,
       errors: [
         {
-          message: 'StyleX property key ":focus" should be above ":active"',
+          message:
+            'StyleX property key ":when:descendant:focus" should be above ":when:siblingAfter:active"',
         },
         {
           message: 'StyleX property key "display" should be above "width"',
@@ -1326,7 +1327,7 @@ eslintTester.run('stylex-sort-keys', rule.default, {
         base: {
           display: 'flex',
           width: {
-            [when.siblingAfter(':active')]: 30,
+            [stylex.when.siblingAfter(':active')]: 30,
             [when.descendant(':focus')]: 20,
           },
         },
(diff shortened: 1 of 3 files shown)


yjoer:
On top of that, I have wired up the shorthand before longhand and logical properties before physical properties rules in `property-priorities` to the default order. It was alphabetical previously for these non-at-rules, non-pseudo-elements, and non-pseudo-classes properties. It is now arranged according to those four categories and falls back to alphabetical order for properties having the same priority.

For stylex.when selectors, I have extracted the relation (ancestor, descendant, anySibling, siblingBefore, siblingAfter) within the callee in addition to the call argument to help retrieve a more precise priority. If the relation is unknown, i.e., owing to a dynamically specified property, it will fall back to alphabetical order.

mellyeliu:
Overall this looks great, and an important change with `stylex.when` priorities being more on the arbitrary side. Let me think a bit more about what we want to do with the shared package. It's a bit confusing to have half of the old shared package within the `babel-plugin` alongside a top-level public shared package, perhaps we can consider moving it all out again.

yjoer:
Wouldn't it be more confusing if we moved the shared package primarily serving visitors within the babel-plugin out? I assume it has more clarity when the shared package contains code that's used across multiple plugins.

I think it's more compelling to move the package out if the functions there have the potential to be shared across multiple plugins. If that's the case, it probably wouldn't have been placed within the babel-plugin to begin with, though I'm not certain. Otherwise, we could consider renaming Babel's shared package to something like visitor-utils.

mellyeliu:
@yjoer The `shared` package was originally a standalone public package, it was moved inside the babel plugin for simplicity. The reason it was separate was to keep things decoupled from the babel-plugin in case we ever wanted to integrate with Babel alternatives. I'll review this PR soon!

mellyeliu (approved):
Overall this change looks fine. Thanks for making this change, we can revisit whether we want to move the rest of shared back into a public package later.

PR #1289, opened 2025-10-20
[babel-plugin] create new defineConsts file extension

mellyeliu:
Depending on the way users serve their CSS, you may want to declare a separate filepath for constants. For example, when using directory-based bundling, you'd need to include all constants files alongside each folder to have constants work globally. 

This PR adds a `.const` appended variant to the suffix defined in `themeFileExtension` (`.stylex.js` by default) so packagers can statically analyze which files contain constants. I considered making this configurable like `themeFileExtension`, but decided to keep this hardcoded for now for simplicity. 

As a follow up, we should update the `enforce-extension` lint rule to handle `defineConsts` under the new extension, with a config to allow users to hard enforce the `.stylex.const.js` extension (ie. no `defineConsts` in the default theme extension). The default behaviour will be that both `.stylex.js` and `.stylex.const.js` will be considered valid extensions for files declaring `defineConsts` exports.

Also contains a fix to `valid-styles` to respect user configured `themeFileExtension`

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/utils/state-manager.js
+++ b/packages/@stylexjs/babel-plugin/src/utils/state-manager.js
@@ -458,7 +458,7 @@ export default class StateManager {
     const themeFileExtension =
       this.options.unstable_moduleResolution?.themeFileExtension ?? '.stylex';
 
-    const constsFileExtension = `${themeFileExtension}.const}`;
+    const constsFileExtension = `${themeFileExtension}.const`;
 
     if (filename == null || this.options.unstable_moduleResolution == null) {
       return null;
@@ -528,7 +528,7 @@ export default class StateManager {
     const themeFileExtension =
       this.options.unstable_moduleResolution?.themeFileExtension ?? '.stylex';
 
-    const constsFileExtension = `${themeFileExtension}.const}`;
+    const constsFileExtension = `${themeFileExtension}.const`;
 
     const transformedVarsFileExtension = '.transformed';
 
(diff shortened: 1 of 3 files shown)


packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -458,15 +458,21 @@ export default class StateManager {
     const themeFileExtension =
       this.options.unstable_moduleResolution?.themeFileExtension ?? '.stylex';
 
-    if (
-      filename == null ||
-      !matchesFileSuffix(themeFileExtension)(filename) ||
-      this.options.unstable_moduleResolution == null
-    ) {
+    const constsFileExtension = `${themeFileExtension}.const}`;
```
ezzak: Incorrect ending curly brace here?

packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -521,14 +527,23 @@ export default class StateManager {
 
     const themeFileExtension =
       this.options.unstable_moduleResolution?.themeFileExtension ?? '.stylex';
+
+    const constsFileExtension = `${themeFileExtension}.const}`;
```
ezzak: Same?

mellyeliu:
Tested on `example-nextJS` for safety;

Compiler error with
```
import { tokens } from './ConstantTokens.stylex.constss';
```

Resolves correctly with
```
import { tokens } from './ConstantTokens.stylex.const';
```

PR #1296, opened 2025-10-24
Split disallowed properties (keys) in stylex into stylex-no-disallowed-properties.js

jcperez-ch:

We wan to split fixable rules related to disallowed properties in stylex onto a separate eslint rule, out of stylex-valid-styles.
Why? Because having fixable rules will serve as an education leverage for users to understand the pool of properties that are no valid in stylex.  It will alleviate the extreme burden of checking for way too many rules with different purposes on a single eslint rule. Similar to what we did with `no-nonstandard-styles` rule.

It also abstracts several repeated code in functions under /utils folder, including:
* `isStylexCreateDeclaration`
* `isStylexCreateCallee`

## Linked PR/Issues

Issue in progress (issue [1201](https://github.com/facebook/stylex/issues/1201)).
PR Related [1246](https://github.com/facebook/stylex/pull/1246)

## Additional Context

Tests for the rule have also been ported from one test file to a new one

## Pre-flight checklist


packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -2121,22 +2047,7 @@ revert`,
       ],
     },
     // test for disallowed raw CSS variable overrides
-    {
-      code: `
-        import * as stylex from '@stylexjs/stylex';
-        const styles = stylex.create({
-          foo: {
-            '--bar': '0',
```
mellyeliu, line 2129: hmm, we likely want to keep this in the original `valid-styles` rule, as CSS variable overrides aren't really a CSS property

PR #1301, opened 2025-10-27
[eslint-plugin] Add transform-media-query rule and soft validation config

### LINKED ISSUE #1200: [style-value-parser] add config for "soft" validation
Add a config to the try/catch to flag hard errors in enableMediaQueryOrder and fallback to pre-existing behaviour. Find a way to log or capture errors, possibly in new lint rule for the parser

Prakshal-Jain:
## What changed / motivation?

Introducing a new ESLint rule `stylex-validate-media-queries` that validates media query syntax in StyleX declarations. The motivation is to provide early feedback to developers about invalid media queries during development and flag when media query order cannot be enforced.

### Key Changes:
- Added `stylex-validate-media-queries` rule that checks media query syntax in `stylex.create()` and `stylex.createTheme()` calls
- Implemented a `softValidation` option in the babel plugin that allows media query validation to be handled by ESLint instead of failing at build time
- The rule validates:
  - Media query syntax and structure
  - Valid media types (screen, print, all, etc.)
  - Valid media features (width, height, orientation, etc.)
  - Proper use of logical operators (and, or, not)
  - Range syntax and value formats
- **Developer Experience**: Provides clear, actionable error messages with specific guidance on what's wrong and how to fix it

## Linked PR/Issues
Fixes #1200

### Tests
- Added test suite covering valid and invalid media query patterns
- Tests include edge cases like nested conditions, range syntax, and various media features
- All existing tests continue to pass

### Configuration
The ESLint rule can be enabled in `.eslintrc`:
```json
{
  "rules": {
    "@stylexjs/stylex-validate-media-queries": "error"
  }
}
```

The babel plugin's soft validation mode can be enabled via:
```javascript
{
  plugins: [
    ['@stylexjs/babel-plugin', { 
      softValidation: ['mediaQueries'] 
    }]
  ]
}
```

### Breaking Changes
None - this is an opt-in feature that doesn't affect existing functionality.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-transform-media-queries-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-transform-media-queries-test.js
@@ -10,7 +10,7 @@
 jest.disableAutomock();
 
 const { RuleTester: ESLintTester } = require('eslint');
-const rule = require('../src/stylex-validate-media-queries');
+const rule = require('../src/stylex-transform-media-queries');
 
 const eslintTester = new ESLintTester({
   parser: require.resolve('hermes-eslint'),
@@ -20,7 +20,7 @@ const eslintTester = new ESLintTester({
   },
 });
 
-eslintTester.run('stylex-validate-media-queries', rule.default, {
+eslintTester.run('stylex-transform-media-queries', rule.default, {
   valid: [
     {
       code: `
@@ -38,6 +38,7 @@ eslintTester.run('stylex-validate-media-queries', rule.default, {
       const styles = stylex.create({
         main: {
           color: {
+            default: 'yellow',
             '@media (min-width: 768px)': 'blue',
             '@media (min-width: 1024px)': 'green',
           },
@@ -73,7 +74,8 @@ eslintTester.run('stylex-validate-media-queries', rule.default, {
     `,
       errors: [
         {
-          message: /Media query order may not be respected/,
+          message:
+            /Media query order may not be respected\. Invalid media query syntax or unsupported edge cases in style-value-parser\. Error:/,
         },
       ],
     },
--- a/packages/@stylexjs/eslint-plugin/src/index.js
+++ b/packages/@stylexjs/eslint-plugin/src/index.js
@@ -15,7 +15,7 @@ import noUnused from './stylex-no-unused';
 import sortKeys from './stylex-sort-keys';
 import validShorthands from './stylex-valid-shorthands';
 import validStyles from './stylex-valid-styles';
-import validateMediaQueries from './stylex-validate-media-queries';
+import transformMediaQueries from './stylex-transform-media-queries';
 
 const rules: {
   'enforce-extension': typeof enforceExtension,
@@ -26,7 +26,7 @@ const rules: {
   'sort-keys': typeof sortKeys,
   'valid-shorthands': typeof validShorthands,
   'valid-styles': typeof validStyles,
-  'validate-media-queries': typeof validateMediaQueries,
+  'transform-media-queries': typeof transformMediaQueries,
 } = {
   'enforce-extension': enforceExtension,
   'no-legacy-contextual-styles': noLegacyContextualStyles,
@@ -36,7 +36,7 @@ const rules: {
   'sort-keys': sortKeys,
   'valid-shorthands': validShorthands,
   'valid-styles': validStyles,
-  'validate-media-queries': validateMediaQueries,
+  'transform-media-queries': transformMediaQueries,
 };
 
 export { rules };
--- a/packages/@stylexjs/eslint-plugin/src/stylex-transform-media-queries.js
+++ b/packages/@stylexjs/eslint-plugin/src/stylex-transform-media-queries.js
@@ -14,12 +14,12 @@ import createImportTracker from './utils/createImportTracker';
 import { lastMediaQueryWinsTransform } from 'style-value-parser';
 /*:: import { Rule } from 'eslint'; */
 
-const stylexValidateMediaQueries = {
+const stylexTransformMediaQueries = {
   meta: {
     type: 'problem',
     docs: {
       description:
-        'Warn when media query syntax may not be correctly ordered due to parser limitations',
+        'Warn when media query syntax order is not properly enforced due to invalid media query syntax or parser limitations',
       recommended: false,
       url: 'https://github.com/facebook/stylex/tree/main/packages/@stylexjs/eslint-plugin',
     },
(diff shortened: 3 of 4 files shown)


packages/@stylexjs/eslint-plugin/src/stylex-validate-media-queries.js
```diff
@@ -0,0 +1,239 @@
...
+'use strict';
+
+import type { CallExpression, Node, Property, ObjectExpression } from 'estree';
+import createImportTracker from './utils/createImportTracker';
+/*:: import { Rule } from 'eslint'; */
+
+let lastMediaQueryWinsTransform = null;
+try {
+  // $FlowFixMe Dynamic import for style-value-parser since it might not be available in all environments
+  const styleValueParser = require('style-value-parser');
```
Prakshal-Jain: @mellyeliu Not sure if the import statement is safe here?

mellyeliu: hmm why cjs here? let's use esm `import` instead

packages/@stylexjs/eslint-plugin/src/stylex-validate-media-queries.js
```diff
@@ -0,0 +1,227 @@
...
+import createImportTracker from './utils/createImportTracker';
+import { lastMediaQueryWinsTransform } from 'style-value-parser';
+/*:: import { Rule } from 'eslint'; */
+
+const stylexValidateMediaQueries = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Warn when media query syntax may not be correctly ordered due to parser limitations',
```
mellyeliu:
```suggestion
        'Warn when media query syntax order is not properly enforced due to invalid media query syntax or parser limitations',
```

packages/@stylexjs/eslint-plugin/src/stylex-transform-media-queries.js
```diff
@@ -0,0 +1,227 @@
...
+    ): Property[] {
+      const mediaQueryProps: Property[] = [];
+
+      for (const prop of obj.properties) {
+        if (prop.type === 'SpreadElement') {
+          continue;
+        }
+
+        if (prop.computed) {
+          continue;
```
mellyeliu, line 82: this is fine for now, we'll eventually need to think of how to handle `defineConsts` computed keys here though

packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/flatten-raw-style-obj.js
```diff
@@ -31,7 +31,17 @@ export function flattenRawStyleObject(
       ? lastMediaQueryWinsTransform(style)
       : style;
   } catch (error) {
-    throw new Error(messages.INVALID_MEDIA_QUERY_SYNTAX);
+    if (options.softMediaQueryValidation) {
+      console.warn(
+        `StyleX: ${messages.INVALID_MEDIA_QUERY_SYNTAX}\n` +
+          'Media query order will not be respected. ' +
+          'This could be due to invalid media query syntax or unsupported edge cases in style-value-parser.\n' +
+          `Error: ${error.message || error}`,
```
mellyeliu, line 39: can we validate that this error message contains the erroring media query?

packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/flatten-raw-style-obj.js
```diff
@@ -31,7 +31,17 @@ export function flattenRawStyleObject(
       ? lastMediaQueryWinsTransform(style)
       : style;
   } catch (error) {
-    throw new Error(messages.INVALID_MEDIA_QUERY_SYNTAX);
+    if (options.softMediaQueryValidation) {
+      console.warn(
+        `StyleX: ${messages.INVALID_MEDIA_QUERY_SYNTAX}\n` +
```
mellyeliu, line 36: probably don't need to "StyleX: " prefix here

packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -254,6 +255,14 @@ export default class StateManager {
         'options.enableMediaQueryOrder',
       );
 
+    const softMediaQueryValidation: StyleXStateOptions['softMediaQueryValidation'] =
+      z.logAndDefault(
+        z.boolean(),
+        options.softMediaQueryValidation ?? false,
+        false,
```
mellyeliu, line 262: let's make this true for now

packages/@stylexjs/eslint-plugin/__tests__/stylex-transform-media-queries-test.js
```diff
@@ -0,0 +1,81 @@
...
+      })
+    `,
+    },
+    {
+      code: `
+      import * as stylex from '@stylexjs/stylex';
+      const styles = stylex.create({
+        main: {
+          color: {
+            '@media (min-width: 768px)': 'blue',
```
mellyeliu, line 42: we require a `default` with every contextual style:
```suggestion
            default: 'yellow',
            '@media (min-width: 768px)': 'blue',
```

packages/@stylexjs/eslint-plugin/__tests__/stylex-validate-media-queries-test.js
```diff
@@ -0,0 +1,81 @@
...
+        main: {
+          color: {
+            '@media not ((not (min-width: 400px))': 'blue',
+          },
+        },
+      })
+    `,
+      errors: [
+        {
+          message: /Media query order may not be respected/,
```
mellyeliu: Curious why the rest of the error is not showing up here? Seems out of sync with the rule file itself

packages/@stylexjs/eslint-plugin/src/index.js
```diff
@@ -15,6 +15,7 @@ import noUnused from './stylex-no-unused';
 import sortKeys from './stylex-sort-keys';
 import validShorthands from './stylex-valid-shorthands';
 import validStyles from './stylex-valid-styles';
+import validateMediaQueries from './stylex-validate-media-queries';
```
mellyeliu: Let's call this `transform-media-queries` or something similar

packages/@stylexjs/eslint-plugin/package.json
```diff
@@ -18,7 +18,8 @@
   "dependencies": {
     "css-shorthand-expand": "^1.2.0",
     "micromatch": "^4.0.5",
-    "postcss-value-parser": "^4.2.0"
+    "postcss-value-parser": "^4.2.0",
+    "style-value-parser": "*"
```
mellyeliu, line 22: I believe `style-value-parser` is already added to `workspaces` at the root level so it's inlined instead, as it's a private package, see: https://github.com/facebook/stylex/blob/main/package.json#L20

PR #1317, opened 2025-11-08
[playground] Introduce new potential replacement playground

henryqdineen:

The StyleX playground can be slow to load, to the point that it seems broken. I was reading old issues and got inspired by some impressive work by Jan Potoms (https://github.com/facebook/stylex/issues/732#issuecomment-2768404971) on an alternative StyleX playground. I think a playground like this can be really useful for somebody learning how StyleX works and the behavior of the transform.

This PR introduces a new `/playground-new` route (not in nav yet) that implements a playground based on (but less full-featured than) Jan's design and following Naman's [original requirements](https://github.com/facebook/stylex/issues/732#issue-2574858902). 

#### Technical details
- I noticed that the React docs used [Sandpack](https://sandpack.codesandbox.io/) so I figured it would be okay here (Jan's version used `react-runner`). I ran into some React version incompatibility with `@codesandbox/sandpack-react` and the old version of Docusaurus so I ended up using the lower-level `@codesandbox/sandpack-client` package. I can look into updating Docusaurus in a separate PR if you would like. 
- Since `@stylexjs/babel-plugin` cannot run in the browser I had to add some custom webpack configuration get it to work. An alternative solution could be to publish a second rollup bundle from `@stylejs/babel-plugin` that is designed to work in the browser ([POC](https://gist.github.com/henryqdineen/4ae7d7fbad32538a324b561aa3fd8a7f#file-rollup-config-mjs-L61-L109)).

## Linked PR/Issues

Related to https://github.com/facebook/stylex/issues/732

## Additional Context

localhost_3000_playground-new_

## Pre-flight checklist

### DIFF
--- a/packages/docs/components/PlaygroundNew.js
+++ b/packages/docs/components/PlaygroundNew.js
@@ -14,12 +14,7 @@ import { loadSandpackClient } from '@codesandbox/sandpack-client';
 import Editor from '@monaco-editor/react';
 import path from 'path-browserify';
 import { useColorMode } from '@docusaurus/theme-common';
-import {
-  useQueryParam,
-  StringParam,
-  ObjectParam,
-  withDefault,
-} from 'use-query-params';
+import { useQueryParam, ObjectParam, withDefault } from 'use-query-params';
 import { Tabs } from './playground-components/Tabs';
 import prettier from 'prettier';
 import * as babelPlugin from 'prettier/plugins/babel.js';
@@ -105,10 +100,7 @@ export default function PlaygroundNew() {
     'inputFiles',
     withDefault(ObjectParam, encodeObjKeys(INITIAL_INPUT_FILES)),
   );
-  const [activeInputFile, setActiveInputFile] = useQueryParam(
-    'activeInputFile',
-    withDefault(StringParam, 'App.js'),
-  );
+  const [activeInputFile, setActiveInputFile] = useState('App.js');
   const [transformedFiles, setTransformedFiles] = useState([]);
   const [cssOutput, setCssOutput] = useState('');
   const [sandpackInitialized, setSandpackInitialized] = useState(false);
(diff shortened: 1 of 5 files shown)


mellyeliu:
We have been wanting this for so long!

nmn:
@henryqdineen Thank you so much for working on this. As a related development, I have been working on an unplugin package that has a working plugin for esbuild, vite, rollup, rspack and webpack. It even works with RSCs (I tested redwoodSDK).

If we want to stick to using webContainers, that might improve performance.

> I can look into updating Docusaurus in a separate PR if you would like.

I'm hoping that the unplugin plugin can unblock that work as well.

> An alternative solution could be to publish a second rollup bundle from @stylejs/babel-plugin that is designed to work in the browser ([POC](https://gist.github.com/henryqdineen/4ae7d7fbad32538a324b561aa3fd8a7f#file-rollup-config-mjs-L61-L109)).

I'm open to shipping this if you can make a PR. I attempted this in the past and ran into some issues. Things have come a long way since so it may not be a problem anymore.

henryqdineen:
👍 Maybe the web containers-based playground and the one in the PR both have a place in the docs but I'm not sure how to differentiate them. Personally I more comfortable with my local IDE and don't find myself using tools like StackBlitz to any meaningful prototyping in the browser. The playground in this PR is definitely more along the lines of tools like AST explorer, Babel Repl, and TypeScript playground, which I have used countless times over the years.

I'll try to put up a PR for a browser version of the Babel plugin soon.

nmn:
@henryqdineen Sorry for the long delay while I've been catching up on a bunch of chores for the repo. After looking at the details, I think sandpack should be fine as the technology for the one and only playground for the website. What you've implemented is already doing most of what I would want from the playground.

I'll do some testing to check edge-cases for theming etc, and after that I'd feel happy to merge it.

henryqdineen:
Great! Let me know how I can help.

packages/docs/components/PlaygroundNew.jsx
```diff
@@ -0,0 +1,899 @@
...
+      if (sandpackClientRef.current) {
+        sandpackClientRef.current.destroy();
+      }
+    };
+  }, []);
+
+  return (
+    <div {...stylex.props(styles.container)}>
+      <div {...stylex.props(styles.column, styles.resizeH)}>
+        <Panel header="Source JS">
```
mellyeliu: Can we change these to "Input" "Output" or something similar

nmn:
Thank you @henryqdineen! This was a huge help.

PR #1320, opened 2025-11-12
[Experiment] Migrating docs to fumadocs + waku

nmn:

This is partially just a stress test for the Vite implementation of the unplugin package that this PR is stacked on top of.

But this could also be an opportunity to overhaul the documentation website with modern React, RSCs etc.

### The choices:

1. Fumadocs - this is the most popular documentation framework that uses RSCs
2. Waku - fumadocs is primarily for Next.js, but the unplugin package doesn't work with Next.js. If the issues with this setup cannot be resolved, we may consider using Next instead and leverage the postcss plugin there.
3. fumadocs-ui - This entire package has been inlined into the project so we can migrate the code from tailwind to stylex and properly dogfood stylex for its own website.

## Additional Context

Screenshot 2025-11-12 at 1 11 52 AM

Screenshot 2025-11-12 at 1 12 17 AM

## Status

Almost everything works with some annoying bugs when building for prod
- import statements in MDX throw an error in the parser
- There are two other errors in the SSR code somewhere where React is getting an invalid child to render


nmn:
The second experiment is working better than this.

Issue #889, opened 2025-02-11
Types are wrong for @stylexjs/babel-plugin


jfirebaugh:
### Describe the issue

The resolved types use `export default` where the JavaScript file appears to use `module.exports =`. This will cause TypeScript under the `node16` module mode to think an extra `.default` property access is required, but that will likely fail at runtime.

### Expected behavior

These types should use `export =` instead of `export default`.

### Steps to reproduce

See https://arethetypeswrong.github.io/?p=%40stylexjs%2Fbabel-plugin%400.10.1

zaydek:
Just noticed this myself. Thanks, also following this issue.

RavenColEvol:
I spent some time understanding this issue, here are my findings
Since output generated is in cjs format it results in `module.exports` output. `d.ts` file should have typescript `export =` instead of `export default` since output is cjs. 

The functionality which generates the types today ( .js -> .flow -> .d.ts ) there is no way for us to tell translateFlowDefToTSDef to use `export =` instead of `export default`

I tried changing export in index.js from export default to common js module.exports but it throws error in translateFlowDefToTSDef method.

@jfirebaugh @nmn

nmn:
This is a bug in the `flow-api-translator` package due to differences in how Flow and Typescript.

The easiest fix would be to add a manual `.d.ts` file. Our script already prioritize those when they exist.

AlexanderOMara:
I created another slight-related issue because the TypeScript types appear to be super invalid beyond what is discussed here.

https://github.com/facebook/stylex/issues/969

mizoz:
Hi! I'd love to contribute. Could you assign this issue to me?

PR #1322, opened 2025-11-12
[types] fix @stylexjs/babel plugin cjs compat

henryqdineen:

In ce530d0d91b24e522ffcd0525598eddd1887009b I pushed a fix for a TypeScript error that we were seeing.

```
node_modules/@stylexjs/babel-plugin/lib/index.d.ts:53:1 - error TS2309: An export assignment cannot be used in a module with other exported elements.

53 export = $$EXPORT_DEFAULT_DECLARATION$$;
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I'm assuming most other's were not hitting this if they set `skipLibCheck: true`. The `v0.16.3` release fixed the error and our StyleX integration works as expected. 

I was reading through old issues I discovered https://github.com/facebook/stylex/issues/889 and the attempted fix (#924). I think my commit ultimately reverted their fix and likely caused a regression for some people. See: https://arethetypeswrong.github.io/?p=%40stylexjs%2Fbabel-plugin%400.16.3

My fix in this PR is to go back to using `export =` but use a namespace so we don't run into the `"An export assignment cannot be used in a module with other exported elements"` error.  This commit also corrects the `Rule` type and `config` argument for `processStylexRules`, which have changed since this file was introduced. 

Sorry for the regression. Thanks!

## Linked PR/Issues

https://github.com/facebook/stylex/issues/889

## Pre-flight checklist


packages/@stylexjs/babel-plugin/src/index.d.ts
```diff
@@ -10,44 +9,55 @@
...
-): string;
-export type StyleXTransformObj = Readonly<{
-  (): PluginObj;
-  withOptions: typeof stylexPluginWithOptions;
-  processStylexRules: typeof processStylexRules;
-}>;
-declare const $$EXPORT_DEFAULT_DECLARATION$$: StyleXTransformObj;
-export default $$EXPORT_DEFAULT_DECLARATION$$;
+
+declare namespace styleXTransform {
```
nmn, line 15: Can we avoid using a namespace here? Other than that the changes look good.

nmn (changes requested):

packages/@stylexjs/babel-plugin/src/index.d.ts
henryqdineen: Thanks @nmn. I was following this advice from the [DefinitelyTyped README](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md). Do you have a recommended solution without a namespace?


> 
> When the implementation package uses `module.exports = ...`, the DefinitelyTyped package should use `export =`, not `export default`. (Alternatively, if the `module.exports` is just an object of named properties, the DefinitelyTyped package can use a series of named exports.) The most common obstacle to correcting this problem is confusion about how to export types in addition to the primary export. For example, assume these types are incorrectly using `export default`:
> 
>   ```ts
>   export interface Options {
>       // ...
>   }
>   export default function doSomething(options: Options): void;
>   ```
> 
>   Changing the `export default` to an `export =` creates an error:
> 
>   ```ts
>   export interface Options {
>       // ...
>   }
>   declare function doSomething(options: Options): void;
>   export = doSomething;
>   // ^^^^^^^^^^^^^^^^^
>   // Error: An export assignment cannot be used in a module with other exported elements.
>   ```
> 
>   To fix this, move the types inside a namespace with the same name as the function:
> 
>   ```ts
>   declare namespace doSomething {
>       export interface Options {
>           // ...
>       }
>   }
>   declare function doSomething(options: doSomething.Options): void;
>   export = doSomething;
>   ```

packages/@stylexjs/babel-plugin/src/index.d.ts
```diff
@@ -10,44 +9,55 @@
...
-  rules: Array<Rule>,
-  config?:
-    | boolean
-    | {
-        useLayers?: boolean;
-        enableLTRRTLComments?: boolean;
-        legacyDisableLayers?: boolean;
-      },
-): string;
-export type StyleXTransformObj = Readonly<{
```
nmn, line 58: I wonder if you can use this type and use the `export =` syntax to make it work with cjs?

henryqdineen: I can't seem to find a way to do it with running in to "An export assignment cannot be used in a module with other exported elements.(2309)"

PR #1323, opened 2025-11-13
[feature] Add stylex.env for compile time constants

nmn:

Partially implements #1358

StyleX already supports `stylex.defineConsts` but this has a few limitations and trade-offs:
- It can only store individual values. It cannot store objects with different values for different pseudo-classes and media queries for example.
- It generates CSS variables at compile time, which are then inlined at bundle-time. This means it generates different className hashes than what would've been generated if the value was used directly.

This PR introduces a new feature called `stylex.env`. This is fundamentally a very simple feature that is similar to how `process.env` is already used in many applications. It allows you to configure the stylex babel plugin with constants that are inlined *before* the stylex compiler runs. It can also contain arbitrary values.

The main trade-off here is that since this is something that is part of the babel config, every change to the constants provided will invalidate the entire cache of stylex compilation.

## Additional Context

Some basic tests have been included. Will add more tests with more advanced use-cases.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/utils/state-manager.js
+++ b/packages/@stylexjs/babel-plugin/src/utils/state-manager.js
@@ -150,8 +150,8 @@ const checkEnvOption: Check<$ReadOnly<{ [string]: mixed }>> = (
 ) => {
   if (typeof value !== 'object' || value == null || Array.isArray(value)) {
     return new Error(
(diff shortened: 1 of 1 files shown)


mellyeliu:
Some test failures. Let's also discuss whether shareable functions will fit in here as well or under a different API

PR #1330, opened 2025-11-20
[babel-plugin] add debugFilePath option 

henryqdineen:

I have an extremely basic chrome extension locally where I want to show the styles for an element and link to the "Sources" tab.

Screenshot 2025-11-20 at 4 03 21 PM

The problem I ran into was that filenames in `data-style-src` do not have enough information for me to make a `chrome.devtools.panels.openResource()` call when the link is clicked. 

My solution in this PR is to add `debugFilePath` option that we can use to customize the filenames used in `data-style-src`. In our case the `debugFilePath` implementation will be very similar to our logic in the our [Rspack `output.devtoolModuleFilenameTemplate` config](https://rspack.rs/config/output#outputdevtoolmodulefilenametemplate), which determines the filenames in the source map. 

We don't have an immediate need for this feature but I figured it was worth creating a PR to open the discussion. We are interested in any future devtools and out want to make sure they work for us. Thanks!

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/utils/add-sourcemap-data.js
+++ b/packages/@stylexjs/babel-plugin/src/utils/add-sourcemap-data.js
@@ -42,10 +42,9 @@ function createShortFilename(
   absolutePath: string,
   state: StateManager,
 ): string {
-  if (typeof state.options.createShortFilename === 'function') {
-    return state.options.createShortFilename(absolutePath);
+  if (typeof state.options.debugFilePath === 'function') {
+    return state.options.debugFilePath(absolutePath);
   }
-  const isHaste = state.options.unstable_moduleResolution?.type === 'haste';
 
   const cwdPackage = state.getPackageNameAndPath(process.cwd());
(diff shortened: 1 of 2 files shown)


mellyeliu:
Awesome! Just needs flow fixes. We have it in the backlog to create a more comprehensive dev tools extension.

nmn:
One change I would suggest is to not nest the config option under `unstable_moduleResolution`.

That is very specific to theming APIs and this work is more general than that, so it probably makes sense to make it a top-level config option.

henryqdineen:
Sorry for the delay. I just pushed a commit that fixes the flow types and moves `createShortFilename` to be a top-level config option. Let me know if there is a more appropriate name for the option. `debugFilePath`?

nmn:
Let's rename the option to `debugFilePath` and please rebase this PR.

nmn (approved):
The changes themselves look good. Should be good to merge after a rebase and change the name of the config option.

henryqdineen:
Rebased, renamed, and updated PR subject/body. Thanks!

PR #1331, opened 2025-11-21
[unplugin] Add support for compiling stylex from imported packages

nmn:

The unplugin package today has two known issues:
1. The documentation in the Readme is incorrect and misleading
2. It doesn't handle stylex within external packages correctly in a vite project.

This PR attempts to fix both those issues.

For external packages, the plugin will attempt to detect them automatically, but it also lets you pass in a config option as an escape hatch

## Next Steps

1. The documentation should be a lot better, but it needs to be verified and rewritten for clarity.
2. Additional testing for building packages that use packages that use StyleX.

### DIFF
--- a/examples/example-react-router/package.json
+++ b/examples/example-react-router/package.json
@@ -3,13 +3,14 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "example:build": "vite build",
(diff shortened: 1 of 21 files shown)


sfc-gh-alisowski (commented):
Looks great!

PR #1332, opened 2025-11-21
[fix] Show full debug file paths

nmn:

For the `data-style-src` attribute to be useful for chrome dev tools extensions etc, it is important to have the full file path available.

Today the path always just shows the arbitrary last two parts of the path.

This new implementation searches for the nearest `pcakage.json` file uses that as the "root" to show the relative file path from. The package name from the `package.json` file is used as a prefix

Additionally, it detects your own `package.json` file by walking up from `cwd` and drops the package name prefix if the file is not from a third-party package.

This code, similar to the code used by theming APIs, looks for package.json files so it can work with esoteric package managers where the package may or may not be in a `node_modules` folder. We do, however, expect the existance of the `package.json` file for this to work.

--

As a fallback, it will use the configured `rootDir` as a fallback root.

## TODO:

- [ ] Improve tests, to avoid long `../../` prefixes
- [ ] Verify that files within `node_modules` folder are being treated correctly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1025,7 +1025,8 @@
         "react": "19.3.0-canary-561ee24d-20251101",
         "react-dom": "19.3.0-canary-561ee24d-20251101",
         "react-server-dom-webpack": "19.3.0-canary-561ee24d-20251101",
-        "rwsdk": "1.0.0-beta.31"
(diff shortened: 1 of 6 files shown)


jcperez-ch:
LGTM

PR #1335, opened 2025-11-22
Chore/use yarn

nmn:

I have seen repeated issues with NPM not installing packages correctly.
- Multiple versions of react in the same app because different packages rely on different versions
- Errors being thrown because sub-dependencies cannot be resolved

This is in addition to issues such as slow installs etc.

---

The current implementation uses `yarn 1` because Meta uses it and may benefit from having a `yarn.lock` file available.

However, we may also be able to use `bun` as the package manager as it can be configured to generate a `yarn 1` compatible `yarn.lock` file. This would give us an even faster package managed during OSS development while Meta can still leverage the `yarn.lock` file.


nmn:
Closing this for now as the `package-lock.json` file has been fixed to let the examples compile for now.

Also, all the dependencies have already been updated.

PR #1336, opened 2025-11-23
[chore] Update dependencies

nmn:

This PR just updates a bunch NPM dependencies to their latest versions. 

(This PR was extracted from the `yarn` PR but without the change from NPM to YARN)

The NPM issues remain, at least on my machine where some of the dependencies within examples can't resolve their sub-dependencies, but the core packages are all working.

With the latest version of Flow and hermes, it is also be possible to use the `match` syntax. I've added one usage of it so far as a proof of concept.

## Pre-flight checklist


nmn:
I have managed to get the package-lock file in a shape that the examples are now compiling successfully. We can hopefully put off a migration to `yarn` for now.

PR #1338, opened 2025-11-23
[chore] Upgrade to Docusaurus 3

nmn:

Upgrade the documentation website to docusaurus 3. With some tweaks to the unplugin configuration, production builds are handled by the `unplugin` package while it still relies on runtime injection during development.

There was a small tweak needed to the unplugin package but there doesn't seem to be a regression in the webpack example as a result.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1057,41 +1057,6 @@
         "wrangler": "^4.50.0"
       }
     },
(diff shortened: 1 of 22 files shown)


mellyeliu:
should we just wait for the rewrite a fumadocs? too many docs PRs

nmn:
No longer using Docusaurus

PR #1340, opened 2025-11-24
[chore] Update and unify default config options

nmn:

Updating our default config options to streamline `debug` and `dev` options.

- `enableDebugClassNames` is now `false` by default
- `enableDevClassNames` will once again be turned on when `dev` is true

Outside of that, the default values of all config options have been unified.

---

Additionally, The `unplugin` package has been updated with better types and to default with `treeshakingCompensation` on for vite, rollup and rolldown.

These bundlers need this option to handle theme imports correctly.


mellyeliu (approved):
fix tests

PR #1352, opened 2025-12-01
[babel-plugin] optimize and inline variables in processStylexRules

mellyeliu:
Currently, we output all `defineVars` and `createTheme` variables in the compiled sheet even if such variables are never used. This is costly for design systems that define large token objects via `defineVars`, because the browser ends up allocating memory for thousands of unused variables.

This PR optimizes two cases:
1. Removes all `defineVars` generated variables if they are never referenced
2. Inlines `defineVars` CSS variables with their values if CSS variables are used once

Important: This is behind a param `optimizeDefineVars` because this cannot yet be used internally. We don't have global variable knowledge and would need to modify our build step to bundle all defineVars files with each sheet. `optimizeDefineVars` must be set to false where we pipe out to `processStylexRules ` on next sync

Todo:
- [ ] Improve tests 
- [ ] Refactor and clean up regex patterns

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -182,13 +182,9 @@ describe('@stylexjs/babel-plugin', () => {
           useLayers: false,
           enableLTRRTLComments: false,
         }),
-      ).toMatchInlineSnapshot(`
-        ":root, .xsg933n{--blue-xpqh4lw:blue;--marginTokens-x8nt2k2:10px;--colorTokens-xkxfyv:red;}
-        :root, .xbiwvf9{--small-x19twipt:2px;--medium-xypjos2:4px;--large-x1ec7iuc:8px;}
-        @media (prefers-color-scheme: dark){:root, .xsg933n{--colorTokens-xkxfyv:lightblue;}}
-        @media (min-width: 600px){:root, .xsg933n{--marginTokens-x8nt2k2:20px;}}
-        @supports (color: oklab(0 0 0)){@media (prefers-color-scheme: dark){:root, .xsg933n{--colorTokens-xkxfyv:oklab(0.7 -0.3 -0.4);}}}"
-      `);
+      ).toMatchInlineSnapshot(
+        '"@supports (color: oklab(0 0 0)){@media (prefers-color-scheme: dark){:root, .xsg933n{--colorTokens-xkxfyv:oklab(0.7 -0.3 -0.4);}}}"',
+      );
     });
 
     test('all rules (useLayers:false)', () => {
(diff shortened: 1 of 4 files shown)


packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -182,13 +182,9 @@ describe('@stylexjs/babel-plugin', () => {
...
         }),
-      ).toMatchInlineSnapshot(`
-        ":root, .xsg933n{--blue-xpqh4lw:blue;--marginTokens-x8nt2k2:10px;--colorTokens-xkxfyv:red;}
-        :root, .xbiwvf9{--small-x19twipt:2px;--medium-xypjos2:4px;--large-x1ec7iuc:8px;}
-        @media (prefers-color-scheme: dark){:root, .xsg933n{--colorTokens-xkxfyv:lightblue;}}
-        @media (min-width: 600px){:root, .xsg933n{--marginTokens-x8nt2k2:20px;}}
-        @supports (color: oklab(0 0 0)){@media (prefers-color-scheme: dark){:root, .xsg933n{--colorTokens-xkxfyv:oklab(0.7 -0.3 -0.4);}}}"
-      `);
+      ).toMatchInlineSnapshot(
+        '"@supports (color: oklab(0 0 0)){@media (prefers-color-scheme: dark){:root, .xsg933n{--colorTokens-xkxfyv:oklab(0.7 -0.3 -0.4);}}}"',
```
nmn, line 186: This is likely not valid. This transform ONLY transforms the main file and NO styles are used. So the expected result is an empty string.

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -643,25 +619,21 @@ describe('@stylexjs/babel-plugin', () => {
...
         @supports (color: oklab(0 0 0)){@media (prefers-color-scheme: dark){:root, .xsg933n{--xkxfyv:oklab(0.7 -0.3 -0.4);}}}
         .x4hn0rr.x4hn0rr, .x4hn0rr.x4hn0rr:root{--x1ec7iuc:20px;--xypjos2:10px;--x19twipt:5px;}
         .x1coplze.x1coplze, .x1coplze.x1coplze:root{--xpqh4lw:lightblue;}
         .xufgesz{--orange-theme-color:red}
         .xymmreb{margin:10px 20px}
-        .x1s2izit{padding:var(--x1ec7iuc)}
+        .x1s2izit{padding:8px}
         .x1bg2uv5{border-color:green}
-        @media (max-width: 1000px){.xio2edn.xio2edn{border-color:var(--xpqh4lw)}}
+        @media (max-width: 1000px){.xio2edn.xio2edn{border-color:blue}}
```
nmn, line 629: This is also invalid because it is inlining values of variables. We don't want to do this in most cases.

nmn:
> 2. Inlines defineVars CSS variables with their values if CSS variables are used once

This needs to be a separate option and not enabled as the same thing.

Also, we need to special-case variable names that start with `--` as that is an indication that the variables may be used outside of StyleX.

packages/@stylexjs/babel-plugin/src/index.js
```diff
@@ -389,6 +389,137 @@ function getLogicalFloatVars(rules: Array<Rule>): string {
     : '';
 }
 
+function optimizeDefineVarsRules(rules: Array<Rule>): {
+  optimizedRules: Array<Rule>,
+  varsToInline: Map<string, string>,
+} {
+  const defineVarsVariables: Map<string, string> = new Map();
+  const varUsageCount: Map<string, number> = new Map();
```
nmn, line 397: Prefer using objects for this. They are faster when the key is a string.

packages/@stylexjs/babel-plugin/src/index.js
```diff
@@ -389,6 +389,137 @@ function getLogicalFloatVars(rules: Array<Rule>): string {
...
+
+  for (const [, ruleObj] of rules) {
+    const { ltr, rtl } = ruleObj;
+    const isDefineVarsRule = /^(?:@[^{]*\{)?:root,\s*\.[\w-]+\{/.test(ltr);
```
nmn, line 401: We probably want to change the metadata generated by `defineVars` before this PR and avoid regex-based fragile implementations.

mellyeliu:
Should've marked this as a draft. Just put it up as a placeholder

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
mellyeliu: Let's put this behind a separate config

PR #1368, opened 2025-12-03
[stylex] improve treeshakability

henryqdineen:

This PR makes a couple changes to improve the treeshakability of the `@stylexjs/stylex` package. 

- Set `sideEffects` flag to `false` in the `package.json`.  Although rare in typical usage this would allow bundlers to remove unused imports `from '@stylexjs/stylex'`. This is also a general best practice if the library has no side effects. I noticed that `styleq` already has this flag set
- Moved the code that patches properties onto `legacyMerge` into a IIFE and annotated it with `/*@__PURE__*/` with a rollup plugin. Even with `sideEffect: false` set some bundlers (definitely webpack from my testing) would consider setting a property like `_legacyMerge.create =` to be a side effect and the code would not be marked for dead code elimination. I tried  adding`/*@__PURE__*/` in the source but it was not being preserved, so I resorted to a rollup plugin. I think there is something I couldn't easily figure out about the flow or hermes-parser plugins that remove comments.

With webpack and this change I am seeing a bundle size reduction from `3.4KB` to `1.4KB` for a simple entry point that only uses the `props` import (webpack is able to analyze namespace imports in most cases)

```
console.log(stylex.props)
```

### Questions

- Can we just drop `legacyMerge` or at least the patching of the variables onto it? It seems undocumented?

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/stylex/src/stylex.js
+++ b/packages/@stylexjs/stylex/src/stylex.js
@@ -339,5 +339,6 @@ export const legacyMerge: IStyleX = /*@__PURE__*/ (function () {
   _legacyMerge.types = types;
   _legacyMerge.when = when;
   _legacyMerge.viewTransitionClass = viewTransitionClass;
+  _legacyMerge.env = env;
   return _legacyMerge;
 })();


nmn:
> Can we just drop legacyMerge or at least the patching of the variables onto it? It seems undocumented?


We're already discussing this...

packages/@stylexjs/stylex/package.json
```diff
@@ -65,5 +65,6 @@
   },
   "files": [
     "lib/*"
-  ]
+  ],
+  "sideEffects": false
```
nmn, line 68: 👍

nmn (approved):
Sorry for the long delay in merging this. We should rebase and merge this ASAP!

PR #1376, opened 2025-12-04
[unplugin] handle file suffixes with query params (e.g., Vite) correctly

eungyeole:
## What changed / motivation

When using external StyleX packages (e.g., `@stylex/shared-ui`) in a **Vite dev** environment,  
Vite automatically appends cache-busting query strings like `?v=xxxx` to file requests.

Example:
```
/node_modules/@stylex/shared-ui/token.stylex.js?v=c165e250
```

Previously, the plugin passed this entire id (including the query string) directly to `runBabelTransform`.  
Because the id contained `?v=xxxx`, the file extension could not be correctly identified, which resulted in the following error:

```
Pre-transform error: /Users/.../node_modules/@stylex/shared-ui/token.stylex.js?v=c165e250: Unable to generate hash for defineVars(). Check that the file has a valid extension and that unstable_moduleResolution is configured. Plugin: @stylexjs/unplugin
```

### Changes

- Extracted a clean filename by stripping everything after `?` or `#`,  
  and passed this normalized filename to `runBabelTransform`  
  (`const filename = id.split(/[?#]/)[0]`)

### Result

- `.stylex.js` files are now correctly recognized even when Vite appends query strings in dev mode  
- Fixes the Pre-transform error when importing external StyleX packages

## Additional Context

- This issue occurs **only in Vite dev mode** when external StyleX packages are used  
- The problem does not appear during build because no query strings are added

```ts
// vite.config.ts
import stylex from "@stylexjs/unplugin";
import react from "@vitejs/plugin-react";
import { defineConfig } from "vite";

export default defineConfig({
  plugins: [react(), stylex.vite({})],
});

```

```tsx
// index.tsx
import { create, props } from "@stylexjs/stylex";
import { createRoot } from "react-dom/client";

import { token } from "@stylex/shared-ui/token.stylex";
import "./index.css";

function Playground() {
  return <div className={props(styles.root).className}>Hello</div>;
}

const root = createRoot(document.getElementById("root")!);

root.render(<Playground />);

const styles = create({
  root: {
    backgroundColor: token.colorSurfacePrimary,
  },
});

```

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/unplugin/src/index.js
+++ b/packages/@stylexjs/unplugin/src/index.js
@@ -234,14 +234,14 @@ const unpluginInstance = createUnplugin((userOptions = {}, metaOptions) => {
     : path.join(process.cwd(), 'node_modules', '.stylex');
   const DISK_RULES_PATH = path.join(DISK_RULES_DIR, 'rules.json');
 
-  async function runBabelTransform(inputCode, id, callerName) {
+  async function runBabelTransform(inputCode, filename, callerName) {
     const result = await transformAsync(inputCode, {
       babelrc: false,
-      filename: id,
+      filename,
       presets,
       plugins: [
         ...plugins,
-        /\.jsx?/.test(path.extname(id))
+        /\.jsx?/.test(path.extname(filename))
           ? flowSyntaxPlugin
           : [typescriptSyntaxPlugin, { isTSX: true }],
         jsxSyntaxPlugin,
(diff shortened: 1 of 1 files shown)


pawelblaszczyk5:
Nice find! That'll probably impact some frameworks also (so build time as well) which e.g. split files via plugins and use query params to address specific parts. Both React Router and TanStack Router/Start does this.

mellyeliu:
Thanks for fixing this!

PR #1345, opened 2025-11-26
[docs] add LLM context for style authoring

mellyeliu:
## Context

Adding a LLM reference section with two context files for style authoring and installation. 

image

## Follow-ups
- [ ] Add download to code blocks
- [ ] Add these appended files to a `llm.txt` endpoint alongside the existing `llms-full.txt`endpoint
- [ ] ESLint fixer guides for LLMs and instructions to disallow certain CSS properties or practices (border shorthands, disallowed properties within StyleX, prefer conditional styles over dynamic styles when possible, etc)
- [ ] Migration instructions from other styling frameworks to StyleX

...etc

PR #1381, opened 2025-12-07
[eslint] no-conflicting-props rule

henryqdineen:

>  Never apply `className` and `style` properties alongside a `{...stylex.props(...)}` spread

I saw the above recommendation in the draft LLM context PR (#1345) and I thought it made sense as an ESLint rule.

This PR adds a new `no-conflicting-props` ESLint rule which disallows usage of `className` and `style` props on the same JSX element as a `{...stylex.props(...)}` spread, both before and after the spread. I don't have a great sense for how often people might run into this footgun but it seemed easy enough to statically analyze and imakes sense as an ESLint rule.

I considered that we might want to have the rule disallow JSX spreads after `stylex.props` (e.g `<div {...stylex.props(...)} {...rest}>` in case the `rest` has `className` or `style` but I did not include that in this PR.

#### Notes
- The `estree` types that are used in this and other ESLint rules don't have JSX types so I had to roll my own. I considered adding a dev dependency on `hermes-estree`, which seems to have a more complete set of types but I noticed that the `JSXAttribute.name` type was not accurate (or at least consistent with `@babel/types` and `ast-types`).

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/src/stylex-no-conflicting-props.js
+++ b/packages/@stylexjs/eslint-plugin/src/stylex-no-conflicting-props.js
@@ -42,10 +42,6 @@ const stylexNoConflictingProps = {
       category: 'Best Practices',
       recommended: true,
     },
-    messages: {
-      conflictingProp:
-        'The `{{propName}}` prop should not be used when spreading `stylex.props()` to avoid conflicts.',
-    },
     schema: [
       {
         type: 'object',
@@ -113,7 +109,8 @@ const stylexNoConflictingProps = {
             context.report({
               // $FlowFixMe[incompatible-type]
               node: attr,
-              messageId: 'conflictingProp',
+              message:
+                'The `{{propName}}` prop should not be used when spreading `stylex.props()` to avoid conflicts.',
               data: { propName: attr.name.name },
             });
           } else if (
@@ -130,7 +127,8 @@ const stylexNoConflictingProps = {
                 context.report({
                   // $FlowFixMe[incompatible-type]
                   node: prop,
-                  messageId: 'conflictingProp',
+                  message:
+                    'The `{{propName}}` prop should not be used when spreading `stylex.props()` to avoid conflicts.',
                   data: { propName: prop.key.name },
                 });
               }


packages/@stylexjs/eslint-plugin/src/stylex-no-conflicting-props.js
```diff
@@ -0,0 +1,125 @@
...
+  +type: 'JSXOpeningElement',
+  +attributes: $ReadOnlyArray<JSXAttribute | JSXSpreadAttribute>,
+};
+
+const stylexNoConflictingProps = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Disallow using `className` or `style` props on elements that use `stylex.props()`',
```
mellyeliu:
```suggestion
        'Disallow using `className` or `style` props on elements that spread `stylex.props()`',
```

packages/@stylexjs/eslint-plugin/__tests__/stylex-no-conflicting-props-test.js
```diff
@@ -0,0 +1,274 @@
...
+          main: { color: 'red' },
+        });
+        function Component() {
+          return <div {...css.props(styles.main)} />;
+        }
+      `,
+    },
+  ],
+  invalid: [
+    {
```
mellyeliu, line 114: can we add a test that checks for a conflict via another spread?

henryqdineen: What are you thinking by "conflict via another spread"? With the current implementation these would be false negatives:

```jsx
<div {...stylex.props(styles.main)} {...{ className: 'foo' }} />
<div {...stylex.props(styles.main)} {...objThatHasClassName} />
```

The first example just seems unlikely and the second one seems difficult to implement without deeper analysis and/or type awareness. The simplest thing I can think of is to enforce that the `stylex.props` spread is the last spread so it overwrites previous conflicting props.

Let me know what you were thinking. Thanks!

mellyeliu: The inline spread, the latter is probably overkill. We've seen some ugly jsx before (though uncommon!) and it's an easy check.

We shouldn't enforce `stylex.props` being last, since the goal of this guidance is to avoid multiple sources of truth for `className` and `style` altogether.

nmn: > We shouldn't enforce stylex.props being last

It might make sense to enforce that it is the last spread?

---

Separately, I was thinking of an enhancement to Babel to check for cases like:

```tsx
<div {...props} {...stylex.props} />
```

And logging an error in `dev` if `props` contains `className` or `style`?

henryqdineen: Thanks. I pushed a commit in e6fa4b810b98c73c236b4560c225ca91d92b8d82 to handle the inline spread (e.g `{...{className: 'foo'}`). I considered handling stuff like `{...{...{className: 'foo'}}` but it seemed like an extreme edge case.

mellyeliu: > It might make sense to enforce that it is the last spread?

Personally, would rather just disallow the pattern altogether than patch over the problem to have StyleX overwrite things when we don't know the author's intent. But no strong opinion.

> Separately, I was thinking of an enhancement to Babel

Yeah good call! We should add more compiler validation in general.

henryqdineen:
> Happy with this, good to go aside from Flow

Thanks. I fixed flow in [b335eb0](https://github.com/facebook/stylex/pull/1381/commits/b335eb026a5e516c06670b0d0fcd25adc1b5ec1a)

PR #1392, opened 2025-12-17
[playground] add @stylexjs/stylex types to editor

henryqdineen:

This PR adds the `@stylexjs/stylex` types to the playground editor so we can have intellisense when using StyleX APIs.

## Additional Context

Screenshot 2025-12-16 at 7 50 41 PM

## Pre-flight checklist

### DIFF
--- a/packages/docs/components/PlaygroundNew.js
+++ b/packages/docs/components/PlaygroundNew.js
@@ -360,6 +360,24 @@ export const vars = stylex.defineVars({
                 key={activeInputFile}
                 onChange={handleEditorChange}
                 onMount={(editor) => {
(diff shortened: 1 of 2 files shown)


nmn:
What would it take to change the playground itself to use .tsx files instead? I think it’s important for us to emphasize the usage of types with StyleX.

henryqdineen:
> What would it take to change the playground itself to use .tsx files instead? I think it’s important for us to emphasize the usage of types with StyleX.

I agree and I'll look into that. I think `@babel/standalone` has the TypeScript transform. Another follow-up would be to include the React types.

mellyeliu:
preview at https://stylex-4f3fabb93-fbopensource.vercel.app/playground/

PR #1372, opened 2025-12-04
[babel-plugin] improve AST detection of named exports for defineVars/defineMarker/defineConsts 

mellyeliu:
We enforce named exports for APIs like `defineVars` for cacheability reasons. Previously, we only allowed named exports like `export const x;`, we should expand this to allow for syntax like below. This issue has been reported a few times and is a blocker for some internal migrations.

```jsx
// allowed
export const x = stylex.defineVars(...)

// allowed
const x = stylex.defineVars(...)
export { x }

// not allowed
export default x;

// not allowed
export { x } from './y'
```

Tested:
- [x] export const x = stylex.defineVars(...)
- [x] const x; export { x }
- [x] export default x
- [x] export { x } from './y'
- [x] const x; export {x as y};
- [x] export {x as y} from './y';

Fixes https://github.com/facebook/stylex/issues/348, https://github.com/facebook/stylex/issues/1098

PR #1394, opened 2025-12-18
[babel-plugin] fix named export detection 

mellyeliu:
Improves upon the previous fix https://github.com/facebook/stylex/pull/1372, where named exports are not reliably detected after module transformations (Next.js with Turbopack). The previous logic relied on identifier names/AST structure to match variables with exports.

This new logic avoids those assumptions and instead relies on variable bindings. 

Tested on NextJS + Turbopack (where bug was originally found), docs (unplugin) (already worked):

Allowed:
 - export const x = stylex.defineVars(...)
- const x; export { x }

Disallowed:
- export default x
-  export { x } from './y'
-  const x; export {x as y};

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
+++ b/packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
@@ -211,62 +211,63 @@ export function getProgramStatement(path: NodePath<>): NodePath<> {
  * - Direct named exports: `export const x = ...`
  * - Locally declared named exports: `const x = ...; export { x }`
  *
- * Default exports and re-exports from other files (e.g., `export { x } from './other'`) are NOT allowed.
+ * Default exports, aliasing and re-exports from other files (e.g., `export { x } from './other'`) are NOT allowed.
  */
 export function isVariableNamedExported(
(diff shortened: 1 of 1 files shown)


packages/@stylexjs/babel-plugin/src/utils/ast-helpers.js
```diff
@@ -211,62 +211,63 @@ export function getProgramStatement(path: NodePath<>): NodePath<> {
...
       }
-      if (node.specifiers == null || node.specifiers.length === 0) {
-        return;
+
+      for (const s of node.specifiers ?? []) {
+        if (s.type !== 'ExportSpecifier') continue;
+        if (s.local.type !== 'Identifier') continue;
+        if (s.exported.type !== 'Identifier') continue;
+
+        if (s.local.name !== s.exported.name) continue;
```
henryqdineen, line 260: Super edge case but it could also help to also continue `if (s.exportKind === 'type')` to prevent a false positive.

```ts
type foo = any;
const foo = {};

export {
 type foo
}
```

https://babeljs.io/docs/babel-types#exportspecifier

PR #1401, opened 2025-12-20
[feat] StyleX Dev Tools Extension

nmn:

A work-in-progress implementation of a devtools extension for StyleX and atomic styles in general.

Features:
- [x] Use `data-style-src` to list the sources of styles in order
  - [x] Preview the relevant StyleX style objects inline
  - [x] One click action to show the original styles in the original file in the `sources` tab.
  - [ ] Detect which style properties from each object are overridden
- [x] List all atomic styles that are actually applied by className
- [x] Ability to edit/override styles
- [x] Ensure it works without runtime injection
- [x] Good styles with light and dark mode
- [x] Good styles for editing styles
- [x] Show computed values on hover 

## Additional Context

Current Status:
(hovering on a property shows the computed style)

Screenshot 2025-12-21 at 2 12 19 AM

## Pre-flight checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -14,4 +14,4 @@ examples/example-storybook/storybook-static
 examples/example-cli/src
 packages/docs/.source
 devtools-extension/extension
-devtools-extension/extension/assets/*
+**/devtools-extension/extension/assets/**


henryqdineen:
I'm curious how resource-matching works when things are bundled in dev mode like in webpack/rspack. In my experimentation I found that `getResources()` only returns the actual resources that are loaded over the network. So if the styles are in `Button.tsx` and bundled into `app.js` how do you make that connection?

PR #1406, opened 2025-12-22
[atoms] support for inline styles

mellyeliu:
## Implementation

Adding support for inline styles authoring with a new `@stylexjs/atoms` package. This allows you to write styles inline in `stylex.props()` that are pre-compiled within the props visitor using much of the `create` logic for static and dynamic styles.

This is something we went back and forth on a few times, but ultimately felt that it's good to provide the optionality. We still recommend (and want to bake in enough friction) for people to use `stylex.create` namespaces for the majority of usecases for readability/maintainability reasons. 

```tsx
import x from '@stylexjs/atoms';

<div {...stylex.props(
  x.display.flex,
  x.padding._16px,
  x.width['calc(100% - 20cqi)'],
  x.color(color),
)} />
```

The naming of all the above is up for discussion, as well as what styles we should build into this API. I'm also using `Properties` type from estree for type checking for a first pass, but there might be something better we can use.

## Testing

Style types:
- [x] Static values 
- [x] Leading underscores 
- [x] Object key syntax
- [x] Dynamic inline styles

Tested within docs page the above +
- [x] Different merges of dynamic, regular, and inline styles
- [x] Named imports 

Deferring to follow ups:
- [ ] Rethink contextual styles
- [ ] Make work with valid-styles / some way to add linter validation
- [ ] Add docs
- [ ] Rethink package name + standard import alias
- [ ] figure out import and package naming?


packages/@stylexjs/inline-css/src/index.d.ts
```diff
@@ -0,0 +1,16 @@
+import type { StyleXClassNameFor, StyleXStyles } from '@stylexjs/stylex';
+import type { Properties } from 'csstype';
+
+type InlineValue<V> = {
+  [Key in string | number]: StyleXClassNameFor<string, V>;
+} & ((value: V) => StyleXStyles);
+
+type InlineCSS = {
+  [Key in keyof Properties<string | number>]: InlineValue<
```
henryqdineen: It looks like everything in Properties is optional and the `Required` utility type can help.

```suggestion
  [Key in keyof Required<Properties<string | number>>]: InlineValue<
```

It would be cool if  `StyleXCSSTypes.js` also used `csstype` or at least the same thing.

packages/@stylexjs/atoms/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+const valueProxy = (_propName) =>
+  new Proxy(function () {}, {
+    get() {
+      return valueProxy('');
+    },
+    apply() {
+      return undefined;
+    },
+  });
```
nmn: It should probably not be a recursive proxy and the `get()` should throw an error similar to the stylex package saying you need to compile this away.

packages/@stylexjs/atoms/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+const inlineCSS = new Proxy(function () {}, {
+  get(_target, prop) {
+    if (typeof prop === 'string') {
+      return valueProxy(prop);
+    }
+    return undefined;
```
nmn: You can probably just throw here directly and not define `valueProxy` at all.

nmn: Also, let's add Flow types to this and auto-generate the TS types from that.

packages/@stylexjs/babel-plugin/src/visitors/imports.js
```diff
@@ -22,6 +24,28 @@ export function readImportDeclarations(
     return;
   }
   const sourcePath = node.source.value;
+
+  if (INLINE_CSS_SOURCES.has(sourcePath)) {
+    for (const specifier of node.specifiers) {
+      if (specifier.type === 'ImportNamespaceSpecifier') {
+        state.inlineCSSImports.set(specifier.local.name, '*');
+      } else if (specifier.type === 'ImportDefaultSpecifier') {
+        state.inlineCSSImports.set(specifier.local.name, '*');
```
nmn: ESM doesn't allow an arbitrary proxy for exports from a package. So maybe we ONLY allow a named or default import here?

I think limiting this to a named import makes the most sense right now.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-props-test.js
```diff
@@ -231,6 +231,451 @@ describe('@stylexjs/babel-plugin', () => {
       });
     });
 
+    describe('props calls with inline-css', () => {
+      test('inline static styles match stylex.create', () => {
+        const inline = transform(`
+          import stylex from 'stylex';
+          import * as css from '@stylexjs/inline-css';
+          stylex.props(css.display.flex);
```
nmn:
```suggestion
          import {css} from '@stylexjs/inline-css';
          stylex.props(css.display.flex);
```

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-props-test.js
```diff
@@ -231,6 +231,451 @@ describe('@stylexjs/babel-plugin', () => {
...
+            priority: 3000
+          });
+          _inject2({
+            ltr: "@property --x-color { syntax: \\"*\\"; inherits: false;}",
+            priority: 0
+          });
+          stylex.props([{
+            "color": color != null ? "x14rh7hd" : color,
+            "$$css": true
+          }, {
```
nmn: It is important that that the style objects created by the inline-css package are hoisted to the module level as a constant and not inlined like this. We already have a utility for doing this somewhere.

nmn (dismissed):
Good start, but a few changes are needed.

packages/@stylexjs/inline-css/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+  },
+  apply() {
+    throw new Error(
+      '@stylexjs/inline-css is a compile-time helper. Attempted to call it as a function, but the StyleX compiler did not run.',
+    );
+  },
+});
+
+module.exports = inlineCSS;
+module.exports.default = inlineCSS;
```
henryqdineen: I feel like it's rare at this point to see frontend code be published CJS-only. Could this be written in ESM and then use babel at build time to create a CJS version?

I have a similar nitpick about styleq but I see there is [an open PR for that](https://github.com/necolas/styleq/pull/12). One advantage is that it would make the @stylexjs/stylex rollup bundle slightly smaller.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-props-test.js
```diff
@@ -231,6 +231,501 @@ describe('@stylexjs/babel-plugin', () => {
...
+      test('inline static with inline dynamic', () => {
+        const output = transform(`
+          import stylex from 'stylex';
+          import * as css from '@stylexjs/utility-styles';
+          stylex.props(css.display.flex, css.color(color));
+        `);
+        expect(output).toContain('.x78zum5{display:flex}');
+        expect(output).toContain('.x14rh7hd{color:var(--x-color)}');
+        expect(output).toContain('--x-color');
+        expect(output).toContain('_v != null ? "x14rh7hd" : _v');
```
nmn: Can we stick to snapshot tests here? You can keep these checks in addition if you want to, but I think this is a bad pattern than LLMs generate.

packages/@stylexjs/babel-plugin/__tests__/transform-stylex-props-test.js
```diff
@@ -231,6 +231,501 @@ describe('@stylexjs/babel-plugin', () => {
...
+          };
+          _inject2({
+            ltr: ".x78zum5{display:flex}",
+            priority: 3000
+          });
+          const _temp = {
+            k1xSpc: "x78zum5",
+            $$css: true
+          };
+          stylex.props(_temp, styles.opacity(0.5));"
```
nmn, line 728: Can you add a few more tests with this kind of pattern. I want to see what happens to inline styles when applied conditionally but `stylex.props` cannot be pre-compiled more.

Also, to validate the hoisting logic, can you make add some examples where the `stylex.props()` exists inside of a function and not the top level.

packages/@stylexjs/babel-plugin/__tests__/validation-stylex-create-test.js
```diff
@@ -10,7 +10,7 @@
 jest.autoMockOff();
 
 import { transformSync } from '@babel/core';
-import { messages } from '../src/shared';
+import { messages } from '@stylexjs/shared';
```
nmn: Did you just decide to move things back to the `shared` package?

packages/@stylexjs/babel-plugin/rollup.config.mjs
```diff
@@ -46,7 +46,7 @@ const config = {
     ? external
     : [...external, '@dual-bundle/import-meta-resolve', '@stylexjs/stylex'],
   plugins: [
-    babel({ babelHelpers: 'bundled', extensions, include: ['./src/**/*'] }),
+    babel({ babelHelpers: 'bundled', extensions, include: ['./src/**/*', '../shared/src/**/*'] }),
```
nmn: This should not be needed. The shared package itself should be transformed so no need to run it through Babel.

(Maybe we should move the refactor work to a separate PR?)

packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -165,12 +166,20 @@ export default class StateManager {
   +stylexViewTransitionClassImport: Set<string> = new Set();
   +stylexDefaultMarkerImport: Set<string> = new Set();
   +stylexWhenImport: Set<string> = new Set();
+  // Map of local identifier -> imported name.
+  // For namespace/default imports we store '*'.
+  +inlineCSSImports: Map<string, string> = new Map();
```
nmn: let's discuss the right naming here.

packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -165,12 +166,20 @@ export default class StateManager {
...
 
   injectImportInserted: ?t.Identifier = null;
 
   // `stylex.create` calls
   +styleMap: Map<string, CompiledNamespaces> = new Map();
   +styleVars: Map<string, NodePath<>> = new Map();
+  +inlineStylesCache: Map<string, FlatCompiledStyles> = new Map();
```
nmn: What is this? Why is it needed?

packages/@stylexjs/babel-plugin/src/visitors/imports.js
```diff
@@ -12,6 +12,8 @@ import type { NodePath } from '@babel/traverse';
 import * as t from '@babel/types';
 import StateManager from '../utils/state-manager';
 
+const UTILITY_STYLES_SOURCES = new Set(['@stylexjs/utility-styles']);
```
nmn: `/style-properties`?

packages/@stylexjs/atoms/src/babel-transform.js
```diff
@@ -0,0 +1,249 @@
...
+        transformDynamicStyle(path, dynamicStyle);
+      }
+    },
+  };
+}
+
+function isUtilityStylesIdentifier(ident, state, path) {
+  if (state.atomImports && state.atomImports.has(ident.name)) {
+    return true;
+  }
```
vincentriemer: I think if the import's name is shadowed (i.e. you import the name `css`, but you're in a function that has an argument named `css`) it would still cause the replacement to happen.

vincentriemer (approved):

stephenh:
(Edit: for those interested, I have a [WIP PR](https://github.com/homebound-team/truss/pull/132) where I thought I could get away with codegen-ing `stylex.create`s into our `Css.ts` file, but after getting it working surprisingly well, I unfortunately realized that selectors fundamentally need their own per-file/per-usage `stylex.create` calls, so I'll have to reset this approach.)

Fwiw when this PR is released, I'm hoping to port our [truss](https://github.com/homebound-team/truss) "DSL on top of emotion" over to this/stylex, by doing things like (very proof-of-concept):

```ts
// in our Css.ts file
class CssBuilder {
  styles = {}
  get df() {
    return new CssBuilder({ ...this.styles, ...x.display.flex });
  }
  get aic() {
    return new CssBuilder({ ...this.styles, ...x.alignItems.center });
  }
  get $() {
    return this.styles;
  }
}
export const Css = new CssBuilder();

// in a Component.tsx file
// use/write a `css` prop that calls `stylex.props(...)` on the hash from $
return <div css={Css.df.aic.$} />

// Hope that this is semantically the same as the PR's
return <div {...stylesx.props(x.display.flex, x.alignItems.center)} />
```

With the admittedly naive hope/assumption that each `x.display.etc` (and other atoms in the `Css.ts` builder) will be rewritten into top-level/build-time stylex.creates or what not, and after our `CssBuilder` accumulates them on each `.df`, `.aic`, etc call, we can just do a `styles.props(...)` on the resulting hash & get the appropriate css classes (or close enough).

Feel free to ignore me, but if you have any "that just might work" / "ofc that will obviously not work" / etc feedback, I'd appreciate it -- thank you!

PR #1409, opened 2025-12-23
[docs] Playground: compress url-state

nmn:

- [x] Use `lz-string` to compress the URL state
- [x] Ensure that the existing URLs keep working
  - [x] Automatically convert old URLs to new URLs when opened
  - [x] Ensure that the initial preview shows URL state from old URLs

Extra benefit: The new URLs are generally more compatible with social networks and are more portable.

### DIFF
--- a/packages/docs/components/PlaygroundNew.js
+++ b/packages/docs/components/PlaygroundNew.js
@@ -92,12 +92,6 @@ function transformSourceFiles(sourceFiles) {
   return { transformedFiles, generatedCSS };
 }
 
-// const encodeObjKeys = (obj) => {
-//   return Object.fromEntries(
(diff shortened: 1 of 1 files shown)


packages/docs/components/PlaygroundNew.js
```diff
@@ -83,22 +92,40 @@ function transformSourceFiles(sourceFiles) {
   return { transformedFiles, generatedCSS };
 }
 
-const encodeObjKeys = (obj) => {
-  return Object.fromEntries(
-    Object.entries(obj).map(([key, value]) => [key, encodeURIComponent(value)]),
-  );
-};
+// const encodeObjKeys = (obj) => {
```
mellyeliu: rm?

Issue #1412, opened 2025-12-30
`npm install` and `npm run build` fails on Windows due to cp usage


Noname01010101:
### Describe the issue

build script fails on windows:
```powershell
npm run build
```

It fails due to the `cp` command, which seems to be available only in unix systems.

Error log: 
``` powershell
> @stylexjs/stylex@0.17.4 prebuild
> rimraf lib && gen-types -i src/ -o lib/cjs && cp -r lib/cjs lib/es

Successfully generated type definitions
'cp' is not recognized as an internal or external command,
operable program or batch file.
npm error Lifecycle script `build` failed with error:
npm error code 1
npm error path D:\projects\forks\stylex\packages\@stylexjs\stylex
npm error workspace @stylexjs/stylex@0.17.4
npm error location D:\projects\forks\stylex\packages\@stylexjs\stylex
npm error command failed
npm error command C:\WINDOWS\system32\cmd.exe /d /s /c rimraf lib && gen-types -i src/ -o lib/cjs && cp -r lib/cjs lib/es
```

### Expected behavior

Correctly complete prebuilds with cross-platform tools, instead of platform dependent commands.

### Steps to reproduce

1. create a windows 11 VM with docker (if running on linux)
2. run `npm install` or `npm run build` on root.

### Additional comments

Solution seems to be very simple: 
1. maybe use a node copy command

PR #1413, opened 2025-12-30
fix: cross-platform support for scripts that use `cp` etc.

rayterion:

instead of using linux's `cp`, i replaced this part of the command with a simple js file that does the same thing with node, for cross-platform support.

## Linked PR/Issues

- #1412

## Additional Context

`npm run build` now works on windows 11:
image_2025-12-30_115933910

## Pre-flight checklist

- [yes] I have read the contributing guidelines
- [yes] Performed a self-review of my code

### DIFF
--- a/examples/example-nextjs/package.json
+++ b/examples/example-nextjs/package.json
@@ -3,7 +3,7 @@
   "name": "example-nextjs",
   "version": "0.17.4",
   "scripts": {
-    "clean": "rimraf .next",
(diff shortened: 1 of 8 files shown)


rayterion:
I will add tests and check if the implementation is correct before reopening

rayterion:
Fixed the command by adding a more abstract `copy-from-to` into the `scripts/` folder and adding fail-safes

PR #1419, opened 2026-01-02
Fix/website style fixes

nmn:

- [x] Long headings of code should wrap on mobile now. (break-word)
- [x] CodeBlocks can now scroll horizontally if they don't fit
- [x] The typing word animation on Firefox has been fixed
  - Firefox doesn't seem to be applying the `display:none` as part of a keyframe animation
  - This has been fixed by also animation `opacity` and `font-size`.
  - Now the animation works correctly in all three browser engines.

### DIFF
--- a/packages/docs/src/components/TypingWord.tsx
+++ b/packages/docs/src/components/TypingWord.tsx
@@ -60,18 +60,22 @@ const typingAnim = stylex.keyframes({
 const hidden = stylex.keyframes({
   '0%': {
     display: 'inline',
+    fontSize: '1em',
     opacity: 1,
   },
   '20%': {
     display: 'inline',
+    fontSize: '1em',
     opacity: 1,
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/mdx/CodeBlock.tsx
```diff
@@ -122,14 +122,14 @@ export function CodeBlock({
...
         role="region"
         tabIndex={0}
-        {...stylex.props(styles.viewport, !title && styles.viewportPadded)}
-        style={{
-          counterSet: props['data-line-numbers']
-            ? `line ${Number(props['data-line-numbers-start'] ?? 1) - 1}`
-            : undefined,
-          ...viewportProps.style,
-        }}
+        // style={{
```
mellyeliu, line 125: rm

PR #1420, opened 2026-01-02
[playground] redesign

mellyeliu:
## Context 

Overhaul of the playground to improve the base example + some functionality + redesign. We can continue to iterate on the final design as we go.

See deploy [here](https://stylex-cw2ygugnz-fbopensource.vercel.app/playground).

image

- [x] Playground redesign
- [x] Improved default example
- [x] Double click and longpress to rename file
- [x] Combine JS and CSS panels for now
- [x] x button to delete file + more descriptive delete panel message
- [x] Fix light mode code editor theme bug 
- [x] Color tweaks for docs
- [x] Fix code block heights
- [x] Fix rename resizing
- [x] Cleanup and refactor

## Testing
- [x] File rename, delete works as before
- [x] Playground works as expected

### DIFF
--- a/packages/docs/src/components/Playground/index.tsx
+++ b/packages/docs/src/components/Playground/index.tsx
@@ -774,6 +774,8 @@ const styles = stylex.create({
     containerType: 'inline-size',
   },
   frame: {
+    // eslint-disable-next-line @stylexjs/valid-styles
+    cornerShape: 'squircle',
     boxSizing: 'border-box',
     width: '100%',
     height: '100%',
@@ -784,7 +786,6 @@ const styles = stylex.create({
     borderStyle: 'solid',
     borderWidth: 1,
     borderRadius: 20,
-    cornerShape: 'squircle',
   },
   row: {
     display: 'flex',
--- a/packages/docs/src/components/mdx/Details.tsx
+++ b/packages/docs/src/components/mdx/Details.tsx
@@ -95,14 +95,17 @@ const DURATION = '0.3s';
 const EASING = 'cubic-bezier(0.4, 0, 0.2, 1)';
 const styles = stylex.create({
   group: {
+    // eslint-disable-next-line @stylexjs/valid-styles
+    cornerShape: 'squircle',
     overflow: 'hidden',
     borderColor: vars['--color-fd-border'],
     borderStyle: 'solid',
     borderWidth: 1,
     borderRadius: 16,
-    cornerShape: 'squircle',
   },
   container: {
+    // eslint-disable-next-line @stylexjs/valid-styles
+    cornerShape: 'squircle',
     interpolateSize: 'allow-keywords',
     paddingBlock: 8,
     paddingInline: 16,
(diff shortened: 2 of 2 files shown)


packages/docs/src/components/Footer.tsx
```diff
@@ -206,6 +212,10 @@ const styles = stylex.create({
     transitionDuration: '300ms',
     transitionProperty: 'background-color, border-color',
   },
+  footerNoBorder: {
+    borderTopStyle: 'none',
+    borderTopWidth: 0,
+  },
```
nmn, line 219:
```suggestion
  footerNoBorder: {
    borderTopColor: null,
    borderTopStyle: null,
    borderTopWidth: null,
  },
```

packages/docs/src/theming/vars.stylex.ts
```diff
@@ -152,8 +152,8 @@ export const vars = stylex.defineVars({
   ),
   '--color-fd-accent': lightDark('hsl(223, 16%, 83%)', 'hsl(237, 16%, 23%)'),
   '--color-fd-accent-foreground': lightDark(
-    'hsl(248, 58.70%, 62.90%)',
-    'hsl(226, 70.70%, 70.60%)',
+    ' hsl(291, 66.30%, 67.50%)',
+    'hsl(226, 66.30%, 67.50%)',
   ),
```
nmn, line 155: The same color for light and dark is probably not what we want.

packages/docs/src/theming/vars.stylex.ts
```diff
@@ -200,9 +200,24 @@ export const EASINGS = stylex.defineConsts({
...
-  '--pg-border': lightDark('hsla(0, 0%, 80%, 50%)', 'hsla(0, 0%, 40%, 20%)'),
+  '--pg-border': lightDark('hsla(0, 0%, 80%, 30%)', 'hsla(0, 0%, 40%, 20%)'),
   '--pg-header-fg': lightDark('hsl(0, 0%, 3.9%)', 'hsl(0, 0%, 98%)'),
   '--pg-preview': lightDark('hsl(0, 0%, 94.7%)', '#222'),
+  '--pg-panel-surface': lightDark('#ffffff', '#0b0b0f'),
+  '--pg-header-surface': lightDark('#ffffff', '#1a1a1a'),
+  '--pg-header-shadow': lightDark(
+    `inset 0 0 18px color-mix(in srgb, ${vars['--color-fd-primary']} 30%, transparent)`,
+    'none',
+  ),
```
nmn: light-dark doesn't work for non color values.

nmn:
light-mode seems to switch the pink and purple colors, so probably not what we want to land on, but I can see the way the pink is overused in light mode right now.

mellyeliu:
I preferred the indigo over the pink because I found it too loud, the swap is intentional.

packages/docs/src/theming/vars.stylex.ts
jcperez-ch: I wonder what was the motivation for this change.

jcperez-ch:
Regarding colors, are we making sure they meet the color contrast ratio between foreground and background? I don't want this change to back fire us because of this little accessibility norm.

nmn:
@jcperez-ch There are a handful of colors that don't have enough contrast in light mode. I'll tweak the colors more.

PR #1422, opened 2026-01-03
[eslint] Improve shorthand expansion and add support for corner-shape

nmn:

The eslint rule for `valid-shorthands` was using an old and outdated package that wouldn't expand many shorthands as expected. It has been replaced by a custom implementation based on `@csstools/css-tokenizer`

Support for `corner-shape` has been added

Expansion of border properties has been improved.

### DIFF
--- a/packages/docs/src/components/mdx/Callout.tsx
+++ b/packages/docs/src/components/mdx/Callout.tsx
@@ -40,10 +40,11 @@ export function Callout({ children, title, ...props }: CalloutProps) {
   );
 }
 
-export interface CalloutContainerProps extends Omit<
-  HTMLAttributes<HTMLDivElement>,
-  'className' | 'style' | 'title'
-> {
+export interface CalloutContainerProps
+  extends Omit<
+    HTMLAttributes<HTMLDivElement>,
+    'className' | 'style' | 'title'
+  > {
   /**
    * @defaultValue info
    */
@@ -93,10 +94,8 @@ export function CalloutContainer({
   );
 }
 
-export interface CalloutTitleProps extends Omit<
-  HTMLAttributes<HTMLParagraphElement>,
-  'className' | 'style'
-> {
+export interface CalloutTitleProps
+  extends Omit<HTMLAttributes<HTMLParagraphElement>, 'className' | 'style'> {
   children: ReactNode;
 }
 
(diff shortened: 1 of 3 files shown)


packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
```diff
@@ -962,17 +1019,17 @@ eslintTester.run('stylex-valid-shorthands', rule.default, {
       const styles = stylex.create({
         main: {
           borderTopWidth: '4px',
-          borderRightWidth: '5px',
+          borderInlineEndWidth: '5px',
```
mellyeliu: These shouldn't be changing. We need to map these to respect the `preferInline` config like we did for `borderRadius` below

packages/@stylexjs/eslint-plugin/src/utils/splitShorthands.js
```diff
@@ -115,102 +116,578 @@ const toCamelCase = (str: string) => {
...
+  'smaller',
+  'larger',
+]);
+const COLOR_KEYWORDS = new Set(['transparent', 'currentcolor']);
+const COLOR_FUNCTION_REGEX =
+  /^(?:rgb|rgba|hsl|hsla|hwb|hsb|lab|lch|oklab|oklch|color)\(/i;
+const IMAGE_FUNCTION_REGEX =
+  /^(?:url|image-set|linear-gradient|radial-gradient|conic-gradient|repeating-linear-gradient|repeating-radial-gradient|repeating-conic-gradient|cross-fade|element)\(/i;
+const LENGTH_FUNCTION_REGEX = /^(?:calc|min|max|clamp)\(/i;
+const LENGTH_REGEX = /^-?(?:\d+|\d*\.\d+)(?:[a-z%]+)?$/i;
```
mellyeliu: this seems too permissive?

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
nmn: I saw the setting being ignored so I ignored it, missed that it was only sometimes being ignored.

PR #1424, opened 2026-01-03
feat(babel-plugin): add disallowedPropertiesValidation option

### LINKED ISSUE #1423: [babel-plugin] Soft validation for disallowed properties
We currently error on a lot of (relatively) innocuous properties like `border` and `background` for XPR reasons in `property-specificity`. Let's add a config with something like 'log' | 'throw' | 'silent'

Steps:
1. Add config in https://github.com/facebook/stylex/blob/main/packages/%40stylexjs/babel-plugin/src/utils/state-manager.js
2. Configure errors appropriately in https://github.com/facebook/stylex/blob/main/packages/%40stylexjs/babel-plugin/src/shared/preprocess-rules/property-specificity.js#L21-L90

Faizanq:
Add config option to control validation behavior for disallowed shorthand properties (border, background, animation, etc).

## Changes
- Added `disallowedPropertiesValidation` option with three modes:
  - `'throw'` (default): Throws error for disallowed properties
  - `'warn'`: Logs warning and skips property
  - `'silent'`: Silently skips property

## Use case
Allows gradual migration of existing codebases that use disallowed shorthand properties without breaking the build.

Fixes #1423

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/index.js
+++ b/packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/index.js
@@ -31,7 +31,7 @@ export default function flatMapExpandedShorthands(
   objEntry: $ReadOnly<[string, TStyleValue]>,
   options: $ReadOnly<{
     styleResolution: StyleXOptions['styleResolution'],
-    disallowedPropertiesValidation?: StyleXOptions['disallowedPropertiesValidation'],
+    propertyValidationMode?: StyleXOptions['propertyValidationMode'],
     ...
   }>,
 ): $ReadOnlyArray<[string, TStyleValue]> {
@@ -54,13 +54,11 @@ export default function flatMapExpandedShorthands(
     try {
       return expansion(value);
     } catch (error) {
-      const validationMode = options.disallowedPropertiesValidation ?? 'throw';
+      const validationMode = options.propertyValidationMode ?? 'silent';
       if (validationMode === 'throw') {
         throw error;
       } else if (validationMode === 'warn') {
-        console.warn(
-          `[stylex] ${error.message} Property "${key}" will be skipped.`,
-        );
+        console.warn(`[stylex] ${error.message}`);
         return [];
       } else {
         // silent mode - skip the property without any output
--- a/packages/@stylexjs/babel-plugin/src/shared/utils/default-options.js
+++ b/packages/@stylexjs/babel-plugin/src/shared/utils/default-options.js
@@ -13,7 +13,7 @@ export const defaultOptions: StyleXOptions = {
   classNamePrefix: 'x',
   dev: false,
   debug: false,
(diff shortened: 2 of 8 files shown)


nmn (approved):
This looks good for now. We should update the babel plugin to never throw on these properties by default, and create a new option to manually validate for React Native instead.

packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/index.js
```diff
@@ -50,7 +51,22 @@ export default function flatMapExpandedShorthands(
...
-    return expansion(value);
+    try {
+      return expansion(value);
+    } catch (error) {
+      const validationMode = options.disallowedPropertiesValidation ?? 'throw';
+      if (validationMode === 'throw') {
+        throw error;
+      } else if (validationMode === 'warn') {
+        console.warn(
+          `[stylex] ${error.message} Property "${key}" will be skipped.`,
```
mellyeliu: These properties aren't skipped, they're just discouraged for RN compatibility/resolution edge cases in some modes. Probably fine to `console.warn` the error directly here.

packages/@stylexjs/babel-plugin/src/shared/utils/default-options.js
```diff
@@ -13,6 +13,7 @@ export const defaultOptions: StyleXOptions = {
   classNamePrefix: 'x',
   dev: false,
   debug: false,
+  disallowedPropertiesValidation: 'throw',
```
mellyeliu: We can probably log in dev mode. Throwing by default is too noisy.

```suggestion
  propertyValidationMode: 'silent',
```

packages/@stylexjs/babel-plugin/src/utils/state-manager.js
```diff
@@ -339,6 +340,15 @@ export default class StateManager {
...
 
+    const disallowedPropertiesValidation: StyleXStateOptions['disallowedPropertiesValidation'] =
+      z.logAndDefault(
+        z.unionOf3(z.literal('throw'), z.literal('warn'), z.literal('silent')),
+        options.disallowedPropertiesValidation ??
+          defaultOptions.disallowedPropertiesValidation,
+        'throw',
+        'options.disallowedPropertiesValidation',
+      );
+
```
mellyeliu:
```suggestion
const propertyValidationMode: StyleXStateOptions['propertyValidationMode'] =
  z.logAndDefault(
    z.unionOf3(z.literal('throw'), z.literal('warn'), z.literal('silent')),
    options.propertyValidationMode ??
      (dev ? 'warn' : 'silent'),
    dev ? 'warn' : 'silent',
    'options.propertyValidationMode',
  );

```

mellyeliu:
We might want to log in dev or for RN but can follow up with a patch

PR #1426, opened 2026-01-03
[docs] misc edits

henryqdineen:

This PR includes a couple edits I made while re-reading the docs:
- I noticed that `genConditionalClasses` was still used in a couple places so I updated that. 
- I added `<DevInstallExample>` in the unplugin docs for consistency.
- I used ES6 object property shorthand syntax in the PostCSS example code.

On the subject of consistency: Does anyone have a strong preference on if references to functions like `create` and `props` should be written like `create()` and `props()`? The latter is used in the "Introduction" but the former is used in most other places. I think I prefer with the parenthesis. 

## Pre-flight checklist


mellyeliu (approved):
> On the subject of consistency: Does anyone have a strong preference on if references to functions like create and props should be written like create() and props()?

This is something we attempted to consolidate a few times. I don't have a strong preference but consistency would be nice, for this as well as the `stylex.` prefix. The parentheses read a bit clearer to me.

PR #1430, opened 2026-01-04
[docs] Rewrite tabs component to be RSC compatible

nmn:

The `<Tabs>` component wasn't working when used with RSCs as React renders children as lazy components. So it has been rewritten to not be dependent on reading the `children` prop.

### DIFF
--- a/packages/docs/content/docs/learn/thinking-in-stylex.mdx
+++ b/packages/docs/content/docs/learn/thinking-in-stylex.mdx
@@ -62,8 +62,8 @@ let a = stylex.props(styles.red);
 
 Compiles down to:
 
-<Accordions>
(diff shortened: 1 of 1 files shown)


packages/docs/content/docs/learn/thinking-in-stylex.mdx
```diff
@@ -149,8 +147,8 @@ function MyComponent({ style }) {
 }
 ```
 
-</Accordion>
-</Accordions>
```
mellyeliu, line 153: Are we still using this anywhere? Rm?

nmn: Yes. We are using accordians in other places, although not as groups like here.

PR #1437, opened 2026-01-05
[stylex] enable disableMix for styleq instance

henryqdineen:

I was reading the `react-strict-dom` codebase and noticed that they set `disableMix: true` on their styleq instance: https://github.com/facebook/react-strict-dom/blob/a5cbf07a03cff88dac6682b4409c1ad49e723d65/packages/react-strict-dom/src/web/css/merge.js#L32-L38

I have no real concerns about the runtime performance of `stylex.props()` but, unless I'm missing something obvious, this seems like an easy win. By default, styleq removes CSS classes for any properties that are also set by inline styles, but StyleX hashes the property keys so the comparisons are pointless. Setting `disableMix: true` will just bypass this logic. 

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/stylex/src/stylex.js
+++ b/packages/@stylexjs/stylex/src/stylex.js
@@ -56,6 +56,10 @@ export type {
 
 import { styleq } from 'styleq';
 
+const optimizedStyleq = styleq.factory({
+  disableMix: true,
+});
+
 const errorForFn = (name: string) =>
   new Error(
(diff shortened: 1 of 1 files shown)


nmn:
>  StyleX hashes the property keys 

This is behind an option and something we may disable. The hashing didn't have the size wins we expected and it creates complexity in other ways.

henryqdineen:
Ah I just learned about `enableMinifiedKeys`. Thanks!

PR #1449, opened 2026-01-13
Support attribute selectors in stylex.when and conditional styles

vincentriemer:
"Styling-agnostic" react component systems, such as base-ui, expose state based styling hooks through the use of dom element attributes. Actually listening to these in StyleX, while possible through usage of the `:is()` psedudoselector, is overly verbose and uncessary. This diff addresses this by adding support for attribute selectors in stylex.when calls and conditional styles.

### Motivation

- Allow `stylex.when` and conditional style objects to accept attribute selectors (including value-matching forms) rather than only attribute-existence markers.
- Treat attribute selectors like pseudo-classes in the preprocessing and validation pipeline so downstream CSS generation remains correct.
- Surface attribute-selector usage in types and docs so consumers get proper editor/compile-time guidance.

### Description

- Accept keys starting with `[` in validation and preprocessing by updating `basic-validation.js`, `flatten-raw-style-obj.js`, and `PreRule.pseudos` to treat attribute selectors like pseudos.
- Extend `when` helpers to accept attribute selectors by adding a `WhenSelector` union, updating `validatePseudoSelector` to allow `[`...`]`, and applying the selector when building `:where` / `:has` clauses in `when.js`.
- Update Flow/TypeScript definitions in `StyleXTypes.js` and `StyleXTypes.d.ts` to permit attribute selectors for `stylex.when.*` signatures.
- Add/update tests and documentation to use value-based attribute selectors (tests and snapshots in `transform-stylex-when-test.js` and `flatten-raw-style-obj-test.js`, and docs in `when.mdx`).

### Testing

- Ran `yarn workspace @stylexjs/babel-plugin test -- transform-stylex-when-test --updateSnapshot`, where the test suite passed and snapshots were updated but the overall command returned non-zero due to global coverage thresholds being unmet.
- The changed test file `transform-stylex-when-test.js` passed (all tests green) and snapshots were updated successfully.
- Ran `npm run lint`, which completed successfully.
- Pre-commit formatting/lint hooks ran as part of the commit and completed successfully.

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/__tests__/flatten-raw-style-obj-test.js
+++ b/packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/__tests__/flatten-raw-style-obj-test.js
@@ -483,6 +483,27 @@ describe('Flatten Style Object with legacy shorthand expansion', () => {
         ],
       ]);
     });
+    test('Attribute selector conditions', () => {
+      expect(
+        flattenRawStyleObject(
+          {
+            color: {
+              default: 'blue',
+              '[data-panel-state="open"]': 'red',
+            },
+          },
+          options,
+        ),
+      ).toEqual([
+        [
+          'color',
+          PreRuleSet.create([
+            new PreRule('color', 'blue', ['color', 'default']),
+            new PreRule('color', 'red', ['color', '[data-panel-state="open"]']),
+          ]),
+        ],
+      ]);
+    });
   });
   describe('Multiple levels of nesting', () => {
     test('Fallback styles within nested objects', () => {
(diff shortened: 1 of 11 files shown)


packages/@stylexjs/babel-plugin/src/shared/preprocess-rules/__tests__/flatten-raw-style-obj-test.js
```diff
@@ -483,6 +483,27 @@ describe('Flatten Style Object with legacy shorthand expansion', () => {
         ],
       ]);
     });
+    test('Attribute selector conditions', () => {
```
mellyeliu, line 486: could we add a test for nested attribute selectors combined with pseudo-classes (ie. `':hover': { '[data-state="open"]': ... }` and the inverse) to check for the same generated class? pretty sure we sort but would be good to capture explicitly

PR #1454, opened 2026-01-15
Add support for Svelte

srikrsna:

The unplugin doesn't work with Svelte out of the box. This PR adds support for Svelte along with an example of using the unplugin in a sveltekit project.

## Pre-flight checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -15,3 +15,4 @@ examples/example-cli/src
 packages/docs/.source
 devtools-extension/extension
 **/devtools-extension/extension/assets/**
+examples/example-sveltekit/.svelte-kit/**


examples/example-sveltekit/src/routes/+page.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts" module>
+  import * as stylex from '@stylexjs/stylex';
+  import { attrs } from '$lib';
```
henryqdineen: I don't see an `attrs` function in this PR? I'm guessing your `lib` directory got ignored because of the `.gitignore` in the root of the repo

PR #1478, opened 2026-02-09
[create-stylex-app] Improve CLI launcher

abhakat:
## Summary
- Improves the CLI launcher for `create-stylex-app`
- Updates package.json and dependency lockfile
- We basically want to make it easier to cd/run
image

- Manually running is clunky so I wanted to streamline it.
- Fix the "Permission denied" error that occurred when running npx create-stylex-app — Babel's build output doesn't preserve the executable bit, so the CLI entry point needs chmod +x after each build.      

## Test plan
```
cd packages/@stylexjs/create-stylex-app                                                                                  │
npm run build
```
Testing on these 3 templates:

  1. nextjs — Tests the most common/default path
                                                                                                                                            
This was the biggest change:
image

  2. vite-react — Tests the usesSharedUI branch
  - When usesSharedUI is true, fetchTemplate also calls fetchSharedUI() which downloads packages/shared-ui into a shared-ui/ subdirectory (fetch-template.js:43-45)
  - 6 of the 13 templates use this path, so it's worth covering

image

  3. webpack — Tests a different bundler with different scripts

image

4. esbuild and cli dont have dev scripts

Its kinda misleading but not sure if its important enough?
image

2/10: after addressing Melissa's suggestions: 
image

## Additional Context

Ran and updated tests.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/create-stylex-app/package.json
+++ b/packages/@stylexjs/create-stylex-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-stylex-app",
-  "version": "0.17.6-alpha.2",
+  "version": "0.17.6",
   "description": "Scaffold a new StyleX project from official templates",
   "repository": "https://www.github.com/facebook/stylex",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


packages/@stylexjs/create-stylex-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-stylex-app",
-  "version": "0.17.6-alpha.2",
```
mellyeliu, line 3: let's wait until we bump everything at once in the next patch

PR #1479, opened 2026-02-09
[examples] Modernize example-nextjs template with theme switcher

abhakat:
## Summary
- Modernizes the example-nextjs template with a theme switcher
- Replaces `Counter.tsx` with `InteractiveCard.tsx`
- Updates global tokens, layout, page, and card components

## Test plan
- Ran the example-nextjs app and verify the theme switcher works
- Verified all theme variants (red, green, blue) apply correctly
- Check that the interactive card renders and functions properly

Before:
image

After:
image

## Pre-flight checklist

### DIFF
--- a/examples/example-nextjs/app/InteractiveCard.tsx
+++ b/examples/example-nextjs/app/InteractiveCard.tsx
@@ -37,17 +37,24 @@ export const blueTheme = stylex.createTheme(colors, {
 });
 
 export const THEMES = [
-  { name: 'Red', theme: redTheme, dotColor: '#e03131' },
-  { name: 'Green', theme: greenTheme, dotColor: '#2f9e44' },
-  { name: 'Blue', theme: blueTheme, dotColor: '#1c7ed6' },
+  { name: 'Red', theme: redTheme },
+  { name: 'Green', theme: greenTheme },
+  { name: 'Blue', theme: blueTheme },
 ] as const;
 
 type Props = Readonly<{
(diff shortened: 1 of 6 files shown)


examples/example-nextjs/app/InteractiveCard.tsx
```diff
@@ -0,0 +1,236 @@
...
+}>;
+
+export default function InteractiveCard({ themeIndex, onThemeChange }: Props) {
+  const [count, setCount] = useState(0);
+
+  return (
+    <div {...stylex.props(styles.card)}>
+      <div {...stylex.props(styles.label)}>Theme</div>
+      <div {...stylex.props(styles.themeRow)}>
+        {THEMES.map((t, i) => (
```
mellyeliu, line 64: Looking at this code, I don't actually see the themes being used. `t.theme` is not referenced.

mellyeliu (commented):
Looks awesome! Please add the dark mode equivalent in the test plan and fix the theming logic

examples/example-nextjs/app/InteractiveCard.tsx
mellyeliu: I see it's exported and used in another file. Let's add the dot color into the theme instead of using the conditionals.

abhakat: fixed, hopefully

mellyeliu (approved):

PR #1488, opened 2026-02-22
RFC: `stylex.namedVar` for ergonomic custom property names in `defineVars`

vincentriemer:
RFC: `stylex.namedVar` for ergonomic custom property names in `defineVars`

### Summary
This RFC introduces a new compile-time helper:

```ts
stylex.namedVar(name: string, value: DefineVarValue)
```

It lets developers keep ergonomic dot-notation keys in `defineVars` while explicitly choosing the emitted CSS custom property name.

Example:

```ts
export const vars = stylex.defineVars({
  surfaceBg: stylex.namedVar('--surface-bg', 'black'),
  textPrimary: stylex.namedVar('--text-primary', 'white'),
});
```

This preserves usage ergonomics (`vars.surfaceBg`) and emits explicit custom property names (`--surface-bg`, `--text-primary`).

### Motivation
Current `defineVars` patterns force a tradeoff:

1. Hash-based var names are ergonomic but not explicit.
2. Legacy explicit custom names via `'--foo'` keys are explicit but less ergonomic for dot-notation authoring.

We want both explicit naming and ergonomic authoring, with hard-fail diagnostics when misconfigured.

### Goals
1. Keep dot-notation variable references ergonomic.
2. Allow explicit custom property names in `defineVars`.
3. Fail loudly and actionably on invalid declarations.
4. Preserve cross-file correctness for imported theme references.
5. Keep legacy `'--foo'` behavior fully supported.

### Non-goals
1. No runtime execution semantics for `namedVar`.
2. No auto-prefixing missing `--`.
3. No relaxation of static-analysis constraints.

---

### New API
```ts
stylex.namedVar(name, value)
```

`name` must be:

1. A static string literal at compile time.
2. Prefixed with `--` (mandatory).
3. A valid CSS custom property name under strict validation.

`value` supports existing `defineVars` value shapes, including conditional objects and typed values.

### Validation contract (hard-fail)
Compilation fails with actionable errors when:

1. `name` is missing the `--` prefix.
2. `name` is dynamic/non-static.
3. `name` violates syntax validation (`/^--[A-Za-z_][A-Za-z0-9_-]*$/`).
4. Multiple declarations resolve to the same custom property name.
5. `namedVar` collides with legacy `'--foo'` declarations.

### Runtime behavior
`stylex.namedVar` is compile-time-only. If uncompiled code reaches runtime, it throws a clear error instructing users to ensure StyleX compilation is correctly configured.

---

### Transform semantics
1. If a `defineVars` value uses `stylex.namedVar('--x', value)`, `--x` is used as the emitted property name instead of hash generation.
2. The wrapped `value` is unwrapped and processed through existing logic, including conditional and typed variable handling.
3. Mixed declarations are supported: regular hashed variables and explicitly named variables can coexist in the same object.

### Cross-file theme reference semantics
Imported `defineVars` references preserve explicit names from both:

1. `stylex.namedVar('--x', value)`, and
2. legacy `'--x': value` keys.

If imported declarations are invalid, compilation fails deterministically with cause-preserving diagnostics.

---

## Diagnostics
This RFC adds specific error modes for `defineVars` + `namedVar`:

1. Missing required `--` prefix.
2. Invalid custom property syntax.
3. Dynamic/non-static name argument.
4. Duplicate custom property name collisions.
5. Invalid `namedVar` call shape (argument count, etc.).

All failures are hard errors with code-frame context and remediation guidance.

---

## Backward Compatibility
1. Existing `defineVars` usage remains valid.
2. Existing legacy explicit custom property keys (`'--foo'`) remain valid.
3. This change is additive for valid programs and stricter for invalid ones.

---

### 1) Auto-prefix missing `--`
Rejected. Auto-prefixing hides mistakes and reduces predictability. Mandatory prefixing creates explicit intent and better diagnostics.

### 2) Object wrapper syntax
Example: `{ name: '--surface-bg', value: 'black' }`.
Rejected due to weaker ergonomics, easier accidental misuse, and less clear compile-time intent than a dedicated helper call.

### 3) Keep only legacy `'--foo'` keys
Rejected because it does not preserve ergonomic dot-notation authoring patterns.

---

## Test Plan
1. Transform tests validate explicit-name emission, conditional values, typed values, and mixed named/hashed usage.
2. Validation tests cover missing prefix, invalid syntax, dynamic names, and duplicate/collision errors.
3. Evaluate-path tests verify imported refs resolve explicit names and invalid imported declarations fail deterministically.
4. Runtime API tests verify the uncompiled runtime stub throws with actionable guidance.

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineVars-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-defineVars-test.js
@@ -400,6 +400,64 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
 
+    test('namedVar allows custom property names with dot-notation keys', () => {
+      const { code, metadata } = transform(`
+        import * as stylex from '@stylexjs/stylex';
+        export const vars = stylex.defineVars({
+          background: stylex.namedVar('--surface-bg', 'black'),
+          accent: stylex.namedVar('--brand-accent', 'blue'),
+        });
+      `);
+
+      expect(code).toContain('background: "var(--surface-bg)"');
+      expect(code).toContain('accent: "var(--brand-accent)"');
+      expect(metadata.stylex[0][1].ltr).toContain('--surface-bg:black;');
+      expect(metadata.stylex[0][1].ltr).toContain('--brand-accent:blue;');
+    });
+
+    test('namedVar supports conditional values', () => {
+      const { code, metadata } = transform(`
+        import * as stylex from '@stylexjs/stylex';
+        export const vars = stylex.defineVars({
+          accent: stylex.namedVar('--brand-accent', {
+            default: 'blue',
+            '@media (prefers-color-scheme: dark)': 'lightblue',
+          }),
+        });
+      `);
+
+      expect(code).toContain('accent: "var(--brand-accent)"');
+      expect(metadata.stylex[0][1].ltr).toContain('--brand-accent:blue;');
+      expect(metadata.stylex[1][1].ltr).toContain('--brand-accent:lightblue;');
+    });
+
+    test('namedVar supports typed vars and emits @property for custom names', () => {
+      const { code, metadata } = transform(`
+        import * as stylex from '@stylexjs/stylex';
+        export const vars = stylex.defineVars({
+          accent: stylex.namedVar('--brand-accent', stylex.types.color('red')),
+        });
+      `);
+
+      expect(code).toContain('accent: "var(--brand-accent)"');
+      expect(metadata.stylex[0][1].ltr).toContain(
+        '@property --brand-accent { syntax: "<color>";',
+      );
+      expect(metadata.stylex[1][1].ltr).toContain('--brand-accent:red;');
+    });
+
+    test('namedVar works with named imports', () => {
+      const { code, metadata } = transform(`
+        import { defineVars, namedVar } from '@stylexjs/stylex';
+        export const vars = defineVars({
+          background: namedVar('--surface-bg', 'black'),
+        });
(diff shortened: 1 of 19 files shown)


nmn:
Before shipping this, let’s document the recipe for achieving this today, which is to use `defineVars` with the literal names with the prefixes and then to also use `defineConsts` to “forward” those vars with ergonomic keys.

It may be a little more boilerplate, but it is already possible with the APIs today.

And because of that, I think it would be more useful to ship the `stylex.env` features for defining constants and helper function as a config.

so you’d be able to do something like:

```tsx
const vars = stylex.env.defineNamedVar(…)
```

The PR for StyleX.env is already up. It should be another small lift to enable it to call StyleX APIs behind the scenes.

Thoughts?

vincentriemer:
Yeah I see your point so if you think this doesn't provide enough value I'm fine with closing.

My concerns with the boilerplate workaround and using the env API is the same: LLM authorship. It's not that I don't think an LLM can write the extra boilerplate, but I think that the instructions necessary to tell the LLM to compose the defineVars & defineConsts could cause it to assume those APIs are more composable than they really are. When it comes to the env API I think that would struggle with the fact that the APIs defined by env would be different between every project which I think could harm an LLM's ability to leverage those APIs.

mellyeliu:
I do think it'd be nice to keep small helpers like these in `stylex.env` to keep the API surface small

I think if we do go down the `stylex.env` approach, we should pre-bake in a standard set of functions or a helper package that would include this API. Definitely don't think the workaround is useable otherwise.

vincentriemer:
closing in favor of `stylex.env` APIs

PR #1495, opened 2026-02-24
[docs] add stylex.env api

mellyeliu:


packages/docs/content/docs/api/javascript/env.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+```tsx title="tokens.stylex.js"
+import * as stylex from '@stylexjs/stylex';
+
+export const colors = stylex.defineVars(stylex.env.colors);
+```
+
+### Functions
+
+Define shareable functions using the `env` API. All function arguments must be locally defined. Functions
```
vincentriemer: This documentation says that all function arguments must be locally defined yet in the examples below you're using a color imported from a `.stylex.js` file which seems to contradict this constraint.

PR #1501, opened 2026-03-03
[unplugin] Stop assigning to bundle object in unplugin generateBundle hook

### LINKED ISSUE #1387: StyleX plugin does not trigger Vite file hash to update
#### Description

When using the StyleX plugin with Vite, the generated CSS styles are appended to the output files after the initial build. However, Vite's hashed filenames are based only on the original file content, not the appended StyleX styles. As a result, updating StyleX styles does **not** change the hash in the output filename, causing browsers and CDNs to serve stale CSS.

#### Impact

- StyleX style changes are not reflected until a full cache purge or manual intervention.
- Users may see outdated styles after a deployment.
- This breaks the expected cache busting behavior of Vite's asset pipeline.

#### Steps to Reproduce

1. Build the project with StyleX and Vite.
2. Deploy and load the site.
3. Change a StyleX style and rebuild.
4. Observe that the output file hash does not change, and the new styles are not loaded.

#### Expected Behavior

Updating StyleX styles should result in a new hash for the affected output files, ensuring browsers and CDNs fetch the latest styles.

#### Suggested Solution

The StyleX plugin should ensure that appended styles are included in the hash calculation for output files, so any style change triggers proper cache busting.

### Expected behavior

Updating StyleX styles should result in a new hash for the affected output files, ensuring browsers and CDNs fetch the latest styles.

skovhus:

Vite build warns due to a stylex unplugin and vite incompatibility.

`[plugin @stylexjs/unplugin] Error: This plugin assigns to bundle variable. This is discouraged by Rollup and is not supported by Rolldown. This will be ignored.`

The `replaceCssAssetWithHashedCopy` function was directly assigning to the Rollup `bundle` object (`bundle[key] = asset`), which is [discouraged](https://rollupjs.org/plugin-development/#generatebundle) by Rollup and unsupported by Rolldown. This also caused stale content hashes in CSS filenames since the hash was computed before StyleX CSS was appended.

The function already calls `this.emitFile()` which properly adds the new asset to the bundle with a correctly-hashed filename. Remove the redundant manual bundle manipulation and rely solely on `emitFile` for adding entries and `delete bundle[key]` for removing the old asset, both of which are supported by the Rollup plugin API.

Chatted with @sapphi-red and he pointed me in the right direction.

## Linked PR/Issues

Fixes #1387

## Additional Context

CleanShot 2026-03-03 at 21 49 09@2x

See https://rollupjs.org/plugin-development/#generatebundle

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/unplugin/src/core.js
+++ b/packages/@stylexjs/unplugin/src/core.js
@@ -109,16 +109,7 @@ export function replaceCssAssetWithHashedCopy(ctx, bundle, asset, nextSource) {
     return;
   }
   replaceBundleReferences(bundle, oldFileName, nextFileName);
-  const emitted = bundle[nextFileName];
-  if (emitted && emitted !== asset) {
-    delete bundle[nextFileName];
-  }
-  asset.fileName = nextFileName;
-  asset.source = nextSource;
-  if (bundle[oldFileName] === asset) {
-    delete bundle[oldFileName];
-  }
-  bundle[nextFileName] = asset;
+  delete bundle[oldFileName];
 }
 
 function readJSON(file) {


skovhus:
@sapphi-red Does the fix look right to you? Seems to work from my testing.

packages/@stylexjs/unplugin/src/core.js
sapphi-red, line 108: I guess the warning won't happen any more with this change. But this line is updating the content after the file is generated, which can cause stale content hashes. (as this line changes the content without changing the file hash)

PR #1514, opened 2026-03-06
feat: allow overriding type of `stylex.env` in TypeScript

### LINKED ISSUE #1507: `stylex.env` type safety

Hello 👋

`stylex.env` is a game changer imho and I can't even describe how useful that will be. There's one issue though when using it with Typescript 😄 

Image

Ideally it'd be nice to somehow type what exactly is in my own `stylex.env` - in TS it's usually achieved by module augmentation/module declaration/declaration merging which allows to extend any interface that library exposes. That's the trick used by popular JS libraries that require top notch type-safety, few examples:

[TanStack Router](https://tanstack.com/router/latest/docs/decisions-on-dx#declaring-the-router-instance-for-type-inference)
[Effect](https://effect.website/docs/schema/annotations/#custom-annotations)

Maybe it'd be nice to expose this type in such a way that it can be overwritten by TS consumers? In the future maybe it could even be done in such a way that enforces the same stuff being passed to bundler settings and the same stuff being used in the actual app code (but due to many possibilities of how to integrate stylex into the pipeline that'd be probably future step).

Would it be valuable addition in your opinion? I'm open to contribute this and make a PR if you see the value here. I'm also not sure how Flow fits into the equation here however that's why I'd need some guidance and why I created an issue in the first place. 

Thanks a lot in advance!

pawelblaszczyk5:

As explained in https://github.com/facebook/stylex/issues/1507 this allows user to override type of `stylex.env` if they're using TypeScript. It uses declaration merging and exposes `Register` type that can be augmented in the user land. `stylex.env` is typed by either the default, existing type or whatever user sets as `env` in the `Register` type.

Example usage:

```ts
declare module '@stylexjs/stylex/lib/types/StyleXTypes' {
  interface Register {
    env: {
      utilities: {
        opacity: (color: string, value: number) => string;
      };
      tokens: {
        colors: {
          primary: string;
        };
      };
    };
  }
}

stylex.env.utilities.opacity('#f00', 0.15);
stylex.env.tokens.colors.primary;
```

image

You get sweet autocompletion and type checking here and if you adjust types of the `env` you'll get type errors everywhere where you didn't update the usage. Both constants and utility functions can be typed this way.

## Linked PR/Issues

Fixes #1507 

## Additional Context

Open points:

- Currently it exports `Register` from `@stylexjs/stylex/lib/types/StyleXTypes` which user needs to refer when augmenting module. It could go directly to `@stylexjs/stylex` but then Flow version would need to export some basically unused type, because the entry point for TS is generated from Flow file if I understand correctly?
- Flow story? I've did some research and if I understand correctly there's no way to achieve something similar here? 
- In `typescript-tests` it'll currently impact all files and it doesn't verify whether `stylex.env` is still the fallback if user didn't override this. This could be fixed by splitting `tsconfig.json` into multiple projects but I didn't want to pollute it too much and I'm not 100% sure about the value.
- I've added some basic documentation but I'm not 100% sure whether it's a best place for this. `stylex.env` currently it's only in the API section and it doesn't feel fully like API, it feels more like guide. 
- As pointed out in the original issue, verifying whether this type matches what user passed to babel config is a bit tricky. These are often separate packages "@stylexjs/stylex" vs "@stylexjs/babel-plugin". Babel config is often untyped. And there're many ways to integrate, e.g. there's also "@stylexjs/unplugin".
- Currently I'm not verifying whether what user passed here can indeed be what `stylex.env` allows - object with constants and functions. Not sure how important it is, I've checked out and many TS libraries doesn't verify this. 

Open to addressing any of them, I guess some of them could be nice possible improvements for the future.

## Pre-flight checklist

### DIFF
--- a/packages/docs/content/docs/api/javascript/env.mdx
+++ b/packages/docs/content/docs/api/javascript/env.mdx
@@ -92,3 +92,31 @@ const styles = stylex.create({
   },
 });
 ```
+
+### TypeScript usage
+
+You can use _module declaration_ to override the `env` type to match your exact
+usage. You only need to do this once per project. As long as this file is
+included by your `tsconfig` it'll override the `env` type in all other files in this
+project.
+
+_Note: This approach doesn't currently check whether this type matches what
+you're passing to
+[babel plugin options](/docs/api/configuration/babel-plugin/#env)._
+
+```tsx
+declare module '@stylexjs/stylex/lib/types/StyleXTypes' {
+  interface Register {
+    env: {
+      tokens: {
+        colors: {
+          primary: string;
+        };
+      };
+    };
+  }
+}
+
+// Usage
+stylex.env.tokens.colors.primary; // Type checks correctly and you get intellisense
+```


packages/@stylexjs/stylex/src/types/StyleXTypes.js
```diff
@@ -284,3 +284,5 @@ export type StyleX$When = {
     _customMarker?: MapNamespace<{ +marker: 'custom-marker' }>,
   ) => ':where-any-sibling',
 };
+
+export type StyleX$Env = $ReadOnly<{ [string]: mixed }>;
```
pawelblaszczyk5, line 288: Moved the type into `StyleXTypes.d.ts` and `StyleXTypes.js` to make it possible to make them different across Flow/TS

packages/docs/content/docs/api/javascript/env.mdx
```diff
@@ -92,3 +92,31 @@ const styles = stylex.create({
...
+usage. You only need to do this once per project. As long as this file is
+included by your `tsconfig` it'll override the `env` type in all other files in this
+project.
+
+_Note: This approach doesn't currently check whether this type matches what
+you're passing to
+[babel plugin options](/docs/api/configuration/babel-plugin/#env)._
+
+```tsx
+declare module '@stylexjs/stylex/lib/types/StyleXTypes' {
```
pawelblaszczyk5, line 108: As mentioned in the description - these could be just "@stylexjs/stylex" if we exported this from entry point, that would probably also mean adding some "mock" type in Flow, because entry point is auto generated? Or am I missing something obvious here?

PR #1521, opened 2026-03-14
[eslint-plugin] fix stylex/valid-styles false positives for member expressions 

### LINKED ISSUE #1520: ESLint: stylex/valid-styles reports false positives for function parameter references
### Describe the issue

`stylex/valid-styles` reports false positives for function parameter references in stylex.create arrow functions

The valid-styles rule flags `props.xxx` references inside stylex.create arrow functions as invalid values, even though StyleX compiles these into CSS variables at build time.

### Expected behavior

The rule correctly validates static entries but should allow values inside arrow function bodies in stylex.create, since these are dynamic parameters that StyleX handles at compile time. The function's type annotation already constrains the value — the ESLint rule shouldn't duplicate that check with a much narrower set of allowed values.

This affects all CSS properties where a function parameter is used as the value: `backgroundColor`, `color`, `transitionDelay`, `animationDelay`, `boxShadow`, etc. It also affects nested pseudo-class objects like `":hover": props.glowColor`.

### Steps to reproduce

```tsx
  badge: (props: { badgeColor: string }) => ({
    backgroundColor: props.badgeColor, // ❌ "backgroundColor value must be one of: ..."
  }),
});
```

### Test case

_No response_

### Additional comments

_No response_

skovhus:

The valid-styles rule only checked for Identifier nodes when skipping validation of arrow function parameters. MemberExpression nodes like props.badgeColor were not recognized, causing false positives. This adds a check that walks up the object chain of MemberExpressions to find the root identifier and skips validation when it maps to a function argument.

## Linked PR/Issues

Fixes #1520

## Additional Context

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -672,6 +672,28 @@ eslintTester.run('stylex-valid-styles', rule.default, {
       })
     })
     `,
+    // test using member expressions on function params in dynamic styles
+    `
+    import * as stylex from 'stylex';
+    stylex.create({
+      badge: (props) => ({
+        backgroundColor: props.badgeColor,
+        color: props.color,
+      })
+    })
+    `,
+    // test member expressions on function params with pseudo-classes
+    `
+    import * as stylex from 'stylex';
+    stylex.create({
+      root: (props) => ({
+        color: {
+          default: props.textColor,
+          ':hover': props.hoverColor,
+        },
+      })
+    })
+    `,
     // test importing vars from paths including theme file extension
     `
     import * as stylex from '@stylexjs/stylex';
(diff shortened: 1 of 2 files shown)


packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -881,6 +903,26 @@ eslintTester.run('stylex-valid-styles', rule.default, {
         });
       `,
     },
+    {
+      code: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          foo: {
+            outlineOffset: 2,
```
skovhus, line 911: This was added by mistake.

skovhus:
@mellyeliu seems like this caused CI to fails. Not sure why this was able to merge, maybe we need a merge status check for `tests / packages`?

packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
skovhus: Fix is in https://github.com/facebook/stylex/pull/1523

PR #1529, opened 2026-03-17
(babel-plugin): Add layersBefore/layersAfter config to processStylexR…

### LINKED ISSUE #1526: [babel-plugin] Add `layersBefore` and `layersAfter` configs
It's common for users to want to declare their own resets and overrides. StyleX generates layers from @layer priority1, priority2. Let's modify this to allow for configuring the name of the first and last layers.

```
@layer layerBefore, typography, priority1, ..., priority9, layerAfter;
```

We'll want to modify the babel config here: https://github.com/facebook/stylex/blob/main/packages/%40stylexjs/babel-plugin/src/index.js#L567-L571 to output the correct values.

This might look something like `layersBefore`, `layersAfter`. Consider also adding a prefix arg `layerPrefix` so apps can define custom layer names. To avoid cluttering the babel config, we should consider shortening this somehow.

All of this should go within the babel config of StyleX: https://github.com/facebook/stylex/blob/main/packages/%40stylexjs/babel-plugin/src/utils/state-manager.js

chunw-meta:
## What changed / motivation?

When useCSSLayers is enabled, StyleX generates a layer order declaration:

`@layer priority1, priority2, ..., priorityN;`

Users often need to insert custom CSS layers before or after StyleX's priority layers — for example, reset layers that should have lower precedence, or theme override layers that should sit between base component styles and prop-level overrides. Previously this required fragile workarounds like injecting a hardcoded @layer declaration via a separate build plugin, which could break if StyleX generated a different number of priority layers.

This PR adds two new config options to processStylexRules: layersBefore and layersAfter. These prepend/append user-defined layer names to the generated @layer declaration:

`@layer reset, typography, priority1, ..., priorityN, xds.theme;`

This enables the three-level specificity hierarchy described in the issue:

Base component styles (StyleX priority layers) — lowest
Theme overrides (custom layer via layersAfter) — middle
Prop overrides (StyleX higher priority layers) — highest

Usage
```
// Babel plugin
stylexPlugin.processStylexRules(rules, {
  useLayers: true,
  layersBefore: ['reset', 'typography'],
  layersAfter: ['xds.theme'],
});

// Bundler plugins (rollup, postcss, unplugin)
stylexPlugin({
  useCSSLayers: true,
  layersBefore: ['reset', 'typography'],
  layersAfter: ['xds.theme'],
});
```

Changes
@stylexjs/babel-plugin — Added layersBefore/layersAfter to the processStylexRules config type and header generation logic
@stylexjs/rollup-plugin — Thread new options through to processStylexRules
@stylexjs/postcss-plugin — Thread new options through plugin → builder → bundler
@stylexjs/unplugin — Thread new options through to processCollectedRulesToCSS
When useLayers is false, both options are ignored. When omitted or empty, output is unchanged (fully backward compatible).

## Linked PR/Issues

Fixes #1526

## Additional Context
`npx jest`:
5 new test cases added to transform-process-test.js 
```
 PASS  __tests__/transform-process-test.js
  @stylexjs/babel-plugin
    [transform] stylexPlugin.processStylexRules
      ✓ no rules (66 ms)
      ✓ all rules (useLayers:false) (46 ms)
      ✓ all rules (useLayers:true) (45 ms)
      ✓ useLayers with layersBefore (46 ms)
      ✓ useLayers with layersAfter (57 ms)
      ✓ useLayers with both layersBefore and layersAfter (46 ms)
      ✓ layersBefore/layersAfter are ignored when useLayers is false (47 ms)
      ✓ empty layersBefore/layersAfter produce standard layer declaration (46 ms)
      ✓ all rules (legacyDisableLayers:true) (49 ms)
      ✓ legacy-expand-shorthands with logical styles polyfill (4 ms)
      ✓ legacy-expand-shorthands duplicates theme selectors for higher precedence (3 ms)
      ✓ no mutation of rules (48 ms)
      ✓ useLegacyClassnamesSort: false (default behavior) (50 ms)
      ✓ useLegacyClassnamesSort: true (legacy behavior) (50 ms)
      ✓ sort is deterministic regardless of input order
      ✓ sort is deterministic with duplicate rules in different input orders

Tests:       16 passed, 16 total
Snapshots:   19 passed, 19 total
Time:        1.344 s, estimated 2 s
```
and `npm run test:packages `

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -385,6 +385,246 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
 
+    test('useLayers with layersBefore', () => {
+      const { metadata } = transform(fixture);
+      expect(
+        stylexPlugin.processStylexRules(metadata, {
+          useLayers: true,
+          layersBefore: ['reset', 'typography'],
+          enableLTRRTLComments: false,
+        }),
+      ).toMatchInlineSnapshot(`
+        "
+        @layer reset, typography, priority1, priority2, priority3, priority4;
+        @property --x-color { syntax: "*"; inherits: false;}
+        @keyframes x35atj5-B{0%{box-shadow:1px 2px 3px 4px red;color:yellow;}100%{box-shadow:10px 20px 30px 40px green;color:var(--orange-theme-color);}}
+        :root, .xsg933n{--blue-xpqh4lw:blue;--marginTokens-x8nt2k2:10px;--colorTokens-xkxfyv:red;}
+        :root, .xbiwvf9{--small-x19twipt:2px;--medium-xypjos2:4px;--large-x1ec7iuc:8px;}
+        @media (prefers-color-scheme: dark){:root, .xsg933n{--colorTokens-xkxfyv:lightblue;}}
+        @media (min-width: 600px){:root, .xsg933n{--marginTokens-x8nt2k2:20px;}}
+        @supports (color: oklab(0 0 0)){@media (prefers-color-scheme: dark){:root, .xsg933n{--colorTokens-xkxfyv:oklab(0.7 -0.3 -0.4);}}}
+        .x6xqkwy.x6xqkwy, .x6xqkwy.x6xqkwy:root{--blue-xpqh4lw:lightblue;}
+        .x57uvma.x57uvma, .x57uvma.x57uvma:root{--large-x1ec7iuc:20px;--medium-xypjos2:10px;--small-x19twipt:5px;}
+        .--orange-theme-color-xufgesz{--orange-theme-color:red}
+        @layer priority2{
+        .margin-xymmreb{margin:10px 20px}
+        .padding-xss17vw{padding:var(--large-x1ec7iuc)}
+        }
+        @layer priority3{
+        .borderColor-x1bg2uv5{border-color:green}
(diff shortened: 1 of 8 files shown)


packages/@stylexjs/babel-plugin/src/index.js
```diff
@@ -566,7 +570,11 @@ function processStylexRules(
 
   const header = useLayers
     ? '\n@layer ' +
-      grouped.map((_, index) => `priority${index + 1}`).join(', ') +
+      [
```
mellyeliu, line 580: we should probably log a warning if users have `layersBefore` set and `useCSSLayers` off

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -385,6 +385,76 @@ describe('@stylexjs/babel-plugin', () => {
...
+
+    test('empty layersBefore/layersAfter produce standard layer declaration', () => {
+      const { metadata } = transform(fixture);
+      const result = stylexPlugin.processStylexRules(metadata, {
+        useLayers: true,
+        layersBefore: [],
+        layersAfter: [],
+        enableLTRRTLComments: false,
+      });
+      expect(result).toContain(
```
mellyeliu: can we snapshot the results instead? instead of using the toContain approach, use `toMatchInlineSnapshot`

chunw-meta: yep, resolved the comments in amended commit: https://github.com/facebook/stylex/compare/bcd25ef28753042daf6e42bad7c9e640f805a80a..70a88c96fefb48671f77fd9445b0a717ae053668

PR #1537, opened 2026-03-20
[eslint-plugin] Add flex shorthand expansion to stylex-valid-shorthands

abhakat:

Added flex shorthand expansion to `stylex-valid-shorthands` using the same `createSpecificTransformer` pattern already used for `background`, `font`, `border`, `grid`, and `outline`.

Changes:
- Added `flex: createSpecificTransformer('flex')` to shorthand aliases.
- Added `expandFlexShorthand()` in `splitShorthands.js` plus helper classifiers for:
  - unitless flex numbers (`flex-grow` / `flex-shrink`)
  - valid basis values (`auto`, `content`, `min-content`, `max-content`, `fit-content`, `calc()/min()/max()/clamp()/fit-content()`, `var()`, unit/percent values)
- Added `property === 'flex'` dispatch in `splitSpecificShorthands` before generic single-value passthrough so cases like `flex: 1` still expand.
- Added/updated tests covering:
  - keywords: `auto`, `none`, `initial`
  - single number
  - single basis
  - two-value forms (`grow shrink`, `grow basis`)
  - three-value form (`grow shrink basis`)
  - `calc()` basis
  - `!important` propagation

Behavior:
- `flex: '1 0 auto'` now reports and autofixes to:
  - `flexGrow: '1'`
  - `flexShrink: '0'`
  - `flexBasis: 'auto'`

Motivation:
- `flex` shorthand is common in StyleX authoring and migration paths.
- Previously this pattern was flagged but not expanded as a first-class shorthand.
- This brings `flex` in line with existing shorthand autofix behavior while keeping fixes safe and deterministic.

## Linked PR/Issues

N/A

## Additional Context

Example autofix:
- Input: `flex: '1 0 auto'`
- Output: `flexGrow: '1', flexShrink: '0', flexBasis: 'auto'`

Tests:
- `npx jest --watchman=false --coverage=false stylex-valid-shorthands-test.js`
- Result: `PASS` (`94 passed`, `0 failed`)

Screenshots:
- N/A (ESLint rule + autofix behavior change)

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-shorthands-test.js
@@ -1752,5 +1752,303 @@ eslintTester.run('stylex-valid-shorthands', rule.default, {
         },
       ],
     },
+    // flex: single number expands to grow/shrink/basis
+    {
+      code: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          main: {
+            flex: 1,
+          },
+        });
+      `,
+      output: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          main: {
+            flexGrow: '1',
+            flexShrink: '1',
+            flexBasis: '0%',
+          },
+        });
+      `,
+      errors: [
+        {
+          message:
+            'Property shorthands using multiple values like "flex: 1" are not supported in StyleX. Separate into individual properties.',
+        },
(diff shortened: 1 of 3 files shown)


skovhus:
I'm not sure I follow the motivation here, the compiler doesn't require them expanded. 

And the linter didn't flag `flex: "auto"` as problematic before, so why do we force people to expand them? @mellyeliu ? @abhakat ?

skovhus:
I still find this change very restrictive. Can we add a linter config option?

PR #1594, opened 2026-04-06
[eslint-plugin] Add auto-fixers and suggest-fixes to stylex-valid-styles

abhakat:
## Summary

- **Narrow `grid*` glob** in `legacyProps` to only block grid shorthands (`grid`, `gridArea`, `gridColumn`, `gridRow`, `gridTemplate`, `gridGap`, `gridColumnGap`, `gridRowGap`). Grid longhands like `gridTemplateColumns`, `gridRowStart`, etc. are now allowed under `banPropsForLegacy`.
- **Add auto-fixers for grid shorthands** when blocked via `banPropsForLegacy` or `propLimits`. Reuses the existing expansion logic in `splitShorthands.js`. Fixers only fire when the property is already an error — they are conditional on `banPropsForLegacy`/`propLimits`, not unconditional.
- **Upgrade border from suggest-only to auto-fix** — the `border()` function now returns both `fix` and `suggest`.
- **Add font shorthand auto-fixer** — expands to `fontSize`, `fontFamily`, `fontWeight`, etc.
- **Add animation shorthand suggest-fixer** (suggest-only, not auto-fix, because `animationName` requires a `keyframes()` reference that can't be produced deterministically).
- **Add suggest-fix for empty string** (`''` → suggest `null`).
- **Add suggest-fix for bare numeric strings** (e.g. `zIndex: '10'` → suggest `zIndex: 10`).
- **Wire `fix` propagation** through `makeUnionRule` and both reporting paths (`CSSPropertyReplacements` and `validateStyleValue`).
- **Property-aware `serializeValue`** — emits number literals for properties that accept them (`lineHeight`, `fontWeight`, `rowGap`, etc.), quoted strings everywhere else. Handles single-quote escaping.

### Design decisions

- Grid shorthand fixers are gated on `banPropsForLegacy`/`propLimits` (the property must already be blocked). This matches the internal config where `banPropsForLegacy: true` is set on every surface.
- Animation is suggest-only because the expanded `animationName` would be a raw string literal, which fails the same rule (it expects a `stylex.keyframes()` reference).
- The `grid` shorthand itself (as opposed to `gridArea`, `gridColumn`, etc.) has no auto-fixer because it's too complex for deterministic expansion.

## Test plan

- [x] All 532 existing + new tests pass (`jest --watchman=false --no-coverage __tests__/`)
- [x] ESLint: zero errors on changed files
- [x] Flow: zero errors on changed files
- [x] Verified grid auto-fix output is lint-clean on re-run (no secondary lint errors)
- [x] Verified animation suggest-only does not auto-apply

### DIFF
--- a/packages/@stylexjs/eslint-plugin/src/reference/cssProperties.js
+++ b/packages/@stylexjs/eslint-plugin/src/reference/cssProperties.js
@@ -23,10 +23,6 @@ import isPercentage from '../rules/isPercentage';
 import isAbsoluteLength from '../rules/isAbsoluteLength';
 import isRelativeLength from '../rules/isRelativeLength';
 import { borderSplitter } from '../utils/split-css-value';
-import {
-  splitSpecificShorthands,
-  CANNOT_FIX,
-} from '../utils/splitShorthands';
 
 export type RuleResponse = void | {
   message: string,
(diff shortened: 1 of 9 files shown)


packages/@stylexjs/eslint-plugin/src/reference/cssProperties.js
```diff
@@ -495,34 +537,44 @@ const border =
...
+            newRules.push(
+              `border${suffix}Style: ${serializeValue(`border${suffix}Style`, style)}`,
+            );
+          }
+          if (color != null) {
+            newRules.push(
+              `border${suffix}Color: ${serializeValue(`border${suffix}Color`, color)}`,
+            );
+          }
+          return fixer.replaceText(prop, newRules.join(',\n    '));
```
mellyeliu: This feels fragile and may break formatting. Can we check `node.loc.start.column` or similar? I think we also have an indentation function somewhere

mellyeliu (approved):

PR #1612, opened 2026-04-14
[babel-plugin] Fix pseudo-element + pseudo-class selector ordering

### LINKED ISSUE #1473: Pseudo-elements don't work with pseudoclasses
We enforce top-level pseudo-elements, but given something like so:

```js
    foo: {
    "::before": {
      content: '"✓"',
      color: {
        default: 'yellow',
        ':hover': 'red'
      },
   }
})
```

The outputted CSS is as follows, which is invalid, as pseudo-elements must after at the end of the selector.
```css
.x1qdmp1q::before:hover{color:red}
```

skovhus:

Pseudo-elements like `::before` and `::after` must appear at the end of a CSS selector. When combining pseudo-elements with pseudo-class conditions (e.g. `:hover`, `:active`), StyleX generated invalid selectors like `.x123::after:hover` which browsers silently discard. The fix reorders them to `.x123:hover::after`.

## Linked PR/Issues

Fixes #1473

## Additional Context

n/a

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
@@ -2408,7 +2408,7 @@ describe('@stylexjs/babel-plugin', () => {
 
           // Pseudo-classes (:hover, :active) must come before the
           // pseudo-element (::after) in the selector for valid CSS.
-          const rules = metadata.stylex.map(([className, { ltr }]) => ltr);
+          const rules = metadata.stylex.map(([_className, { ltr }]) => ltr);
           expect(rules).toMatchInlineSnapshot(`
             [
               ".x1s928wv::after{content:""}",
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
@@ -2270,7 +2270,6 @@ describe('@stylexjs/babel-plugin', () => {
           `);
         });
 
-        // Generates invalid CSS, need to revisit this API
         test('"::before" containing pseudo-classes', () => {
           const { code, metadata } = transform(`
             import * as stylex from '@stylexjs/stylex';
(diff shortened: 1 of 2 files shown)


packages/@stylexjs/babel-plugin/__tests__/transform-stylex-create-test.js
```diff
@@ -2389,6 +2388,36 @@ describe('@stylexjs/babel-plugin', () => {
...
+                    ':active': '0 0 0 1px darkblue',
+                  },
+                },
+              },
+            });
+          `);
+
+          // Pseudo-classes (:hover, :active) must come before the
+          // pseudo-element (::after) in the selector for valid CSS.
+          const rules = metadata.stylex.map(([className, { ltr }]) => ltr);
```
skovhus:
```suggestion
          const rules = metadata.stylex.map(([_className, { ltr }]) => ltr);
```

packages/@stylexjs/babel-plugin/src/shared/utils/generate-css-rule.js
```diff
@@ -29,7 +29,12 @@ function buildNestedCSSRule(
   atRules: $ReadOnlyArray<string>,
   constRules: $ReadOnlyArray<string>,
 ): string {
-  const pseudo = pseudos.filter((p) => p !== '::thumb').join('');
+  const filteredPseudos = pseudos.filter((p) => p !== '::thumb');
+  // Pseudo-elements (::before, ::after, etc.) must come after pseudo-classes
+  // in the selector. e.g. `.class:hover::before` not `.class::before:hover`
+  const pseudoClasses = filteredPseudos.filter((p) => !p.startsWith('::'));
+  const pseudoElements = filteredPseudos.filter((p) => p.startsWith('::'));
+  const pseudo = [...pseudoClasses, ...pseudoElements].join('');
```
Dwlad90: A more optimized solution with a single-pass loop

```suggestion
  // Pseudo-elements (::before, ::after, etc.) must come after pseudo-classes
  // in the selector. e.g. `.class:hover::before` not `.class::before:hover`
const { classes, elements } = pseudos.reduce((acc, p) => {
  if (p === '::thumb') return acc;

  if (p.startsWith('::')) {
    acc.elements += p;
  } else {
    acc.classes += p;
  }
  return acc;
}, { classes: '', elements: '' });

const pseudo = classes + elements;
```

skovhus: Thanks!

skovhus:
@mellyeliu thanks for wrapping this up

mellyeliu (approved):

mellyeliu:
@skovhus thanks for putting up this fix! sorry for the delay

PR #1613, opened 2026-04-14
Catch evaluate() failures when resolving stylex props

chunw-meta:

Handle exceptions thrown while evaluating `stylex.props()` arguments in the Babel transform.

This change wraps the evaluation path in `parseNullableStyle()` with a `try/catch` and treats evaluation failures as a bailout case (`'other'`) instead of letting the transform crash.

Some expressions passed into `stylex.props()` can cause `evaluate()` to throw during compile-time analysis. Before this change, those exceptions could fail the transform entirely.

This changeset:
- add a `try/catch` around `evaluate(...)` in `stylex/packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js`
- if evaluation throws, return `'other'`
- this preserves the existing bailout behavior for non-optimizable expressions

When the transform cannot confidently resolve a style expression, bailing out is already the intended fallback behavior. This change just extends that behavior to thrown evaluation errors.

## Linked PR/Issues
Caught this issue when running Meta script to generate CSS stylesheets for comparison during the upgrade process. The script throws on valid stylex patterns during evaluation.

## Additional Context

`npm run test`

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js
+++ b/packages/@stylexjs/babel-plugin/src/visitors/stylex-props.js
@@ -363,17 +363,21 @@ function parseNullableStyle(
     }
   }
 
-  const parsedObj = evaluate(path, state, evaluatePathFnConfig);
+  try {
+    const parsedObj = evaluate(path, state, evaluatePathFnConfig);
 
-  if (
-    parsedObj.confident &&
-    parsedObj.value != null &&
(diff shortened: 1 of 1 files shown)


mellyeliu:
Discussed offline. We should resolve the evaluate-path failures from https://github.com/facebook/stylex/pull/1567 instead of silently bypassing

PR #1615, opened 2026-04-15
[eslint-plugin] fix valid-styles rejecting nested stylex.when.* conditions

### LINKED ISSUE #1614: ESLint valid-styles rule rejects nested conditions inside stylex.when.*() computed keys
### Describe the issue

The stylex-valid-styles ESLint rule rejects a media query (or other condition) nested inside a stylex.when.ancestor() / stylex.when.descendant() / etc.
  computed key, even though the Babel compiler handles it correctly and produces valid CSS.

### Expected behavior

The ESLint rule should accept stylex.when.*() as a condition key even when its value is a nested condition object, matching the compiler behavior.

@stylexjs/eslint-plugin — all versions since stylex.when support was added (#1270).

### Steps to reproduce

```tsx

    icon: {
      transitionDuration: {
        default: vars.slow,
        [stylex.when.ancestor(":active")]: vars.fast, // OK
        [stylex.when.ancestor(":hover")]: {
          default: null,
          "@media (hover: hover)": vars.fast, // ERROR
        },
      },
    },
  });
```

### Test case

_No response_

### Additional comments

_No response_

skovhus:

The `ObjectExpression` branch in `checkStyleProperty` did not recognize `stylex.when.*()` computed keys, causing "Keys must be strings" errors when nesting a media query or other condition inside a when call. 

## Linked PR/Issues

Fixes https://github.com/facebook/stylex/issues/1614

## Additional Context

n/a

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -341,6 +341,43 @@ eslintTester.run('stylex-valid-styles', rule.default, {
       `,
       options: [{ allowOuterPseudoAndMedia: true }],
     },
+    // test for nested media query inside stylex.when.ancestor()
+    {
+      code: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          icon: {
+            transitionDuration: {
+              default: '0.5s',
+              [stylex.when.ancestor(":active")]: '0.1s',
+              [stylex.when.ancestor(":hover")]: {
+                default: null,
+                "@media (hover: hover)": '0.1s',
+              },
+            },
+          },
+        })
+      `,
+      options: [{ allowOuterPseudoAndMedia: true }],
+    },
+    // test for nested media query inside when.descendant() with destructured import
+    {
+      code: `
+        import { when, create } from '@stylexjs/stylex';
+        const styles = create({
+          icon: {
+            transitionDuration: {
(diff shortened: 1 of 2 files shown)


skovhus:
Codex might have found a few issues, that we should look into:

```
1. Important — packages/@stylexjs/eslint-plugin/src/stylex-valid-styles.js:450: isStylexWhenCallExpression accepts any stylex.when.<identifier>(...) /
     when.<identifier>(...) member call. The new object-valued path then skips "Keys must be strings" for typos like stylex.when.notAMethod(':hover'), so they
     lint clean if the value is an object. The public API only documents ancestor, descendant, anySibling, siblingBefore, and siblingAfter in packages/docs/
     content/docs/api/javascript/when.mdx:39. I verified this with a one-off Linter.verify(...); it currently returns no diagnostics.
  2. Important — packages/@stylexjs/eslint-plugin/src/stylex-valid-styles.js:475: the new exemption is gated only on level > 0, so it also allows whole-block
     stylex.when groups inside pseudo-elements, e.g. ::after: { [stylex.when.ancestor(':hover')]: { width: 20 } }. That shape is not a per-property conditional
     value; it slips through because the pseudo-element depth exception keeps propName null, then the new isWhenCall path suppresses key validation. The
     authoring types still model ::after as CSSProperties, not arbitrary condition blocks, in packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts:36. I
     verified this also currently produces no lint errors.
```

Issue #70, opened 2023-12-06
Community Work on an SWC Plugin


chentsulin:
It would be nice to have a SWC plugin.
Is there a plan to integrate StyleX with Rust-based [SWC](https://swc.rs/) as an alternative to Babel? Or, is there any team or individual currently working on this?

necolas:
There's no plan for SWC plugin from Meta

dgabriele:
Well, that's a bummer. I'm sure lots of people agree. I don't know the first thing it take to integrate with SWC but anyone out there who does would be greatly appreciated if they undertook it.

chentsulin:
Will the Next.js or SWC team plan to work on this, or will they leave it to the community to implement? 
@kdy1 @leerob Sorry for mentioning, but I just want to make sure that if there's anyone want to implement the plugin, it wouldn't conflict with the plan of the teams.

necolas:
Going to close this as "answered" and for the SWC community to support. Thanks

denexapp:
Just wanted to mention that it's not allowed to use `next/font` if you use the custom babel configuration, which is required if you want to use the stylex plugin.

nmn:
@denexapp We're aware. But we don't use SWC at Meta so we can't justify migrating to it. And maintaining two codebases is out of the question.

As  StyleX grows in popularity, NextJS will eventually find a way to make it work.

tkow:
I'm not so familiar with webpack compiler api, so my plugin had better to be improved, but I wrote [plugin for next swc compiler](https://github.com/tkow/stylex-next-compiler-plugin) work around modifying original next plugin. It can transpile by both swc and babel. Original plugin has problem that passes babelOption including function, causes in next swc compiler.  When  avoiding this and set babelrc option  false,  swc compiler works without babel config files.

nmn:
@tkow Will take a look at your work. We're looking to improve the integration with Next without needing a full blown SWC plugin. Since Next added support for Babel plugins alongside SWC to enable React Compiler, it should be more possible going forward.

tkow:
@nmn 
Thank you for reaction. Sorry. My previous expression may lead to misunderstand. It's babel plugin  with trivial modification from [@stylex/next-lugin](https://github.com/facebook/stylex/tree/main/packages/nextjs-plugin)  though  It looks that rewrite https://github.com/facebook/stylex/tree/main/packages/nextjs-plugin widely and unreadable I want to make it work around as soon as possible. 
[...]

PR #1623, opened 2026-04-19
[feat] SWC plugin

nmn:

Using the tests as the contracts this PR implements an entire SWC plugin (using about 3 days of Codex) in Rust. Some new tests were added to Babel, all the Babel plugin tests were ported to Rust and they all pass now.

## Additional Context

I'm looking to see if the unplugin package works when using this new plugin correctly next to build more confidence.


skovhus:
@nmn happy to see this! Let me know if you would like some help testing this out when you have something that passes the tests.

matclayton:
Likewise, stylex is now about 46% of our build time on vite, so anything we can do to speed this up is really welcome! We're deep in a conversion from Styled-Components to Stylex and the build performance is an ever growing concern!

nmn:
@matclayton we spent a lot of time making sure that StyleX transforms are strictly cache-able on a per-file basis. So if it’s a bottleneck during builds for you, try to improve caching.

@skovhus would love help testing this!

skovhus:
@nmn great, please ping me when this is ready for testing. We are currently using https://github.com/Dwlad90/stylex-swc-plugin for some of our projects.

Dwlad90:
@nmn, this is great news!

> @nmn great, please ping me when this is ready for testing. We are currently using https://github.com/Dwlad90/stylex-swc-plugin for some of our projects.

I'd love to share my experience and the results of implementing `StyleX` in Rust.

To give some background, I initially implemented [`StyleX in Rust`](https://github.com/Dwlad90/stylex-swc-plugin/) using an `SWC` plugin, but that introduced a lot of headaches:

The SWC plugin is tightly coupled to the locally installed version of `@swc/core` - it must be strictly compatible with the `swc_core` crate version used inside the plugin.

Since the SWC plugin is a WebAssembly module, it runs in a sandbox and lacks raw access to the file system. This is a massive limitation when working with paths. You only get access to the directory from which the plugin was executed (PWD), and even then, it comes with various restrictions and inconsistencies depending on the bundler and OS.

While I haven't hit this directly yet, I'm almost 100% certain that SWC plugins don't support the serialization of functions passed through plugin options (like `stylex.env`). This could seriously bottleneck the plugin's capabilities and make it difficult to maintain feature parity with the Babel plugin.

Because of this, and following @SukkaW's [recommendation](https://github.com/facebook/stylex/issues/70#issuecomment-2333287954), I decided to rewrite the plugin using `NAPI-RS` combined with `SWC` for JS parsing and AST transformation.

This instantly solved a lot of our problems. The biggest win was that `NAPI-RS` allows you to generate native JS bindings, meaning you can use it simply by swapping out the Babel plugin function call for the NAPI-RS compiler.

I'd be more than happy to help out and dive deeper into my findings if it would be valuable for the community!

PR #1644, opened 2026-05-03
[eslint-plugin] fix valid-styles rejecting nested stylex.when.* conditions

### LINKED ISSUE #1614: ESLint valid-styles rule rejects nested conditions inside stylex.when.*() computed keys

kaustubh2332:
The ObjectExpression branch in checkStyleProperty did not recognize 
stylex.when.*() computed keys, causing "Keys must be strings" errors 
when nesting a media query or other condition inside a when call.

Fixes #1614

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
@@ -10,7 +10,8 @@
 jest.disableAutomock();
 
 const { RuleTester: ESLintTester } = require('eslint');
-const rule = require('../src/stylex-valid-styles');
+//const rule = require('../src/stylex-valid-styles');
+const rule = require('../lib/stylex-valid-styles');
 
 const eslintTester = new ESLintTester({
   parser: require.resolve('hermes-eslint'),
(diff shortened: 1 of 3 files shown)


packages/@stylexjs/eslint-plugin/__tests__/stylex-valid-styles-test.js
```diff
@@ -10,7 +10,8 @@
 jest.disableAutomock();
 
 const { RuleTester: ESLintTester } = require('eslint');
-const rule = require('../src/stylex-valid-styles');
+//const rule = require('../src/stylex-valid-styles');
```
skovhus: Should no comment out

kaustubh2332: Hi @skovhus I have made the required changes

PR #1657, opened 2026-05-10
fix: use correct export syntax for default export in babel-plugin types

Dasmat13:
Fixes TS type error in babel-plugin types


henryqdineen:
I kind of forget all of intricacies but when I looked at this in the past the best solution I could find was: https://github.com/facebook/stylex/pull/1322

Dasmat13:
Thanks for pointing to #1322 @henryqdineen. I can see this overlaps with your work there. Happy to close this PR if #1322 is the preferred path forward — let me know if there's any specific piece I can help with instead.

PR #1665, opened 2026-05-13
Fix stylex-sort-keys autofix ordering

### LINKED ISSUE #787: [eslint-plugin] sort keys rule should sort object keys when applying fix
The `sort-keys` rule in `@stylexjs/eslint-plugin` correctly detects any key in a style object that is not sorted as it should be. (This is not alphabetical, but ordered according to style specificity).

The issue is with the auto-fix offered. It only moves the particular key that is not sorted correctly, but it can often leave behind additional lint warnings in the same style object.

**Expected behaviour**"
When the auto-fix is applied, the entire object should be sorted as expected such that there are no lint warnings after the auto-fix.

Jaksenc:
## Summary
- sort each contiguous autofix-safe property segment instead of only swapping the current property with the previous one
- preserve blank-line groups, spreads, comments, and existing single-line/multiline formatting boundaries
- add regression coverage for the multi-property case from #787

Fixes #787.

## Testing
- npx -y -p node@22 -p yarn@1.22.22 yarn --ignore-engines jest __tests__/stylex-sort-keys-test.js --runInBand --no-coverage
- npx -y -p node@22 -p yarn@1.22.22 yarn --ignore-engines eslint packages/@stylexjs/eslint-plugin/src/stylex-sort-keys.js packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
- npx -y -p node@22 -p yarn@1.22.22 yarn --ignore-engines flow check --show-all-errors
- git diff --check

### DIFF
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
@@ -396,6 +396,38 @@ eslintTester.run('stylex-sort-keys', rule.default, {
       ],
     },
     {
+      code: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          main: {
+            fontSize: 12,
+            animationDuration: '100ms',
+            padding: 10,
+          }
+        });
+      `,
+      output: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          main: {
+            padding: 10,
+            animationDuration: '100ms',
+            fontSize: 12,
+          }
+        });
+      `,
+      errors: [
+        {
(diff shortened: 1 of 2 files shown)


mellyeliu (approved):

Jaksenc:
I checked the failing `size` and `perf` benchmark jobs after approval. The benchmark steps completed successfully; both jobs are red because the final `peter-evans/create-or-update-comment` step hit `Resource not accessible by integration` when trying to post the benchmark comment.

The size output is unchanged across the measured bundles (`1.00` ratios throughout). The perf output shows small mixed movement (`basic create` 0.97, the other three entries 1.02-1.03). So this does not look like a code-caused benchmark failure from this PR.

PR #1677, opened 2026-05-19
[babel-plugin] Fix sx runtime stylex import injection

### LINKED ISSUE #1676: `sx={…}` on intrinsic DOM elements crashes when `stylex` is type-only imported or not imported
### Describe the issue

The JSX transform rewrites `<svg sx={x}>` to `<svg {...stylex.props(x)}>` with a `stylex` identifier — which assumes that the user actually imported `stylex`, which might not always be the case (e.g. not needing `stylex.create` or `stylex.props` inside the file).

### Expected behavior

The JSX `sx` transform should guarantee a runtime `stylex` namespace binding.

Potential fix: In the `JSXOpeningElement` visitor, before emitting `stylex.props(…)`, ensure a value-level `stylex` namespace binding exists:
- If `state.stylexImport` is non-empty, reuse its first local name.
- Otherwise, generate a hygienic uid (`program.scope.generateUid('stylex')`), prepend `import * as <uid> from "@stylexjs/stylex"` to the program body, and use `<uid>.props(…)` in the emitted call.

This guarantees the plugin owns its runtime contract.

### Steps to reproduce

```tsx
  import type * as stylex from "@stylexjs/stylex"; 
  export function Foo({ sx }: { sx?: stylex.StyleXStyles }) {
    return <svg sx={sx} />;
  }
```

same for `import type { StyleXStyles } from "@stylexjs/stylex"` or files without a StyleX import.

### Test case

_No response_

### Additional comments

_No response_

skovhus:

Fixes `sx` JSX transforms for intrinsic elements when a file has only type-only StyleX imports or no StyleX import at all. The Babel plugin now ensures a value-level StyleX namespace import exists before emitting `stylex.props(...)`, reuses existing runtime imports, and respects configured import sources.

## Linked PR/Issues

Fixes https://github.com/facebook/stylex/issues/1676

## Additional Context

n/a

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-props-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-stylex-props-test.js
@@ -10,11 +10,25 @@
 jest.autoMockOff();
 
 import { transformSync } from '@babel/core';
+import flow from '@babel/plugin-syntax-flow';
 import jsx from '@babel/plugin-syntax-jsx';
+import typescript from '@babel/plugin-syntax-typescript';
 import stylexPlugin from '../src/index';
 import path from 'path';
 
 function transform(source, opts = {}) {
+  return transformWithSyntaxPlugins(source, [flow, jsx], opts);
+}
+
+function transformTypeScript(source, opts = {}) {
+  return transformWithSyntaxPlugins(
+    source,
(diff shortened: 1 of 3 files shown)


mellyeliu (approved):

mellyeliu:
Needs some type fixes!

PR #1707, opened 2026-06-11
fix(unplugin): arithmetic support for defineConsts

skovhus:

JS arithmetic on `defineVars`/`defineConsts` token references inside `stylex.create()` silently produced broken CSS (`C.a + C.b` → `"var(--a)var(--b)"`, `C.a - 1` → `NaN`).

This PR compiles arithmetic to CSS `calc()` instead — **without violating the "transforming a file must not read other files" constraint**: the emitted expression is built from the same hash-derived `var()` strings as today, and `defineConsts` values are substituted by the existing CSS aggregation step.

 - `zIndex: C.a + C.b` → `calc(var(--a) + var(--b))` (→ `calc(26 + 14)` for consts)
 - `zIndex: C.elevated - 1` → `calc(var(--elevated) - 1)`
 - `marginLeft: -C.gutter` → `calc(-1 * var(--gutter))`
 - Works for `defineConsts`, `defineVars` (incl. sibling refs), and mixed expressions

 Everything that can't map to `calc()` is now a **compile-time error**.
 
### Breaking change

⚠️  Comparisons on token refs (e.g. `C.a === 'red' ? x : y`) previously evaluated silently (string comparison, usually picking the wrong branch) and now fail the build with an actionable error. Null/undefined guards are exempt and keep working.

## Linked PR/Issues

Fixes #1597

## Additional Context

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -1230,4 +1230,98 @@ describe('@stylexjs/babel-plugin', () => {
       `);
     });
   });
+
+  describe('[transform] arithmetic on imported defineConsts (#1597)', () => {
+    function transformCrossFile(mainSource) {
+      const pluginOpts = {
+        unstable_moduleResolution: { type: 'haste' },
+      };
+
+      const tokens = transformSync(
+        `
+        import * as stylex from '@stylexjs/stylex';
+        export const consts = stylex.defineConsts({
+          A: 26,
+          B: 14,
+          D: 6,
+          gutter: '16px',
+        });
+        export const vars = stylex.defineVars({
+          gap: '8px',
+        });
+        `,
+        {
+          filename: '/src/app/constants.stylex.js',
+          parserOpts: { flow: 'all' },
+          babelrc: false,
+          plugins: [[stylexPlugin, pluginOpts]],
+        },
+      );
+
+      const main = transformSync(mainSource, {
+        filename: '/src/app/main.js',
+        parserOpts: { flow: 'all' },
+        babelrc: false,
+        plugins: [[stylexPlugin, pluginOpts]],
+      });
+
+      return [
+        ...(tokens.metadata.stylex || []),
+        ...(main.metadata.stylex || []),
+      ];
+    }
+
+    test('numeric const arithmetic resolves to calc() with literal values', () => {
+      const metadata = transformCrossFile(`
+        import * as stylex from '@stylexjs/stylex';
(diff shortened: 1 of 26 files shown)


skovhus:
@nmn if we don't want to support arithmetic for some reason (would be nice if we could), then at least I think we should  fail hard instead of silently emitting broken CSS.

skovhus:
We just had another production issue due to the underlying issue. @nmn @mellyeliu would love some feedback on the approach on this branch. Happy to put up a PR that fails hard if we don't want arithmetic.

mellyeliu:
Thanks for fixing this, overall this implementation looks fine and keeps everything per-file. Mind adding some more tests for different arithmetic operations (a + b * c, (a + b) * 2, with brackets, etc) plus a couple of string-concatenation cases like 'translateX(' + token + ')'?

skovhus:
Another foot gun that I'll need to see if we can solve here:

```tsx
// $zIndex.dialog is defined as a number

const PRESENTER_Z_INDEX = Number($zIndex.dialog) + 1;

  presenter: {
    zIndex: PRESENTER_Z_INDEX, // silently breaks
  },
});
```

Edit: Fixed ✅

skovhus:
@nmn @mellyeliu anything else you need from me here? Again, if we don't go this direction (which I think we should), then at least we should make the compiler fail hard when discovering arithmetic that will end up silently failing at runtime.

skovhus:
Another foot gun that just silently crashed in production:

```tsx
// Constants comes from stylex.ts file and is a defineConsts (41)
const scrollMarginTop = Constants.height + 32;
  
  container: {
    // ends up as 4132!
    scrollMarginTop: scrollMarginTop,
```

mellyeliu (approved):

packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
```diff
@@ -1230,4 +1230,116 @@ describe('@stylexjs/babel-plugin', () => {
...
+          box: {
+            marginTop: consts.A * vars.gap,
+          },
+        });
+      `);
+
+      const css = stylexPlugin.processStylexRules(metadata, {
+        useLayers: false,
+      });
+      expect(css).toMatch(/margin-top:calc\(26 \* var\(--[a-z0-9]+\)\)/);
```
nmn: snapshots of the transformed code would be better here.

skovhus: Done

skovhus:
@nmn how do we feel about this one?

PR #1731, opened 2026-06-22
[babel-plugin] Fix corrupted class names from null coercion in dynamic styles

### LINKED ISSUE #1702: Corrupted class names including string "null" generated from dynamic style with media query conditions
### Describe the issue

When using `stylex.create()` with dynamic style functions having a media query condition evaluating to `null`, StyleX generates class names that incorrectly include the raw string "null" in them leading to corrputed name, like `"nullxafpxz5"`. This corrupted class name results respective style getting completely ignored. According to the StyleX documentation, null values should be ignored and not included in the generated class names.

### Expected behavior

Conditional media query styles that return `null` should be ignored.

### Steps to reproduce

Following component should have `background-color: tomato` when viewport is wider than 600px:

```jsx

  highlightAt: (small, large) => ({
    backgroundColor: {
      default: null,
      "@media (max-width: 599px)": small ? "red" : null,
      "@media (min-width: 600px)": large ? "red" : null,
    },
  }),
});

const MyComponent = () => (
  <div {...stylex.props(styles.highlightAt(false, true))}>
    I should be red when viewport width is more than 600px
  </div>
);
```

- **StyleX version**: 0.18.3
- **Plugins**: @stylexjs/babel-plugin, @stylexjs/postcss-plugin
- **Environments tested**: Astro.js dev server and production builds, Storybook

### Test case

Reproduction steps work as deterministic test case.

### Additional comments

Behavior is likely due to transpilation issue that can be reduced to very minimal Node.js program when standard stylex dependencies (@babel/core @babel/preset-react @stylexjs/babel-plugin) are in place. Self-contained reproduction repository: https://github.com/joonaojapalo/stylex-dynamic-mediaquery-repro

```js
// repro.js
const babel = require("@babel/core");

const input = `

    highlightAt: (small, large) => ({
      backgroundColor: {
        "@media (max-width: 599px)": small ? "red" : null,
        "@media (min-width: 600px)": large ? "red" : null,
      },
    }),
  });

  <div {...stylex.props(styles.highlightAt(false, true))} />
`;

const output = babel.transformSync(input, {
  presets: [["@babel/preset-react"]],
  plugins: [["@stylexjs/babel-plugin"]],
});

console.log(output.code);
```

This outputs transpiled code with following essential part:

```js
 // ... redacted ..
  highlightAt: (small, large) => [{
    kWkggS: (
            (small ? "red" : null) != null
              ? "xkcfsz0 "
              : small
                ? "red"
                : null
        ) +  // <!-- string concatenation leads to invalid css, eg. highlightAt(false, true) -> {"kWkggS": "nullxafpxz5" }
        (
            (large ? "red" : null) != null 
                ? "xafpxz5"
                : large
                    ? "red"
                    : null
        ),
 // ... redacted ..
```

henryqdineen:

When a dynamic style has multiple conditional class names for a single CSS property (e.g. multiple media queries or pseudo-classes), the generated code joins them with `+` string concatenation. If any condition evaluates to `null` or `undefined` at runtime, JavaScript coerces it to the string `"null"` or `"undefined"`, producing corrupted class names like `"nullxafpxz5"`.

The fix changes the else branch of the null-check ternary from `expr` to `t.stringLiteral('')`, so a non-matching condition contributes nothing to the concatenation instead of contributing `"null"`.

This is a regression from 7a1ed95a which refactored the single-class and multi-class code paths into one. The original code before that refactor already used `""` for the multi-class case — the refactor replaced it with `expr`, likely reasoning that `expr` would be null anyway in the else branch. That's true, but `null + "string"` produces `"nullstring"` in JavaScript.

As a side benefit, this also produces smaller output since `""` replaces what was previously a duplicated copy of the original expression in the else branch (unless the expression is a single-character variable).

Note that `""` and `null` are not identical in styleq's merge logic — `null` is the intentional "unset" semantic while `""` is treated as an empty string value. In practice they behave the same (both claim the property slot, neither adds a real class name), and the all-null case was already broken before this fix (producing `"nullnull"`), so this is strictly an improvement.

## Linked PR/Issues

Fixes #1702

## Additional Context

Existing snapshot tests updated to reflect the new `""` fallback. All tests pass.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
@@ -402,7 +402,7 @@ describe('Evaluation of imported values works based on configuration', () => {
         });
         const styles = {
           color: color => [{
-            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__1w8wjxo" : color,
+            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__1w8wjxo" : "",
             $$css: true
           }, {
             "--x---__hashed_var__1jqb1tb": color != null ? color : undefined
--- a/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
@@ -401,7 +401,7 @@ describe('Evaluation of imported values works based on configuration', () => {
           priority: 0
         });
         const styles = {
-          color: color => [{
+          color: color => [color !== undefined && {
             "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__1w8wjxo" : "",
             $$css: true
           }, {
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -240,7 +240,7 @@ describe('@stylexjs/babel-plugin', () => {
             $$css: "components/main.js:71"
           },
           dynamic: color => [{
-            "color-kMwMTN": color != null ? "color-x14rh7hd" : color,
+            "color-kMwMTN": color != null ? "color-x14rh7hd" : "",
             $$css: "components/main.js:74"
           }, {
             "--x-color": color != null ? color : undefined
@@ -333,7 +333,7 @@ describe('@stylexjs/babel-plugin', () => {
             $$css: "main.js:71"
           },
           dynamic: color => [{
-            "color-kMwMTN": color != null ? "color-x14rh7hd" : color,
+            "color-kMwMTN": color != null ? "color-x14rh7hd" : "",
             $$css: "main.js:74"
           }, {
             "--x-color": color != null ? color : undefined
@@ -781,7 +781,7 @@ describe('@stylexjs/babel-plugin', () => {
             $$css: "main.js:71"
           },
           dynamic: color => [{
-            "color-kMwMTN": color != null ? "color-x14rh7hd" : color,
+            "color-kMwMTN": color != null ? "color-x14rh7hd" : "",
             $$css: "main.js:74"
           }, {
             "--x-color": color != null ? color : undefined
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-process-test.js
@@ -239,7 +239,7 @@ describe('@stylexjs/babel-plugin', () => {
             "--orange-theme-color": "--orange-theme-color-xufgesz",
             $$css: "components/main.js:71"
           },
-          dynamic: color => [{
+          dynamic: color => [color !== undefined && {
             "color-kMwMTN": color != null ? "color-x14rh7hd" : "",
             $$css: "components/main.js:74"
           }, {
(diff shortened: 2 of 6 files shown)


packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
```diff
@@ -402,7 +402,7 @@ describe('Evaluation of imported values works based on configuration', () => {
         });
         const styles = {
           color: color => [{
-            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__1w8wjxo" : color,
+            "--__hashed_var__1jqb1tb": color != null ? "__hashed_var__1w8wjxo" : "",
```
nmn, line 405: This is a small regression because the reason the fallback value was `color` is to account for the way `undefined` vs `null` is handled by stylex. `null` means intentionally unsetting a value. `undefined` is to ignored and the previously set value wins. Here all nullish values are represented as `""` which behaves like `null`.

nmn: Perhaps what we need here is:

```tsx
color !== undefined && {"--__hashed_var__1jqb1tb": color != null ? "__hashed_var__1w8wjxo" : "", $$css: true}
```

This way, we can skip the object for undefined entirely.

nmn (changes requested):
Either way can change the way the object is generated in the way I suggested, or we can change the runtime to fix how we merge.

packages/@stylexjs/babel-plugin/__tests__/evaluation-import-test.js
henryqdineen: Ah, thanks for catching that! I worked through it with Claude to implement your suggestion.

From Claude 🤖 :

> Implemented as suggested: a property that resolves to a single conditional class is now emitted as `expr !== undefined && { … }`. When the value is `undefined` the object is skipped (styleq drops `false` styles) so the previously set value wins; when it's `null` the class still resolves to `""` to unset the slot. That restores the `undefined` vs `null` distinction the `""`-only fallback had flattened. So `(color) => ({ color })` now compiles to:
>
> ```js
> color => [color !== undefined && { kMwMTN: color != null ? "…" : "", $$css: true }, { "--x-color": … }]
> ```
>
> Two things shaped the approach:
>
> **Why not just restore the original `expr` fallback?** With `color != null ? cls : color`, passing `undefined` leaves a literal `undefined` as a value in the `$$css` object — and styleq errors on that (`console.error(… is not "string" or "null")`), since it only accepts a string or `null` there. Guarding the object with `!== undefined` means `undefined` never ends up inside it.
>
> **Why one guarded object per property, not one around the whole style?** When two variables drive the same style, a single shared guard is wrong — it would drop a set value whenever the *other* is `undefined`. So each property is guarded on its own value:
>
> ```js
> // (bg, oc) => ({ backgroundColor: bg, outlineColor: oc })
> [bg !== undefined && { kWkggS: … }, oc !== undefined && { kjBf7l: … }, { … }]
> ```
>
> (Multi-class concatenations — e.g. several media queries on one property — keep the `""` fallback instead: the slot is already claimed by the sibling classes, and `undefined` vs `null` can't be expressed inside a joined string anyway.)
>
> Possible followup: properties driven by the *same* value could be grouped into one guarded object rather than one-per-property. Left out to keep the diff focused.
>
> — Claude

nmn (approved):

Issue #1407, opened 2025-12-22
[fix] Better media query sorting


mellyeliu:
Right now we sort media queries by alphanumeric order, leading to unexpected media query ordering and behaviour for overlapping breakpoints. The config `enableMediaQueryOrder` exists to tackle this properly but it has some unresolved bugs and does not yet work with `defineConsts`. We can improve media query ordering by sorting queries in a way that smaller breakpoints (covering less range) win over larger ones.

For example, as a starting point:
- For max-width, sort from smallest number to largest
- For min-width, the inverse

Steps:
- Modify block here https://github.com/facebook/stylex/blob/main/packages/%40stylexjs/babel-plugin/src/index.js#L459-L482
- Look for @media rules and sort as above

PR #1756, opened 2026-07-13
fix(eslint-plugin): respect enableMediaQueryOrder in sort-keys

### LINKED ISSUE #1416: `@stylexjs/sort-keys` ESLint rule breaks max-width media query cascade when using `enableMediaQueryOrder`
## Description

The `@stylexjs/sort-keys` ESLint rule sorts style properties alphabetically, including media query keys inside responsive objects. This conflicts with CSS cascade requirements for `max-width` media queries, where larger breakpoints must appear before smaller ones in source order.

Even when following all documented best practices:
- Using `enableMediaQueryOrder: true` (as recommended in v0.15.0+)
- Using inline string literals for media queries (required for Vite compatibility)
- Ordering breakpoints correctly for max-width cascade (larger first)

Running `eslint --fix` will re-sort the media queries alphabetically, breaking the cascade.

### Setup (following best practices)

**vite.config.js:**
```javascript
styleXUnplugin.vite({
  enableMediaQueryOrder: true,  // Recommended in v0.15.0+
})
```

**eslint.config.mjs:**
```javascript
{
  plugins: { "@stylexjs": stylexPlugin },
  rules: {
    "@stylexjs/sort-keys": "warn",  // Recommended StyleX ESLint rule
  },
}
```

### The Problem

**Correct code (larger breakpoint first for max-width cascade):**
```javascript
  container: {
    padding: {
      default: "2rem",
      "@media (max-width: 768px)": "1.5rem",  // tablet
      "@media (max-width: 640px)": "1rem",    // mobile (overrides tablet)
    },
  },
});
```

**After running `eslint --fix`:**
```javascript
  container: {
    padding: {
      default: "2rem",
      "@media (max-width: 640px)": "1rem",    // now first (alphabetically)
      "@media (max-width: 768px)": "1.5rem",  // now last - WINS on mobile!
    },
  },
});
```

### Why This Breaks

With `enableMediaQueryOrder: true`, StyleX makes "later source order wins." After ESLint sorts:
- `640px` appears first in source
- `768px` appears last in source → takes precedence
- On a 500px viewport, BOTH queries match, but `768px` wins because it's later
- Result: Mobile devices get `1.5rem` instead of `1rem`

## Expected Behavior

The `@stylexjs/sort-keys` rule should either:

1. **Skip sorting inside objects containing media query keys** - Detect `@media` keys and preserve author order
2. **Sort media queries by specificity** - For `max-width`, sort descending (larger first); for `min-width`, sort ascending (smaller first)
3. **Provide a configuration option** - Allow disabling sort-keys for media query objects

## Current Workaround

We must add ESLint disable blocks around every responsive property:

```javascript
  container: {
    /* eslint-disable @stylexjs/sort-keys -- max-width cascade: larger breakpoints first */
    padding: {
      default: "2rem",
      "@media (max-width: 768px)": "1.5rem",
      "@media (max-width: 640px)": "1rem",
    },
    /* eslint-enable @stylexjs/sort-keys */
  },
});
```

This is tedious and error-prone, especially since the ESLint warning says:
> StyleX property key "@media (max-width: 640px)" should be above "@media (max-width: 768px)"

This message actively encourages the wrong order for max-width queries.

## Environment

- `@stylexjs/stylex`: 0.10.1
- `@stylexjs/eslint-plugin`: 0.10.1
- `@stylexjs/unplugin`: 0.6.0
- Vite: 6.0.5
- Node: 22.x

## Related Issues

- #415 - autofix for @stylexjs/sort-keys
- #517 - Detect intended Media query boundaries by order
- #82 - Set CSS rule priority based on media query break points

gunjanjaswal:
## What

Adds an `enableMediaQueryOrder` option to the `@stylexjs/sort-keys` rule. When it is on, the rule keeps the authored order of `@media` condition keys instead of sorting them alphabetically. It defaults to `false`, so existing behavior is unchanged.

## Why

Fixes #1416. When a project compiles with StyleX's `enableMediaQueryOrder`, overlapping media queries are resolved by source order, so the last matching query wins. Sorting `@media` keys alphabetically inverts that cascade. A correctly authored `max-width: 768px` followed by `max-width: 640px` gets flipped by `--fix`, so mobile viewports pick up the tablet value. Right now the only workaround is wrapping every responsive property in an `eslint-disable` block.

## How

All `@media` keys share a single at-rule priority, so the comparator was falling through to an alphabetical tiebreak. With the option on, two media-query keys are treated as already ordered when reporting and compared as equal when autofixing (the sort is stable and falls back to the original index), so their source order is kept. Non-media keys still sort normally. This matches how the compiler orders queries by source order, and it can fall back to the #1407 breakpoint sort once that merges.

## Tests

Added regression tests to `stylex-sort-keys-test.js`: valid cases for max-width and min-width cascades under the option, an invalid case confirming the default still reorders, and one confirming regular keys still sort with the option on. `jest stylex-sort-keys` passes 59/59.

### DIFF
--- a/packages/@stylexjs/eslint-plugin/README.md
+++ b/packages/@stylexjs/eslint-plugin/README.md
@@ -57,11 +57,21 @@ This rule helps to sort the StyleX property keys according to
 
 #### Config options
 
-```json
+```js
 {
-  "validImports": ["stylex", "@stylexjs/stylex"]
+  order: 'default',                            // Sort order: 'default', 'clean', or 'recess'
+  minKeys: 2,                                  // Minimum number of keys before sorting is enforced
+  allowLineSeparatedGroups: false,             // Treat blank-line-separated blocks as independent groups
+  enableMediaQueryOrder: false,                // Preserve authored `@media` order to match StyleX's `enableMediaQueryOrder`
+  validImports: ['stylex', '@stylexjs/stylex']
 }
 ```
+
+When you compile with StyleX's `enableMediaQueryOrder`, overlapping media
+queries are resolved by source order (the last matching query wins). Sorting
+`@media` keys alphabetically inverts that cascade, so set
+`enableMediaQueryOrder: true` here to keep the rule from reordering media
+queries and let you control the cascade through their authored order.
 
 ### @stylexjs/valid-shorthands
 
--- a/packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
+++ b/packages/@stylexjs/eslint-plugin/__tests__/stylex-sort-keys-test.js
@@ -365,6 +365,39 @@ eslintTester.run('stylex-sort-keys', rule.default, {
         },
       });`,
     },
+    {
+      // With `enableMediaQueryOrder`, authored media query order is preserved,
+      // so a max-width cascade written largest-first is reported as valid.
+      options: [{ enableMediaQueryOrder: true }],
+      code: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          container: {
+            padding: {
+              default: '2rem',
+              '@media (max-width: 768px)': '1.5rem',
+              '@media (max-width: 640px)': '1rem',
+            },
+          },
+        });
+      `,
+    },
+    {
+      // min-width breakpoints authored smallest-first are also left untouched.
+      options: [{ enableMediaQueryOrder: true }],
+      code: `
+        import * as stylex from '@stylexjs/stylex';
+        const styles = stylex.create({
+          container: {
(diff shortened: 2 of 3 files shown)


nmn (approved):
LGTM

gunjanjaswal:
Thanks again for approving this, @nmn. It's been green and approved for about a month now — is there anything still needed from me before it can land, or is it just waiting on an import? The only red is the perf/size benchmark jobs, which look non-blocking, but happy to rebase or dig into those if it helps.

gunjanjaswal:
Thanks for landing it, @mellyeliu — and @nmn for the review. Appreciate it!

PR #1758, opened 2026-07-14
Add no-raw-media-queries rule to eslint-plugin

### LINKED ISSUE #725: ESLint Rule to enforce usage of shared media queries and other constants

IBlackVoid:

Adds a new opt-in lint rule, `@stylexjs/no-raw-media-queries`, that flags raw media query strings inside `stylex.create` calls and suggests defining them once with `stylex.defineConsts` and referencing the shared constant instead.

Without this rule nothing stops hardcoded breakpoints from drifting across a codebase (five slightly different "mobile" queries, etc.). Now that `defineConsts` is shipped, this rule makes the shared-constants pattern enforceable.

What it catches:

- raw string keys like `'@media (min-width: 600px)'` at any nesting depth (modern conditional-value syntax and the legacy contextual syntax)
- expressionless template literal keys, e.g. ``[`@media (min-width: 600px)`]``
- `@MEDIA`-style casing (matching is case-insensitive)
- dynamic styles declared as arrow functions, e.g. `root: (width) => ({ ... })`

What it deliberately leaves alone:

- computed keys referencing constants, e.g. `[breakpoints.small]` — that's the pattern being encouraged
- media queries as *values* inside `stylex.defineConsts` — that's where they belong
- `create` calls from imports not listed in `validImports`

Options follow the plugin's conventions: `validImports` (same shape as the other rules) plus an `allowedMediaQueries` exact-match allowlist so large codebases can migrate incrementally.

Scope notes for v1, happy to adjust:

- Only `stylex.create` is checked. `defineVars` / `createTheme` condition keys could be covered in a follow-up if you'd like.
- Only `@media` is flagged (per the rule name). `@supports` / `@container` could be added behind an option.
- The "Advanced: file tree knowledge" idea from the issue (discovering near-equal constants across the codebase) is intentionally out of scope here.

Also happy to rename the rule if you prefer something broader like `enforce-consts`.

## Linked PR/Issues

Fixes #725

## Additional Context

- 16 new RuleTester cases in `__tests__/stylex-no-raw-media-queries-test.js`; full eslint-plugin suite passes (565 tests, 10 suites)
- Documented in the package README and the docs site (`eslint-plugin.mdx`), matching the format of the existing rules
- Side note: #724 is still open but `defineConsts` appears to have shipped (v0.13) — it may be safe to close

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/eslint-plugin/README.md
+++ b/packages/@stylexjs/eslint-plugin/README.md
@@ -273,3 +273,65 @@ This rule disallows using `className` or `style` props on elements that spread
   "validImports": ["stylex", "@stylexjs/stylex"]
 }
 ```
+
+### `@stylexjs/no-raw-media-queries`
+
+This rule disallows raw media query strings within `stylex.create` calls. It
+helps keep breakpoints consistent across a codebase by requiring media queries
+to be defined once with
+[`stylex.defineConsts`](https://stylexjs.com/docs/api/javascript/defineConsts/)
+and referenced as shared constants.
+
+#### Invalid example
+
+```js
+const styles = stylex.create({
+  root: {
+    width: {
+      default: '100%',
+      '@media (max-width: 600px)': '50%',
+    },
+  },
+});
+```
+
+#### Instead...
+
+```js
+// breakpoints.stylex.js
+export const breakpoints = stylex.defineConsts({
+  small: '@media (max-width: 600px)',
+});
+```
+
+```js
+import { breakpoints } from './breakpoints.stylex';
+
+const styles = stylex.create({
+  root: {
+    width: {
+      default: '100%',
+      [breakpoints.small]: '50%',
+    },
+  },
+});
+```
+
+#### Config options
+
+```json
+{
+  "validImports": ["stylex", "@stylexjs/stylex"],
+  "allowedMediaQueries": []
+}
+```
+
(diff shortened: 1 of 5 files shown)


IBlackVoid:
@nmn — this implements #725, which you opened and labeled `good first issue`. Linking it explicitly in case the PR never surfaced against the issue.

GitHub tags this as 534 lines, but the reviewable surface is a lot smaller than that:

| | lines |
|---|---|
| rule implementation | 179 |
| plugin registration | 3 |
| tests (16 RuleTester cases) | 268 |
| README + docs site | 84 |

To keep this a yes/no rather than a scoping discussion, I'll settle the open questions from the description rather than leaving them with you:

- **Name stays `no-raw-media-queries`.** It says what it flags, and it matches the existing `no-unused` / `no-legacy-contextual-styles`.
- **Scope stays `stylex.create` and `@media`.** `defineVars` / `createTheme` condition keys and `@supports` / `@container` are follow-up PRs, not blockers for this one.
- **Opt-in, off by default**, so no one's build changes on upgrade. `allowedMediaQueries` is there for incremental migration in large codebases.

Full eslint-plugin suite passes — 565 tests, 10 suites.

If it would be easier to review, I'm glad to split the docs into a separate PR so this one is just the rule and its tests. Happy to do that on request.

Issue #1766, opened 2026-07-23
Migration codemods from popular styling libraries to StyleX


ahmedsadid:
## The idea (thanks to @mellyeliu )

A set of codemods that translate popular styling solutions (Emotion, MUI, CSS
Modules, styled-components, …) into StyleX. They share one library-agnostic
engine; each source is a swappable adapter, so adding a library doesn't change
the engine. Each converts the safe, common cases mechanically and leaves a
clear `// TODO(stylex-migration: …)` marker on anything it can't prove is safe. 
So never guessing.

## First up: Emotion → StyleX

Built incrementally by usage form:

- [x] Object syntax — `css={{…}}`, `css({…})`, object `styled.div({…})`
- [ ] Template-literal / string styles — `` css`…` ``
- [ ] `styled()` components
- [ ] `@emotion/css` (class-name API) + the long tail (`cx`, `<Global>`)

Ships as a CLI with a dry-run-by-default preview. I'll link the PRs here as
they go up.

## After Emotion

Other libraries as additional adapters on the same engine — MUI and CSS Modules
are the natural next candidates.

## Questions for maintainers

- Package name/location — does `@stylexjs/codemods` work?
- Is `jscodeshift` OK as a dependency (isolated behind one wrapper)?

ahmedsadid:
Closing stale PRs to move migration codemod off-repo.

PR #1768, opened 2026-07-23
[codemods] Scaffold @stylexjs/codemods package + verification checks (Emotion→StyleX, 1/6)

ahmedsadid:

Adds the `@stylexjs/codemods` package — the foundation for an Emotion → StyleX
migration codemod (#1766). First of ~6 stacked PRs; this one is scaffolding and
test infrastructure only, with no transform logic yet, so it can be reviewed on
its own.

It sets up the package and the three checks every future conversion has to pass:

- the output compiles through `@stylexjs/babel-plugin`,
- it passes `@stylexjs/eslint-plugin` at error with no autofixes, and
- its CSS matches the original Emotion output (via `@emotion/serialize`).

Each check is proven to fail on a deliberately-broken input, not only to pass on
a valid one.

The package is split into a shared engine (`src/core/`) and per-library adapters
(`src/adapters/`, Emotion first). `core/` never imports from `adapters/`,
enforced by a test — so a new source library means a new adapter, not an engine
change. The transform itself lands in the next PR.

## Linked PR/Issues

n/a

## Additional Context

6 test suites, all green. Two end-to-end fixture assertions are skipped until the
transform lands (PR 2). Changes outside the package: a `.flowconfig` mapping and
an `.eslintrc.js` ignore for the test fixtures, small `flow_modules/` type
stubs, and `yarn.lock`.

## Pre-flight checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -46,6 +46,9 @@ module.exports = {
     '**/*.d.ts',
     '**/pages.gen.ts',
     '**/devtools-extension/extension/**',
+    // codemod fixtures are deliberately non-conforming (broken-on-purpose
+    // inputs that the correctness gates must fail on)
+    '**/codemods/__fixtures__/**',
   ],
   overrides: [
     {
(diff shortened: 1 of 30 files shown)


ahmedsadid:
image

Prettier CI: failure is a pre-existing upstream line in eslint-plugin/.../cssProperties.js (#1760), untouched by this stack — unrelated to every PR in the series.

PR #1787, opened 2026-07-29
[eslint-plugin] Fix no-unused false positive for styles exported via export specifier

### LINKED ISSUE #1786: [eslint-plugin] no-unused reports every style as unused when the styles object is exported via an export specifier
### Describe the issue

`@stylexjs/no-unused` treats a `stylex.create()` object as entirely unused when it is
exported with the specifier form (`export { styles };`), even though another module may
consume any of its keys.

The rule already exempts the other export forms — `export const styles = stylex.create({...})`,
`export default styles` — so the specifier form looks like an oversight rather than intent.

Because the rule is `fixable: 'code'`, running ESLint with `--fix` **deletes the style keys
from the source file**. So this is not just noise: it silently removes working styles.

### Expected behavior

No report. An exported styles object may have its keys used by another module, so the rule
cannot know whether they are unused — which is exactly why the other export forms are
already exempt.

### Behaviour by export form

| Source | Reported? | Correct? |
| --- | --- | --- |
| `export const styles = stylex.create({...})` | no | ✅ |
| `export default styles` | no | ✅ |
| `const styles = stylex.create({...}); export { styles };` | **yes** | ❌ false positive |
| `const styles = stylex.create({...}); export { styles as s };` | **yes** | ❌ false positive |

### Steps to reproduce

```js

  main: { color: 'red' },
  secondary: { color: 'blue' },
});

export { styles };
```

With `'@stylexjs/no-unused': 'error'`:

```
error  Unused style detected: styles.main       @stylexjs/no-unused
error  Unused style detected: styles.secondary  @stylexjs/no-unused
```

`eslint --fix` strips both keys, rewriting the file to:

```js

});

export { styles };
```

Same behaviour for the renamed form, `export { styles as sharedStyles };`.

### Test case

_No response_

### Additional comments

_No response_

trofim-samusev:
Fixes #1786

`@stylexjs/no-unused` reports every key of a `stylex.create()` object as unused when the
object is exported using the specifier form:

```js

  main: { color: 'red' },
  secondary: { color: 'blue' },
});

export { styles }; // => "Unused style detected: styles.main" (+ styles.secondary)
```

The rule already exempts `export const styles = stylex.create({...})` and
`export default styles`, on the basis that an exported styles object may have its keys used
by another module. `export { styles };` and `export { styles as s };` are the same situation
but were not covered.

The rule is `fixable: 'code'`, so `eslint --fix` acted on the false positive and **deleted
the style keys from the source file**, emptying the `create()` call in the example above to
`stylex.create({\n});`.

Cause: the `ExportNamedDeclaration` visitor only looked at `node.declaration`. For
`export { styles };` that is `null` and the exported bindings live in `node.specifiers[]`,
so the variable was never removed from `stylexProperties` and `Program:exit` reported all of
its keys.

The fix iterates `node.specifiers` and exempts each `local` name. Re-exports
(`export { styles } from './other-styles'`) are skipped via the `node.source == null` guard,
because they export another module's binding and must not exempt a same-named local
variable.

## Linked PR/Issues

Fixes # (issue)

## Additional Context

Tests added to `packages/@stylexjs/eslint-plugin/__tests__/stylex-no-unused-test.js`:

- valid: `export { styles };`
- valid: `export { styles as sharedStyles };`
- valid: `export { theme, styles };` (specifier list with an unrelated binding)
- valid: `export { styles as default };`
- invalid: `export { styles } from './other-styles';` — still reports the locally unused key,
  covering the re-export guard

No documentation change: neither the plugin README nor the docs site describes the rule's
export handling.

## Pre-flight checklist

- [x] I have read the contributing guidelines [Contribution Guidelines](https://github.com/facebook/stylex/blob/main/.github/CONTRIBUTING.md)


packages/@stylexjs/eslint-plugin/src/stylex-no-unused.js
```diff
@@ -209,6 +209,15 @@ const stylexNoUnused = {
 
       // Exempt used styles: export const exportStyles = stylex.create({});
       ExportNamedDeclaration(node: ExportNamedDeclaration) {
+        // Exempt used styles: export { exportStyles };
```
henryqdineen, line 212: This is an edge case but should we make sure there are no false positives for 
`export type { styles };` or `export { type styles };` by checking the exportKind on the declaration and specifiers?

trofim-samusev: Added support for those cases. Can you please check again when you'll have some time?

PR #1791, opened 2026-08-03
[unplugin] Add `babel` option to pass through to internal transformAsync

### LINKED ISSUE #1792: [unplugin] Internal Babel pass discovers root babel.config.* — no opt-out
## Describe the issue

`@stylexjs/unplugin` runs its own internal Babel pass over every StyleX file (to apply the StyleX Babel plugin) by calling `@babel/core`'s `transformAsync`. That call hardcodes `babelrc: false`, but **not** `configFile: false` — so Babel still discovers and applies any root `babel.config.js` / `.cjs` / `.mjs` / `.ts` present in the project.

For any project with a root Babel config (Vite templates, Next.js apps, monorepos with a shared Babel config, most non-toy setups), this silently causes every plugin and preset in that root config — `react-compiler`, `relay`, `module-resolver`, Flow/TS stripping, etc. — to run against every StyleX file **inside the StyleX pass**, in addition to running normally in the consumer's main Vite/Rollup/webpack pipeline.

Symptoms range from wasted CPU (we measured ~15-30% of a Vite dev warm build going to this double-transform) to real correctness bugs (React Compiler running twice, `module-resolver` producing subtly different rewrites depending on which pass runs first).

There is currently no way to opt the StyleX Babel pass out of this — the `transformAsync` options aren't exposed on the plugin's user options at all.

## Expected behavior

Consumers should be able to opt the plugin's internal Babel pass out of root-config discovery — for example via `configFile: false` — without having to fork or `pnpm patch` the plugin. Ideally the plugin would also default to `configFile: false` (since the plugin's caller shouldn't need the consumer's root Babel config to do its work), but that's a bigger behavioural change; the immediate fix is just to make it configurable.

## Steps to reproduce

Package versions: `@stylexjs/unplugin@0.19.0`, `@babel/core@7.x`.

1. Create a Vite project that uses `@stylexjs/unplugin/vite`.
2. Add a root `babel.config.cjs` (typical for a React-Compiler / Relay project):
   ```js
   module.exports = {
     plugins: ['babel-plugin-react-compiler', 'relay'],
   };
   ```
3. Add any file that imports `@stylexjs/stylex` and calls `stylex.create(...)`.
4. Instrument either plugin (a `console.log` in the plugin's `visitor` or wrapping the plugin function) to log when it runs, or profile the build.

**Observed:** `react-compiler` and `relay` run once during the StyleX unplugin's internal Babel pass, and again in the main Vite React pipeline. Same for any other plugin/preset in the root config.

**Expected:** the StyleX unplugin's internal Babel pass should not apply plugins/presets the caller didn't ask for — or at minimum should expose a way to opt out.

## Test case

Any Vite + React-Compiler + StyleX project reproduces it. The direct code path is [`packages/@stylexjs/unplugin/src/core.js`](https://github.com/facebook/stylex/blob/main/packages/%40stylexjs/unplugin/src/core.js) in `runBabelTransform`:

```js
const result = await transformAsync(inputCode, {
  babelrc: false,     // only disables .babelrc* files; root babel.config.* still discovered
  filename,
  presets,
  plugins: [...],
  caller: {...},
});
```

## Additional comments

I've opened #1791 with a proposed fix: a `babel` option on the plugin's user options that is spread into `transformAsync` before the plugin's mandatory options (so consumers can extend but not override the plugin's internals). The primary use case is `babel: { configFile: false }`.

Happy to iterate on API naming or approach — an alternative would be to default `configFile: false` inside the plugin and have consumers opt back **in** if they want the root config applied. That's more aggressive but arguably more correct. Flagging both for maintainer preference.

We've been running the patched version at Mixcloud for a few weeks and it cleanly resolves both the perf and correctness issues.

jonreading81:
Fixes #1792.

## Summary

Adds a `babel` option to `@stylexjs/unplugin` that is passed through to the internal `@babel/core` `transformAsync` call — so consumers can set options like `configFile: false` without having to fork the plugin.

## Problem

`@stylexjs/unplugin` runs its own internal Babel pass over every StyleX file to apply the StyleX Babel plugin. Today that pass hardcodes `babelrc: false`, which prevents `.babelrc*` discovery — but **does not** prevent Babel from discovering a root `babel.config.*`.

For any project that has a root `babel.config.js`/`.cjs`/`.ts` (which is most non-toy projects — Vite templates, Next.js apps, and monorepos with a shared Babel config), the plugin's internal Babel call ends up running **every plugin and preset in that root config** against every StyleX file. In practice that means transforms like `react-compiler`, `relay`, `module-resolver`, and Flow/TS stripping all run twice — once inside the StyleX pass, and again in the consumer's main pipeline.

Symptoms range from wasted CPU (~15-30% of a Vite dev warm build in our case) to real bugs (React Compiler running twice, `module-resolver` producing subtly different rewrites depending on whose config runs first).

The correct escape hatch in Babel is `configFile: false`, but there's currently no way to set it because the plugin doesn't expose `transformAsync`'s options.

## Change

Adds a `babel` option to the plugin's user options. Its contents are spread into `transformAsync` **before** the plugin's own mandatory options:

```js
// packages/@stylexjs/unplugin/src/core.js
const result = await transformAsync(inputCode, {
  ...babelExtraOptions,        // ← new: consumer options first
  babelrc: false,              // ← mandatory, always wins
  filename,
  presets,
  plugins: [...],
  caller: {...},
});
```

Because consumer options come first and the plugin's mandatory options are set after, consumers can extend Babel behaviour but **cannot** override the options the plugin depends on (`babelrc`, `filename`, `presets`, `plugins`, `caller`). The added test asserts this invariant.

## Primary use case

```js
stylex.vite({
  // ...
  babel: { configFile: false },
})
```

This lets the plugin's internal Babel pass ignore your project's root config while your main Vite/Rollup/webpack Babel pass continues using it. Any other Babel option (custom `sourceMaps`, `assumptions`, custom `parserOpts`, etc.) also works through the same passthrough.

## Backwards compatibility

Fully non-breaking:
- The option defaults to `{}`, so existing users see no change.
- The mandatory options set after the spread guarantee that even if a consumer passes `babel: { babelrc: true, filename: 'wrong.js' }`, the plugin still runs correctly (there's a test for this).

## Tests

Two new tests in `__tests__/unplugin.test.js`:

- `forwards \`babel\` option to the internal transformAsync call` — passes `babel: { configFile: '/nonexistent' }` and asserts Babel throws with the expected path, proving the option actually reaches `transformAsync`.
- `consumer-supplied \`babel\` options cannot override plugin internals` — passes `babel: { babelrc: true, filename: '/should/be/ignored.js' }` and asserts the transform still works, proving the plugin's mandatory options still win.

All 11 tests pass: `npx jest` in `packages/@stylexjs/unplugin`.

## Docs

README updated with a new bullet under **Options (shared)** documenting `babel`, the primary `configFile: false` use case, and the "consumer options cannot override internals" contract.

---

We've been running this patched against `@stylexjs/unplugin@0.19.0` at Mixcloud for a few weeks (via `pnpm patch`) and it cleanly resolves both the perf issue and the double-compile problem. Happy to iterate on option naming, docs wording, or split the tests differently if the maintainers prefer.

### DIFF
--- a/packages/@stylexjs/unplugin/README.md
+++ b/packages/@stylexjs/unplugin/README.md
@@ -140,6 +140,13 @@ esbuild.build({
   `['stylex', '@stylexjs/stylex']`.
 - `useCSSLayers`: boolean, emit CSS layers.
 - `babelConfig`: `{ plugins, presets }` to merge into the internal Babel call.
+- `babel`: pass-through options for the internal `@babel/core` call. Useful
+  for opting out of root-config discovery (`babel: { configFile: false }`),
+  which otherwise applies your project's `babel.config.*` — including
+  plugins like `react-compiler`, `relay`, or `module-resolver` — to every
+  StyleX file the plugin transforms. The plugin's own mandatory options
+  (`babelrc: false`, `filename`, `presets`, `plugins`, `caller`) always
+  win and cannot be overridden.
 - `unstable_moduleResolution`: forwarded to the StyleX Babel plugin.
 - `lightningcssOptions`: pass-through options for `lightningcss`. Can override
   `targets`, `exclude`, etc. See [Browserslist & CSS lowering](#browserslist--css-lowering).
--- a/packages/@stylexjs/unplugin/__tests__/unplugin.test.js
+++ b/packages/@stylexjs/unplugin/__tests__/unplugin.test.js
@@ -186,6 +186,45 @@ describe('@stylexjs/unplugin', () => {
     }
   });
 
+  test('forwards `babel` option to the internal transformAsync call', async () => {
+    // Babel will only throw about a bad `configFile` if the option reached transformAsync.
+    const plugin = unplugin.raw({
+      dev: false,
+      devPersistToDisk: false,
+      babel: { configFile: '/definitely/does/not/exist/babel.config.js' },
+    });
+    if (typeof plugin.buildStart === 'function') {
+      plugin.buildStart();
+    }
+    const source = `
+      import * as stylex from '@stylexjs/stylex';
+      const styles = stylex.create({ foo: { color: 'red' } });
+      export default styles;
+    `;
+    await expect(
+      plugin.transform(source, '/virtual/example.js'),
+    ).rejects.toThrow(/definitely\/does\/not\/exist\/babel\.config\.js/);
+  });
+
+  test('consumer-supplied `babel` options cannot override plugin internals', async () => {
+    // Pass-through must not be able to override `babelrc: false` / `filename`.
+    const plugin = unplugin.raw({
+      dev: false,
+      devPersistToDisk: false,
+      babel: { babelrc: true, filename: '/should/be/ignored.js' },
+    });
+    if (typeof plugin.buildStart === 'function') {
+      plugin.buildStart();
+    }
+    const source = `
+      import * as stylex from '@stylexjs/stylex';
+      const styles = stylex.create({ foo: { color: 'red' } });
+      export default styles;
+    `;
+    const result = await plugin.transform(source, '/virtual/example.js');
+    expect(result).not.toBeNull();
+  });
+
   test('marks StyleX deps as non-optimized in Vite', async () => {
     const tempDir = fs.mkdtempSync(
       path.join(os.tmpdir(), 'stylex-unplugin-vite-'),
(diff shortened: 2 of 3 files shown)


matclayton:
@mellyeliu @nmn is there anything else you would need from us to upstream this? Sadly @jonreading81 is moving on from us shortly, so I’m trying to close off any loose ends and upstream the patches were running whilst we have the in-house capacity to help support them

PR #1801, opened 2026-08-09
fix: correct media query breakpoint gap on strict range syntax (#1793)

### LINKED ISSUE #1793: Transformed media queries leave gaps between breakpoints
### Describe the issue

Take this create call:
```js
  foo: {
    gridColumn: {
      default: '1 / 2',
      '@media (width >= 48rem)': '1 / 3',
      '@media (width >= 64rem)': '1 / -1',
    },
  },
});
```

It is now transformed at compile time to:
```js
  foo: {
    gridColumn: {
      default: '1 / 2',
      '@media (width >= 48rem) and (width <= 63.99rem) ': '1 / 3',
      '@media (width >= 64rem)': '1 / -1',
    },
  },
});
```

### Expected behavior
It should be transformed to:
```js
  foo: {
    gridColumn: {
      default: '1 / 2',
      '@media (width >= 48rem) and (width < 64rem) ': '1 / 3',
      '@media (width >= 64rem)': '1 / -1',
    },
  },
});
```

### Steps to reproduce

```js
  foo: {
    gridColumn: {
      default: '1 / 2',
      '@media (width >= 48rem)': '1 / 3',
      '@media (width >= 64rem)': '1 / -1',
    },
  },
});
```

### Test case

Input:
```js
  foo: {
    gridColumn: {
      default: '1 / 2',
      '@media (width >= 48rem)': '1 / 3',
      '@media (width >= 64rem)': '1 / -1',
    },
  },
});
```

Output:
```js
  foo: {
    gridColumn: {
      default: '1 / 2',
      '@media (width >= 48rem) and (width < 64rem) ': '1 / 3',
      '@media (width >= 64rem)': '1 / -1',
    },
  },
});
```

### Additional comments

_No response_

sahariya-divyansh:
When StyleX compiles overlapping media query breakpoints using strict range syntax (e.g. `@media (width >= 48rem)` and `@media (width >= 64rem)`), it previously transformed the lower breakpoint into an epsilon-adjusted upper bound like `(width <= 63.99rem)`. This left a small gap between 63.99rem and 64rem where neither breakpoint applied.

This PR changes the merge logic to track exact interval bounds with inclusivity, so it emits a strict comparison like `(width < 64rem)` instead of a rounded `<=` value — eliminating the gap entirely. Legacy `min-width` / `max-width` syntax keeps its existing epsilon fallback for backward compatibility.

## Linked PR/Issues
Fixes #1793

## Additional Context
- Modified `packages/style-value-parser/src/at-queries/media-query.js` to track bounds with inclusivity and emit strict `<`/`>` operators for range syntax.
- Added a regression test in `media-query-transform-test.js` matching the exact example from the issue.
- Updated snapshot tests in `parse-media-query-test.js` to reflect the new AST (operator + exact value instead of epsilon-adjusted value).

Before:
`@media (width >= 48rem) and (width <= 63.99rem)`

After:
`@media (width >= 48rem) and (width < 64rem)`

## Pre-flight checklist

### DIFF
--- a/packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
+++ b/packages/style-value-parser/src/at-queries/__tests__/media-query-transform-test.js
@@ -696,4 +696,29 @@ describe('Media Query Transformer', () => {
     const result = lastMediaQueryWinsTransform(originalStyles);
     expect(JSON.stringify(result)).toBe(JSON.stringify(expectedStyles));
   });
+
+  test('issue 1793: nested range media queries use strict comparison and leave no gaps', () => {
+    const originalStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (width >= 48rem)': '1 / 3',
+          '@media (width >= 64rem)': '1 / -1',
+        },
+      },
+    };
+
+    const expectedStyles = {
+      foo: {
+        gridColumn: {
+          default: '1 / 2',
+          '@media (width >= 48rem) and (width < 64rem)': '1 / 3',
+          '@media (width >= 64rem)': '1 / -1',
+        },
+      },
+    };
+
+    const result = lastMediaQueryWinsTransform(originalStyles);
+    expect(JSON.stringify(result)).toBe(JSON.stringify(expectedStyles));
+  });
 });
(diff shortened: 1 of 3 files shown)


nmn:
We intentionally did not use `<` in our media query syntax to keep the generated CSS maximally compatible with older browsers. There are no devices that have 0.01 device pixels so there is no real bug being fixed by this PR.

PR #1802, opened 2026-08-09
fix: preserve cascading order for @container queries (#1798)

### LINKED ISSUE #1798: Container query order is not preserved.
### Describe the issue

Container query order is not preserved.

### Expected behavior

enableMediaQueryOrder but for container query

### Steps to reproduce

input:
```js
  foo: {
    color: {
      default: null,
      '@container (width >= 360px)': 'green',
      '@container (width >= 420px)': 'red',
    },
  }
})
```

output:
```css
@container (width >= 420px) {
  .x {
    color: green;
  }
}

@container (width >= 360px) {
  .y {
    color: red;
  }
}
```

### Test case

_No response_

### Additional comments

_No response_

hum4nBeing:
Fixes #1798

### Description
Currently, `@container` queries do not preserve their cascading order because the `lastMediaQueryWinsTransform` logic in `@stylexjs/shared` strictly looks for `@media` prefixes, completely bypassing container queries. 

This PR fixes the issue by:
1. **Expanding the Parser (`style-value-parser`)**: Updated `atRuleNameParser` in `media-query.js` to accept the `@container` keyword alongside `@media`. (The existing range-syntax machinery already natively handles `>=` and `<=` operators).
2. **Fixing Serialization**: Updated the AST formatter to preserve the original at-rule name so it serializes back to `@container` instead of rewriting to `@media`.
3. **Refactoring the Transformer (`shared`)**: Extracted the sorting, negation, and range-slicing logic into a reusable helper function (`applyQueryOrder`) inside `media-query-transform.js`, and applied it sequentially to both `@media` and `@container` query families. 

By routing container queries through this deterministic pipeline, they are correctly converted into mutually exclusive ranges. This guarantees that container rules never collide, making physical CSS injection order irrelevant—which aligns perfectly with StyleX's atomic CSS architecture.

### Test Plan
- Added new regression tests to `parse-media-query-test.js` verifying that `@container (width >= 360px)` successfully parses into a valid AST shape.
- Added regression test cases to `media-query-transform-test.js` verifying that multiple container queries are correctly sorted and transformed into mutually exclusive ranges.
- Ran test suites to confirm all existing and new tests pass successfully.


nmn:
Please rebase to fix the conflicts.

PR #1809, opened 2026-08-14
Allow stylex.env keys in the valid styles lint rule

### LINKED ISSUE #1764: `@stylexjs/valid-styles` lint rule prevents using `stylex.env` values as media-queries
### Describe the issue

`stylex.env` currently works well as a way to define shared media-queries; its values are interpolated early in the transform, and they behave just like literal strings. `enableMediaQueryOrder` works correctly (which is not currently the case for media-queries from `defineConsts`), and the emitted CSS works as intended.

However the lint-rule `valid-styles` from `@stylexjs/eslint-plugin` flags this as incorrect; it enforces that only literal strings can appear in object-key position.

### Expected behavior

The lint rule should not flag values from `stylex.env` as incorrect when used in object-key position; it should use its existing knowledge of `styleXDefaultImports` to recognise that a member-expression coming from exactly `stylex.env` is equivalent to a compile-time literal.

### Steps to reproduce

- In a fresh project, set up StyleX with an `env` object containing `responsive: { belowSmall: '@media (max-width: 600px)' }`.
- Enable the `valid-styles` rule from `@stylexjs/eslint-plugin`.
- Use `stylex.env.responsive.belowSmall` as a media-query key.
- Observe that the lint rule flags this as an error with the message `All keys in a stylex object must be static literal values.`

Image

```tsx
  hideBelowSmall: {
    display: {
      [stylex.env.responsive.belowSmall]: 'none',
    },
  },
});
```

### Test case

_No response_

### Additional comments

At Canva we're currently maintaining a short patch against `@stylexjs/eslint-plugin` to allow `stylex.env.*` member-expressions in this lint rule. We would be happy to open a PR and upstream these changes if there's appetite.

Hashim1999164:
Fixes #1764

valid styles currently reports stylex.env member expressions used as object keys. The compiler already treats those as compile time literals, the same way it treats string media queries.

This change recognizes stylex.env chains and named env imports so the rule no longer flags them. Nested CSS values are still checked as usual.

### DIFF
--- a/packages/@stylexjs/eslint-plugin/src/rules/isStylexEnvMember.js
+++ b/packages/@stylexjs/eslint-plugin/src/rules/isStylexEnvMember.js
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+'use strict';
+
+import type { Expression, Pattern } from 'estree';
+
+type ASTNode = {
+  +type: string,
+  +name?: string,
+  +computed?: boolean,
+  +object?: ASTNode,
+  +property?: ASTNode,
+  ...
+};
+
+/**
+ * True when `node` is a compile time `stylex.env` member chain, e.g.
+ * `stylex.env.responsive.belowSmall` or a named import `env.responsive.belowSmall`.
+ * Those values are interpolated as literals by the compiler, so they are valid
+ * object keys (typically media queries).
+ */
+export default function isStylexEnvMember(
+  node: Expression | Pattern,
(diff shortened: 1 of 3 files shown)


nmn (approved):

packages/@stylexjs/eslint-plugin/src/rules/isStylexEnvMember.js
```diff
@@ -0,0 +1,67 @@
...
+ * object keys (typically media queries).
+ */
+export default function isStylexEnvMember(
+  node: Expression | Pattern,
+  styleXDefaultImports: Set<string>,
+  styleXEnvImports: Set<string>,
+): boolean {
+  if (node == null || node.type !== 'MemberExpression') {
+    return false;
+  }
```
nmn, line 36: This is not fully correct. You could define a local constant which holds a value from `stylex.env` and use that in your styles.

PR #1817, opened 2026-08-24
fix(unplugin): match hashed index and style CSS asset names

### LINKED ISSUE #1815: pickCssAssetFromRollupBundle default regex doesn't match Vite hashed CSS filenames
## Summary

The default CSS asset picker in `@stylexjs/unplugin` uses `/(^|\/)index\.css$/` to find the target asset for StyleX CSS injection. This regex does not match Vite's hashed output filenames (e.g. `index-B5Jdbbfd.css`), causing the picker to fall back to `assets[0]` — which is nondeterministic when multiple CSS chunks exist.

## Impact

When a Vite app has more than one CSS chunk (e.g. from lazy-loaded routes), `assets[0]` may be a route-specific chunk rather than the global stylesheet. All StyleX CSS gets appended to that chunk, which only loads when that route is visited. Every other route loses its styles.

This is silent — no warning, no error. The build succeeds. Styles just don't apply on most routes.

## Reproduction

1. Vite + React app with `@stylexjs/unplugin@0.19.0`
2. Lazy-load a route that imports a CSS file (e.g. `import './page.css'`)
3. Build — Vite emits `index-[hash].css` (global) and `Page-[hash].css` (route chunk)
4. The picker regex `/(^|\/)index\.css$/` fails to match `index-[hash].css`
5. Falls back to `assets[0]` — which may be `Page-[hash].css`
6. All StyleX CSS is appended to `Page-[hash].css`
7. Other routes have no StyleX styles

## Expected behavior

The default picker should match hashed variants of `index.css` (e.g. `index-[hash].css`), or the regex should be broadened to `/index.*\.css$/`.

## Current workaround

Set `cssInjectionTarget` explicitly:

```ts
stylex.vite({
  cssInjectionTarget: (fileName: string) =>
    fileName.startsWith('assets/index-') && fileName.endsWith('.css'),
})
```

## Environment

- `@stylexjs/unplugin@0.19.0` (latest)
- Vite 8.x with Rolldown

Om-singhaI:
# fix(unplugin): match hashed index and style CSS asset names

`pickCssAssetFromRollupBundle` in `packages/@stylexjs/unplugin/src/core.js`
chooses which stylesheet the collected StyleX rules get appended to. Its
documented contract is to prefer a stylesheet named like `index.css` or
`style.css`, and to fall back to the first CSS asset only when neither exists.

The two patterns it used required the file name to end in exactly `index.css`
or `style.css`:

```js
const best =
  assets.find((a) => /(^|\/)index\.css$/.test(a.fileName)) ||
  assets.find((a) => /(^|\/)style\.css$/.test(a.fileName));
return best || assets[0];
```

Vite's default `assetFileNames` is `assets/[name]-[hash][extname]`, so the entry
stylesheet reaches the bundle as `assets/index-B5Jdbbfd.css`. Neither pattern
can match a hashed name, so `best` is always `undefined` in a production Vite or
Rollup build and the picker silently degrades to `assets[0]`, which is whichever
stylesheet the bundler happened to emit first. The documented preference never
runs at all.

When that first stylesheet is not the entry stylesheet, the StyleX rules are
appended to a file the entry page never loads, and the styles simply do not
apply in the production build.

The same two expressions were duplicated across every bundler backend, so the
same miss occurred in the `writeBundle` output directory fallback in `vite.js`
and `rollup.js`, in the `processAssets` hook in `webpack.js`, and in both
picking sites in `esbuild.js`. Twelve regex literals in five files, all with the
same defect.

Fixes #1815

### Reproduction

A two entry Vite 8.2.2 build, default config, no `cssInjectionTarget`. The CSS
assets arrive in the bundle in this order:

```
CSS_ASSETS_IN_ORDER=["assets/admin-CyH6f07N.css","assets/index-W1erjkBN.css"]
```

`assets[0]` is the admin stylesheet, so that is where the StyleX CSS landed.
Building the same project with the plugin compiled from `main` and then from
this branch, and grepping the output for a marker rule:

```
before:  dist/assets/admin-2aotGWHc.css
after:   dist/assets/index-joS6COZB.css
```

Worth stating precisely, since the issue describes a lazy route chunk: on Vite
8.2.2 a single entry app with a dynamically imported route emitted its CSS in
the order `["assets/index-W1erjkBN.css","assets/lazy-uwHDjYoA.css"]`, so
`assets[0]` was the entry stylesheet by luck and the bug stayed hidden. The
multi entry build above is the case I could reproduce end to end. Either way the
underlying defect is the same: the preference is dead code under default asset
naming, and the result depends on bundler emission order rather than on the
documented rule.

### The fix

Both patterns now allow an optional content hash, and are defined once and
shared by all five backends:

```js
export const INDEX_CSS_RE = /(^|\/)index(-[\w-]{8,})?\.css$/i;
export const STYLE_CSS_RE = /(^|\/)style(-[\w-]{8,})?\.css$/i;
```

The character class has to be the full base64url alphabet, not just letters and
digits. Rollup's default `hashCharacters` is `base64`, and the hash is produced
by `xxhashBase64Url`, so roughly one eight character hash in five contains a `-`
or a `_`. Sampling 20000 hashes straight from Rollup 4.59's own hasher gives
4082 such hashes, and a narrower `[a-z0-9]` class misses every one of them:

```
hashes containing - or _:  4082 / 20000  (20.41%)
not matched by [a-z0-9]:   4082
e.g. assets/index-BVm_Qe95.css, assets/index-CnU-v52M.css
```

That is not a hypothetical. Varying the entry CSS in the two entry app above
until its hash lands on such a value gives a real Vite 8.2.2 build where the
narrower pattern reproduces the original bug exactly:

```
CSS_ASSETS_IN_ORDER=["assets/admin-CyH6f07N.css","assets/index-BVm_Qe95.css"]
[a-z0-9] class picks  = undefined      -> falls back to assets[0], the admin sheet
base64url class picks = assets/index-BVm_Qe95.css
```

Building that app with the plugin compiled each way, and grepping the output for
the marker rule, shows the difference is not cosmetic. A hash class of letters
and digits alone leaves the entry stylesheet empty exactly as `main` does:

```
main                    -> dist/assets/admin-2aotGWHc.css
[a-z0-9] hash class     -> dist/assets/admin-2aotGWHc.css   (still wrong)
base64url hash class    -> dist/assets/index-DOdCseHv.css
```

`getAssetBaseName`, a few lines below in the same file, uses the narrower
`(-[a-z0-9]{8,})?` shape. It is left as it is on purpose: it strips a trailing
hash off a name with a lazy `(.*?)` prefix, and widening its class there would
make it over strip, turning `reset-index-BpQ2m1Zx.css` into `reset.css` instead
of `reset-index.css`. Matching a whole name and stripping a suffix want
different classes, so they are deliberately not shared.

The `/i` flag covers Vite's mixed case hashes such as `B5Jdbbfd`.

The leading `(^|\/)` anchor is kept deliberately. Without it a name such as
`assets/reset-index-BpQ2m1Zx.css` would be treated as the entry stylesheet.
There is a test covering that.

Unhashed `index.css` and `style.css` still match through the optional group, and
an explicit `cssInjectionTarget` still takes priority over both, so existing
behaviour is unchanged.

### Tests

Eight tests were added under a new `CSS asset selection` block in
`packages/@stylexjs/unplugin/__tests__/unplugin.test.js`. Every hashed file name
used in them was taken from real `vite build` output rather than invented.

Five of them fail on `main` and pass on this branch:

```
● CSS asset selection › prefers the hashed entry stylesheet over other stylesheets
  Expected: "assets/index-W1erjkBN.css"
  Received: "assets/admin-CyH6f07N.css"

● CSS asset selection › prefers an entry stylesheet whose hash contains an underscore
  Expected: "assets/index-BVm_Qe95.css"
  Received: "assets/admin-CyH6f07N.css"

● CSS asset selection › prefers an entry stylesheet whose hash contains a hyphen
  Expected: "assets/index-CnU-v52M.css"
  Received: "assets/admin-CyH6f07N.css"

● CSS asset selection › prefers a hashed style.css over other stylesheets
  Expected: "assets/style-BpQ2m1Zx.css"
  Received: "assets/admin-CyH6f07N.css"

● CSS asset selection › writeBundle fallback appends to the hashed entry stylesheet
  Expected substring: "987654"
  Received string:    "body{color:red}"

Tests: 5 failed, 9 skipped, 3 passed, 17 total
```

That failure output was captured by reverting only `src/` and keeping the new
tests in place. The two underscore and hyphen cases also fail against a version
of this fix that uses a plain `[a-z0-9]` hash class, which is what pins the
character class down rather than leaving it to taste.

The other three tests guard behaviour that was already correct and must stay
correct: unhashed `index.css` still wins, a longer name ending in `index` is not
mistaken for the entry, and `cssInjectionTarget` still overrides both patterns.

With the fix applied, the full package suite passes:

```
Tests:       17 passed, 17 total
```

`prettier --check` and `eslint` are both clean on the changed files. The changed
files carry no `@flow` pragma, so Flow does not type check them.

### Scope note

`webpack.js` and `esbuild.js` were updated along with the Vite and Rollup paths.
Hashed CSS names are opt in rather than default for those two bundlers, so they
are not what issue #1815 reports, but they carried literal copies of the same
two expressions and would miss a hashed name in exactly the same way. Leaving
half the backends on the old patterns while sharing constants in the other half
seemed worse than making the behaviour uniform. Happy to split that into a
separate change if preferred.

### Known limits

Two things this deliberately does not cover, flagged so they are a choice rather
than an oversight.

Webpack users commonly configure `[name].[contenthash].css`, with a dot rather
than a hyphen before the hash. That still will not match. Accepting `[-.]` as
the separator would cover it, at the cost of claiming names like
`index.tailwind.css` as the entry stylesheet, so it is left out until someone
actually reports it.

Widening the class to base64url does let a hand written name such as
`assets/index-dark-theme.css` match. That is a heuristic replacing another
heuristic: the alternative for such a bundle is `assets[0]`, whichever
stylesheet the bundler emitted first, and a file literally named `index-...` is
the better of the two guesses. `cssInjectionTarget` remains the escape hatch for
anyone who wants to name the target outright.

## Preflight checklist

* [x] I have read the contributing guidelines
* [x] Performed a self review of my code


nmn (approved):

nmn:
Thank you!

PR #1820, opened 2026-08-25
Accept Marker type in `StyleXStyles` for passing between component prop boundaries

titchimoto:

In our Design System, we follow [StyleX's docs recommend pattern](https://stylexjs.com/docs/api/types/StyleXStyles) for constraining component style props using `StyleXStyles<T>`:

```ts
// Design system component
type BoxProps = {
  /** Only layout properties allowed, other styles are intentionally blocked */
  xstyle?: StyleXStyles<{
    margin?: CSSProperties['margin'];
    padding?: CSSProperties['padding'];
    flexGrow?: CSSProperties['flexGrow'];
  }>;
};
```

This pattern works for style overrides but consumers also need to use the marker API to implement ancestor-state patterns, for example, revealing a child element when a parent is hovered, entirely in CSS:

```tsx
  actions: {
    opacity: {
      default: 0,
      [stylex.when.ancestor(':hover')]: 1,
    },
  },
});

// Mark the parent component as the hover anchor...
<Box xstyle={stylex.defaultMarker()}>  {/* ❌ Type error */}
  <ChildComponent xstyle={styles.actions} />
</Box>

// ...or compose layout styles and a marker together
<Box xstyle={[styles.layout, stylex.defaultMarker()]}>  {/* ❌ Type error */}
  <ChildComponent xstyle={styles.actions} />
</Box>
```

Because `StyleXStyles<T>` has no way to express "a marker token is also valid here", in our DS, we've created an internal `XStyle` type which is essentially a union of `StylexStyles<T> | StyleXMarkerToken` (which we've just created internally from the `ReturnType`'s as in this PR), which does _work_ & without that our DS consumers are forced to either add additional wrapping elements, or are generally unable to use our components for their use-case.

_But_ it would be nice to not have to manage this at the DS level & could live in the library instead to pass between component prop boundaries easier.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
+++ b/packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
@@ -230,6 +230,7 @@ export type StyleXStyles<
   | false
   | GenStylePropType<CSS>
   | Readonly<[GenStylePropType<CSS>, InlineStyles]>
+  | StyleXMarkerToken
 >;
 export type StyleXStylesWithout<CSS extends UserAuthoredStyles> = StyleXStyles<
   Omit<CSSPropertiesWithExtras, keyof CSS>
@@ -317,6 +318,14 @@ export type StyleX$CreateTheme = <
 export type StyleX$DefineMarker = () => MapNamespace<{
   readonly marker: symbol;
 }>;
+
+export type StyleX$DefaultMarker = () => MapNamespace<{
+  readonly marker: 'default-marker';
+}>;
+
+export type StyleXMarkerToken =
+  | ReturnType<StyleX$DefineMarker>
+  | ReturnType<StyleX$DefaultMarker>;
 
 export type StyleX$When = {
   ancestor: <
--- a/packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
+++ b/packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
@@ -230,7 +230,7 @@ export type StyleXStyles<
   | false
   | GenStylePropType<CSS>
   | Readonly<[GenStylePropType<CSS>, InlineStyles]>
-  | StyleXMarkerToken
+  | StyleXMarker
 >;
 export type StyleXStylesWithout<CSS extends UserAuthoredStyles> = StyleXStyles<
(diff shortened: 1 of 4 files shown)


packages/@stylexjs/stylex/src/types/StyleXTypes.d.ts
```diff
@@ -318,6 +319,14 @@ export type StyleX$DefineMarker = () => MapNamespace<{
...
 }>;
 
+export type StyleX$DefaultMarker = () => MapNamespace<{
+  readonly marker: 'default-marker';
+}>;
+
+export type StyleXMarkerToken =
+  | ReturnType<StyleX$DefineMarker>
+  | ReturnType<StyleX$DefaultMarker>;
+
```
nmn:
```suggestion
export type StyleXMarker = MapNamespace<{
  readonly marker: symbol;
}>;

export type StyleX$DefineMarker = () => StyleXMarker;

export type StyleX$DefaultMarker = () => StyleXMarker;

```

packages/@stylexjs/stylex/src/types/StyleXTypes.js
```diff
@@ -188,7 +188,8 @@ export type StyleXStyles<
 > = StyleXArray<
   | ?false
   | GenStylePropType<Readonly<CSS>>
-  | Readonly<[GenStylePropType<Readonly<CSS>>, InlineStyles]>,
+  | Readonly<[GenStylePropType<Readonly<CSS>>, InlineStyles]>
+  | StyleXMarkerToken,
```
nmn:
```suggestion
  | StyleXMarker,
```

nmn (changes requested):
Default markers and custom markers should really be combined into a single type.

titchimoto:
> Default markers and custom markers should really be combined into a single type.

@nmn sounds good! updated [here](https://github.com/facebook/stylex/pull/1820/commits/3673e8203469c20638b9e80d1b68d68cc56dd343) to collapse both these marker types down into a single type. I did keep the Flow `custom-marker` type in both cases now if that works? but I think that should keep these both closer in sync now.

nmn (approved):

PR #1826, opened 2026-08-28
docs: fix last-wins wording after property-specificity default

mccafferson:

Since [0.14.0](https://stylexjs.com/blog/v0.14.0/), the default `styleResolution` is `property-specificity`: a longhand such as `margin-top` wins over `margin` regardless of `props()` order. “Last style applied always wins” is `application-order`.

The default-switch commit ([58a569eaac](https://github.com/facebook/stylex/commit/58a569eaacfd380668bc8bb120309fca79f4c01e)) updated the babel-plugin config page, but the learn pages still used the old wording — including the `margin` vs `margin-top` example, which is the case where last-applied does *not* win under the default.

This PR:

- Corrects [Thinking in StyleX](https://stylexjs.com/docs/learn/thinking-in-stylex/)
- Adds the same-property vs shorthand/longhand distinction on [Using styles](https://stylexjs.com/docs/learn/styling-ui/using-styles/)
- Softens the intro feature bullet so it doesn’t overclaim last-wins

Historical blog posts are left as they were when `application-order` was the default.

## Linked PR/Issues

Related to #1064 (default switch).

## Additional Context

Verified against `@stylexjs/babel-plugin` 0.19.0: compiling `stylex.props({marginTop: 0, …}, {margin: 8})` keeps all four longhand classes plus the shorthand; the longhands get extra `:not(#\#)` specificity, so computed margin stays `0`.

## Pre-flight checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/docs/content/docs/learn/index.mdx
+++ b/packages/docs/content/docs/learn/index.mdx
@@ -31,7 +31,7 @@ superior to what could be authored and maintained by hand.
     <ul>
       <li>No "style at a distance"</li>
       <li>No specificity issues</li>
-      <li>“The last style applied always wins!”</li>
+      <li>Last applied style wins for the same property</li>
     </ul>
   }
 />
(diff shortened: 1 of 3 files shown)


nmn (approved):

nmn:
We want to change the default back to "application-order" as it tends to be closer to what people actually want, but there is a small performance cost that comes with that. We're looking into finding a solution to that problem.

PR #1844, opened 2026-09-03
[babel-plugin] use globalThis in the compile-time evaluator

### LINKED ISSUE #1838: Playground is broken
### Describe the issue

Visiting the playground results in an error

/App.tsx: global is not defined

Image

### Expected behavior

The playground is functional

### Steps to reproduce

Visit https://stylexjs.com/playground

### Test case

_No response_

### Additional comments

It looks like it is the usage of the `stylex.when` API that causes a problem. If the usage of that API is deleted from the example, the playground returns to a usable state.

Image

henryqdineen:

**This only affects the playground.** It's a bug in the browser build of the plugin (`lib/index.browser.js`), which the playground is the only known consumer of. Node builds were never affected, since `global` exists there.

The compile-time evaluator read globals off Node's `global` object in three places in `src/utils/evaluate-path.js`. `global` doesn't exist in browsers, so those reads throw `ReferenceError: global is not defined` when the plugin runs client-side, as it does in the docs playground. Babel prefixes the filename, giving the reported `/App.tsx: global is not defined`.

Swapped all three to `globalThis`, which is equivalent in Node and available in browsers.

It broke now because of the read added in #1818: `isBlockedFunction` compares against `global.eval` and runs for every call expression, which is why removing the `stylex.when(...)` call from the example works around it. The other two reads are older and equally broken in a browser, but sit behind the `isValidCallee` allowlist (`String`, `Number`, `Math`, `Object`, `Array`) that no playground example hits.

## Linked PR/Issues

Fixes #1838

## Additional Context

`globalThis` behaves identically to `global` in Node, so this is strictly a widening. No behavior change for any Node consumer of the plugin.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/src/utils/evaluate-path.js
+++ b/packages/@stylexjs/babel-plugin/src/utils/evaluate-path.js
@@ -159,7 +159,7 @@ function isBlockedFunction(fn: mixed): boolean {
     Object.getPrototypeOf(callable) === Function ||
     // Referenced, never called: this is a value the evaluator refuses to run.
     // eslint-disable-next-line no-eval
-    callable === globalThis.eval
+    callable === (globalThis as $FlowFixMe).eval
   );
 }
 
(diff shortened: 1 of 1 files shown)


mellyeliu:
can we fix the flow types?

henryqdineen:
> can we fix the flow types?

Thanks. I always forget about Flow. Should be fixed in [cc26f99](https://github.com/facebook/stylex/pull/1844/commits/cc26f99f2ae7c0440aa61fd9e2243ee1554d8bd6). I considered just having a transform from `global` -> `globalThis` as part of the rollup of the browser bundle but it seems like `globalThis` is generally preferred for NodeJS code.

henryqdineen:
This was fixed by #1847

PR #1850, opened 2026-09-06
[babel-plugin] Fix positionTry emitting every declaration twice

lenamonj:

`stylex.positionTry` emitted every declaration twice, once with the property name as its own value, and produced an RTL variant for every rule. `generateLtr` and `generateRtl` take and return a `[key, value]` pair, but the code mapped only the value, so the returned pair was spliced in as the value: `anchor-name:anchor-name;anchor-name:--myAnchor`. The generators are now applied to the whole entry with `objMapEntry`, the same way the other transforms use them.

## Linked PR/Issues

None.

## Additional Context

The existing snapshots recorded the doubled output; they are updated to the corrected rules and the RTL variant is `null` where nothing flips.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/babel-plugin/__tests__/transform-import-export-test.js
+++ b/packages/@stylexjs/babel-plugin/__tests__/transform-import-export-test.js
@@ -239,14 +239,14 @@ describe('@stylexjs/babel-plugin', () => {
           __varGroupHash__: "xop34xu"
         };
         const viewTransition1 = "xchu1hv";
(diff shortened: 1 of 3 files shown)


lenamonj:
The failing Vercel check is the deploy authorization, which needs someone on the Meta team; the CLA check passes. Happy to adjust anything on review.

nmn (approved):

nmn:
Thank you!

PR #1865, opened 2026-09-10
[RFC] Add compile-time CSS enums and exhaustive matching

nmn:
Adds enums: CSS variables with a fixed set of possible states that can be used as conditions. Read the current state with `stylex.match`, or call the enum to override it on an element.

The most typical example would be to define a custom implementation of `color-scheme` and `light-dark`.

```tsx
// enums.stylex.ts

export const colorScheme = stylex.defineEnum(['light', 'dark'], {
  default: 'light',
  '@media (prefers-color-scheme: dark)': 'dark'
});
```

Then, this value can be used for defining styles as a simple alternative to `light-dark()`, except that this will work with any value, and not just colors.

```tsx
import { colorScheme } from './enums.stylex';

  card: {
    color: stylex.match(colorScheme, {light: 'black', dark: 'white'}),
    boxShadow: stylex.match(colorScheme, {
      light: '0 2px 4px rgb(0 0 0 / 0.25)',
      dark: '0 1px 2px rgb(255 255 255 / 0.1) inset'
    }),
  },
});
```

Enums are not constants. So, the "colorScheme" enum's value can be overridden easily:

```tsx
<div {...stylex.props(colorScheme('dark'))}>{children}</div>
```

---

Of course, enums can be used for other use-cases like density and can have more than 2 cases:

Conceptually, `density` behaves like a typed custom property whose value is `'compact' | 'comfortable'`. It inherits through the element tree, and the browser uses that state to choose each matching CSS value.

```tsx
// density.stylex.ts

export const density = stylex.defineEnum(['compact', 'comfortable'], {
  default: 'comfortable',
  '@media (max-width: 600px)': 'compact',
});

export const expanded = stylex.defineEnum([true, false], false);
```

Here, the initial density follows the viewport in CSS. Boolean enums work the same way as string enums.

A component can choose its own values for each state. Setting the enum on a parent also affects descendants that consume it:

```tsx
import { density } from './density.stylex';

  card: {
    padding: stylex.match(density, { compact: 8, comfortable: 16 }),
    borderRadius: stylex.match(density, { compact: 4, comfortable: 8 }),
  },
});

<section {...stylex.props(density('compact'))}>
  <article {...stylex.props(styles.card)}>Compact card</article>
</section>;
```

The card gets `8px` padding and a `4px` radius. Other components can import the same enum without receiving a density prop or sharing a theme variable group. The local assignment takes precedence over the inherited viewport default.

Assignments compose through `stylex.props`. The last assignment replaces the earlier selection as a whole:

```tsx
stylex.props(density('compact'), density('comfortable')); // comfortable

// For a runtime choice, choose between compiled calls.
stylex.props(isCompact ? density('compact') : density('comfortable'));
```

Enum calls require statically evaluable arguments; `density(runtimeValue)` is not supported. TypeScript and Flow check the allowed states across imports and require every state in a match map.

A property can still use pseudo-classes and at-rules around a match:

```tsx
const hoverStyles = stylex.create({
  card: {
    padding: {
      default: stylex.match(density, { compact: 8, comfortable: 16 }),
      ':hover': stylex.match(density, { compact: 12, comfortable: 24 }),
      '@media print': 0,
    },
  },
});
```

Hover changes this card's padding; it does not change the enum inherited by other consumers. This PR supports scalar enum assignments. Conditional assignments and spreads inside `stylex.create` arrive in #1870.

The implementation uses the CSS space hack: one custom property per state, with exactly one active. Here is the underlying idea with readable names:

```css
.compact {
  --density-compact: initial;
  --density-comfortable: ;
}

.card {
  padding: var(--density-compact, 8px) var(--density-comfortable, 16px);
}
```

`initial` makes the active state use its `var()` fallback. The other variable contains whitespace and contributes no value, leaving `8px`. This gives the conditional-value behavior without requiring native CSS `if()` support.

Validated with compiler snapshots, TypeScript/Flow checks, and browser checks for defaults, inheritance, and root overrides with raw and minified CSS.

PR 1 of 6. Next: #1866.


henryqdineen:
This is really cool. 

> This gives the conditional-value behavior without requiring native CSS if() support.

When CSS if() support is widely available does it make this API obsolete or could we use the same API and just use `if()` under the covers. Excuse my ignorance as I'm just learning about the `var: ;` hack. 

> stylex.defineEnum([true, false], false);

Would it be cleaner to have a `defineBool(defaultVal)` or maybe just enforce only using string enums? I could see not wanting to add another `define` export but `stylex.defineEnum([true, false], false)` is a bit odd at first.


Should your PR stack include an update to the `@stylex/enforce-extension` to allow `defineEnum` or is that premature?

nmn:
> When CSS if() support is widely available does it make this API obsolete or could we use the same API and just use if() under the covers.

yes, that’s the idea.

> or maybe just enforce only using string enums?

kinda leaning towards this as the syntax expect you to use the values as objects keys. So strings make the most sense. Even “true” and “false” are treated like strings.

> Should your PR stack include an update to the @stylex/enforce-extension to allow defineEnum

definitely belongs in this stack. Thanks for catching that!

jeffijoe:
So if you have a component API like

```tsx
interface Props {
  spacing?: 'sm' | 'default' | 'lg'
}

export function Card({ spacing = 'default' }: Props) { ... }
```

How would you map that to the proper enum calls in the most ergonomic way?

nmn:
@jeffijoe this RFC doesn’t affect that. “Variants” for component styling can be done with simple “stylex.create” already.

See: https://stylexjs.com/docs/learn/recipes/variants

Issue #1463, opened 2026-01-25
`url` value for `cursor` property


supagetto:
### Describe the issue

`cursor` property does not allow `url(...), ...` value.

<img width="560" height="114" alt="Image" src="https://github.com/user-attachments/assets/b28eeb8b-102c-4ed4-90a3-5f089a813047" />

<img width="916" height="257" alt="Image" src="https://github.com/user-attachments/assets/88767894-774b-40c9-9b93-333a8dc5494e" />

<img width="634" height="110" alt="Image" src="https://github.com/user-attachments/assets/948e8846-041e-41a0-8f2d-4f0ab9df2c7c" />

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/cursor

### Expected behavior

Should allow `url(...), ...` value.

### Steps to reproduce

@stylexjs/stylex@0.17.4

### Additional comments

Solution: Add the type `string` to the type values.

refirst11:
By using `string & {}` instead of the usual `string` for cursor values, you can maintain the existing keyword autocomplete functionality while also allowing the use of arbitrary strings such as `url(...), default`.

I confirmed that it can indeed be received as a property by using the same type `CompiledStyles` as the `stylex.props()` argument.
```ts
import type { CompiledStyles } from "@stylexjs/stylex";

const Screen = ({ style }: { style: CompiledStyles }) => {
  return <div {...stylex.props(style)}>Hello from StyleX!</div>;
};
```

PR #1878, opened 2026-09-11
[types] Validate CSSProperties against the CSS grammar, Baseline, and Flow

henryqdineen:

> **Stacked on #1877.** GitHub will not let a pull request from a fork target a branch on that fork, so this targets `main` and its diff therefore includes #1877's commit. Review the last commit only (`Test CSSProperties against the CSS grammar, Baseline, and Flow`); it will shrink to just that once #1877 merges.

Stacked on #1877, which made `StyleXCSSTypes.d.ts` a hand-maintained duplicate of the Flow types.

`CSSProperties` is 522 properties and roughly 2,500 keywords, all written by hand, and nothing checks it against CSS. At that size review cannot catch a keyword that is misspelled, one that was never in any spec, or one that is missing — and after #1877 there are two copies that can drift apart.

CSS is machine-readable where it matters: the value definition grammar states what each property accepts, and Baseline states what is safe to use. These tests ask those sources rather than asking a reviewer to know CSS, so the types become checkable instead of trusted.

Deliberately validating, not generating. The types stay hand-written and hand-editable; the tests only report where they disagree with CSS. Where they intentionally differ — browsers and specs disagree too — it goes in an allowlist **with a reason**. Today a missing keyword and a deliberate omission look identical; afterwards they do not. That turns out to be 38 decisions and 66 gaps, all currently indistinguishable from each other and from correct code.

### The six tests

1. The TypeScript definitions match the Flow types — nothing else stops them diverging.
2. Every property accepts the CSS-wide keywords (`initial`, `inherit`, `unset`, `revert`, `revert-layer`) — valid everywhere, in no grammar, so nothing else can check them.
3. Every keyword is real, per the grammar — typos and invented values.
4. Closed unions list every keyword the grammar has — the reverse of 3: values a user cannot write at all.
5. Properties whose grammar needs arbitrary strings accept them. Not "are the literals right?" but "should this be a union of literals at all?" Would have caught #1463.
6. Every Baseline-available property is typed.

### Built so an agent can close the loop

Each allowlist entry is work still to do, and each failure names the property, prints its grammar or Baseline status, and states the two valid fixes. Remove an entry, run the test, read the failure, make the change, re-run — no CSS knowledge required beyond the output.

That is deliberate, because what remains is bulk cleanup rather than design. I had a small model do three of them unattended from the test output alone and it got them right. #1881 is five worked examples; what makes them trustworthy is not care in writing them but that the tests check each against the grammar *and* confirm both files still agree.

## Linked PR/Issues

Stacked on #1877. Test 5 surfaces #1463, fixed in #1880.

## Additional Context

The allowlists are two different things, and it is worth separating them. Some entries are decisions that should stay, because specs and browsers disagree and StyleX sides with browsers. The rest are a to-do list.

**Deliberate deviations (38) — expected to stay:**

| | | |
| --- | --- | --- |
| `KEPT_KEYWORDS` | 19 | Keywords the spec dropped that browsers still ship. `marginTrim` is the clearest: the spec was rewritten to `none \| in-flow \| all`, but Safari ships the original `block`/`inline` |
| `NOT_TYPED` | 17 | SVG presentation attributes. StyleX types `fill`, `stroke` and `strokeWidth`, so excluding the rest is arbitrary — but `x`, `y`, `d` and `r` read oddly as style keys |
| `NOT_TYPED` | 1 | `all`, which resets every property at once and cannot work with styleq's key-based override model |
| `NO_GRAMMAR` | 1 | `WebkitBoxOrient`, which css-tree simply lacks but the `-webkit-line-clamp` idiom needs |

**Known gaps (66) — expected to reach zero:**

| | | |
| --- | --- | --- |
| `INCOMPLETE_UNIONS` | 32 | Closed unions missing keywords the grammar allows — `display` cannot take `flow` or `table-caption`, `overflow` cannot take `overlay` |
| `NO_GRAMMAR` | 15 | Properties in no current spec, to be deleted: abandoned drafts (`motion-*` became `offset-*`), `@font-face` descriptors that were never properties (`src`, `unicode-range`) |
| `CLOSED_DESPITE_GRAMMAR` | 13 | Types too narrow for their grammar. `colorScheme` and `fontStyle` accept neither strings nor numbers; the rest accept a bare number, so only `calc()`, `var()` and units are out of reach |
| `NOT_TYPED` | 6 | Baseline-available properties that should be typed — `whiteSpaceCollapse` is literally commented out in `StyleXCSSTypes.js` awaiting this |

`marginTrim` appears in both halves, which is honest: it keeps values the spec dropped *and* is missing values the spec added.

**Scoping test 4** is what makes it useful: restricted to closed unions it finds 125 real omissions across 32 properties, where unrestricted it demands over ten thousand additions, because shorthands like `background` transitively pull in every named colour.

**Two new devDependencies**, both test-only. `css-tree` supplies the grammar and its parser — a patched fork of `mdn-data`, and the same parser `csstype` uses. `web-features` supplies Baseline; `mdn-data`'s `status` field looks like it would do instead, but it describes the spec rather than browsers, calling `font-stretch` obsolete though every browser has supported it since 2020 while marking its replacement `font-width` experimental. It is pinned exactly so that a version bump lands in `package.json` rather than as a lockfile line inside Dependabot's weekly grouped PR — worth knowing, because when it moves, test 6 can legitimately fail: a property reaching Baseline is the news it exists to deliver.

**Why not codegen, or `csstype`.** Generating these types from the grammar was the first thing I tried; it worked, but produced a ~4,000-line diff, replaced a file people read with generated output, and put css-tree in the build path of a public API — with no existing codegen in the repo to follow. `csstype` is spec-current in ways that clash: it would break roughly 24 of the deviations above, and wrapping it for StyleX's semantics (`null` to unset, bare numbers meaning pixels, `theme`, comma-list arrays) is unpleasant in Flow. The dual Flow/TypeScript requirement is what makes most off-the-shelf options awkward.

**The omissions are not deliberate** — I checked whether StyleX prefixes or normalises values in a way that made them redundant, and it does not.

**Follow-ups.** No TSDoc, unlike csstype, though most of the data is available here. And `packages/@stylexjs/eslint-plugin/src/reference/cssProperties.js` is a second, independent encoding of this knowledge that could use the same treatment — worth knowing before leaning on it that its conversion misassigned values, so `stylex/valid-styles` currently accepts `zIndex: 'ideographic'` and `transform: 'border-box'`.

Each failure was verified by injecting it. The suite runs in ~60ms.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/stylex/__tests__/css-types-test.js
+++ b/packages/@stylexjs/stylex/__tests__/css-types-test.js
@@ -1,0 +1,910 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+/**
+ * Keeps `CSSProperties` honest.
+ *
+ * Two things are hand-maintained and can drift:
+ *
+ *   1. `StyleXCSSTypes.d.ts` duplicates `StyleXCSSTypes.js`, because a Flow
+ *      union containing `string` collapses to `string` and loses its literals,
+ *      so TypeScript needs `(string & {})` instead. Nothing stops the two
+ *      files diverging apart from this test.
+ *   2. The property values themselves are written by hand, so a typo or an
+ *      invented keyword goes unnoticed. `css-tree` ships the CSS value
+ *      definition grammar, so it can say whether a keyword is real.
+ */
+
+const fs = require('fs');
+const path = require('path');
+const { parse } = require('@babel/parser');
+const { lexer } = require('css-tree');
+// `web-features` itself is ESM; its data is exported as JSON.
+const { features } = require('web-features/data.json');
+const { generate } = require('css-tree/definition-syntax');
+
+const TYPES_DIR = path.resolve(__dirname, '../src/types');
+const FLOW_FILE = path.join(TYPES_DIR, 'StyleXCSSTypes.js');
+const TS_FILE = path.join(TYPES_DIR, 'StyleXCSSTypes.d.ts');
+
+/** CSS-wide keywords, accepted by every property and never in a grammar. */
+const GLOBAL_KEYWORDS = new Set(['inherit', 'initial', 'unset']);
+
+/**
+ * Properties that do not accept the CSS-wide keywords at all.
+ *
+ * Every property is meant to compose the `all` alias. These six are instead
+ * written `null | 'a' | 'b'`, which drops `inherit`, `initial` and `unset`.
+ * They are also the only properties in the file with no named type alias --
+ * they inline their union at the property, which is what hid the slip.
+ */
+const NO_GLOBAL_KEYWORDS = new Set([
+  'colorScheme',
+  'marginTrim',
+  'paintOrder',
+  'textJustify',
+  'WebkitBackgroundClip',
+  'WebkitBoxOrient',
+]);
+
+/* ------------------------------------------------------------------ *
+ * Reading the type files
+ * ------------------------------------------------------------------ */
+
+/**
+ * What a property accepts, reduced to the three things worth comparing: the
+ * literal keywords, whether a bare number is allowed, and whether an arbitrary
+ * string is allowed.
+ */
+const blank = () => ({ literals: new Set(), open: false, numeric: false });
+
+const absorb = (into, from) => {
+  from.literals.forEach((literal) => into.literals.add(literal));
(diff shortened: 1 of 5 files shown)


henryqdineen:
Superseded by #1883. Reopened from branches in this repository rather than my fork, so the five PRs can be a proper [stack](https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/creating-stacked-pull-requests) — cross-fork stacks aren't supported, and each PR here was showing the earlier ones' commits in its diff. Same commits, unchanged.

PR #1879, opened 2026-09-11
[types] Accept revert and revert-layer, and fix six properties that accept no CSS-wide keywords

henryqdineen:

> **Stacked on #1878.** GitHub will not let a pull request from a fork target a branch on that fork, so this targets `main` and its diff therefore includes #1877 and #1878. Review the last two commits only; the diff will shrink to just those once the earlier PRs merge.

Two bugs, both found by the CSS-wide keyword test added in #1878.

**`revert` and `revert-layer` were missing from every property.** The `all` alias that 516 properties compose stopped at `unset`. `revert` has been Baseline since 2021 and `revert-layer` since 2022; csstype has carried both for years. Fixed in one place, since `all` is the single definition.

**Six properties accepted none of the CSS-wide keywords** — not even `inherit`. `colorScheme: 'inherit'` was a type error, as were `marginTrim: 'unset'`, `paintOrder`, `textJustify`, `WebkitBackgroundClip` and `WebkitBoxOrient`. They were written `null | 'a' | 'b'` rather than `all | 'a' | 'b'`.

## Linked PR/Issues

Stacked on #1878.

## Additional Context

The cause of the second one is worth recording, because it explains why nobody noticed. Those six are the only properties in the file with no named type alias. Everything else declares `type foo = …` and writes `foo?: all | foo`; these inline their union at the property instead — and reaching for `null`, which is what `all` happens to start with, rather than composing `all`, is the slip that inlining hid. Five date to the initial commit; `WebkitBackgroundClip` was added later by "Chore: Allow additional -webkit- properties (#28)" and copied the shape.

Both fixes are pure widenings: no value that used to typecheck stops doing so. Confirmed after the change that `color: 'revert-layer'`, `colorScheme: 'inherit'`, `marginTrim: 'unset'` and `paintOrder: 'revert'` all compile.

The `NO_GLOBAL_KEYWORDS` allowlist that #1878 introduced for these six is deleted rather than shortened, so the pattern cannot come back.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/stylex/__tests__/css-types-test.js
+++ b/packages/@stylexjs/stylex/__tests__/css-types-test.js
@@ -42,23 +42,6 @@ const GLOBAL_KEYWORDS = new Set([
   'revert-layer',
 ]);
 
-/**
- * Properties that do not accept the CSS-wide keywords at all.
- *
- * Every property is meant to compose the `all` alias. These six are instead
- * written `null | 'a' | 'b'`, which drops `inherit`, `initial` and `unset`.
- * They are also the only properties in the file with no named type alias --
- * they inline their union at the property, which is what hid the slip.
- */
-const NO_GLOBAL_KEYWORDS = new Set([
-  'colorScheme',
-  'marginTrim',
-  'paintOrder',
-  'textJustify',
-  'WebkitBackgroundClip',
-  'WebkitBoxOrient',
-]);
-
 /* ------------------------------------------------------------------ *
  * Reading the type files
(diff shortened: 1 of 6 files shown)


henryqdineen:
Superseded by #1884. Reopened from branches in this repository rather than my fork, so the five PRs can be a proper [stack](https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/creating-stacked-pull-requests) — cross-fork stacks aren't supported, and each PR here was showing the earlier ones' commits in its diff. Same commits, unchanged.

PR #1880, opened 2026-09-11
[types] cursor should accept arbitrary strings

### LINKED ISSUE #1463: `url` value for `cursor` property

henryqdineen:

> **Stacked on #1879.** GitHub will not let a pull request from a fork target a branch on that fork, so this targets `main` and its diff therefore includes #1877 through #1879. Review the last commit only (`cursor should accept arbitrary strings`).

`cursor` takes a comma-separated list of `url()` values before its keyword:

    cursor: [ [ <url> [ <x> <y> ]? , ]* [ auto | default | none | … ] ]

A union of literals can never describe that, so `cursor: 'url(cur.png) 4 12, auto'` was a type error — the bug reported in #1463.

**This obsoletes #1466.** That PR fixes the same issue by adding a bare `string` to `CSSCursor`. Here the TypeScript definitions get `(string & {})` instead, so all 43 cursor keywords keep autocompleting while any string is accepted — which is what I wanted in #1466 and could not have while the definitions were generated from the Flow types. #1466 can be closed in favour of this.

Opening the union also made `cursor`'s `INCOMPLETE_UNIONS` entry unreachable, since that check skips properties accepting arbitrary strings. Rather than leave a dead entry, the seven keywords it recorded are added: `hand` (the pre-standard IE spelling) and the `-moz-` and zoom variants, where StyleX had `-webkit-grab` and `-webkit-grabbing` but not their siblings.

## Linked PR/Issues

Fixes #1463. Obsoletes #1466. Stacked on #1879.

## Additional Context

Found mechanically rather than by inspection: test 5 from #1878 asserts that a property whose grammar admits non-keyword values accepts arbitrary strings, and reported this along with its grammar and the exact two-file edit.

Confirmed after the change: `cursor: 'url(cur.png) 4 12, auto'`, `cursor: '-moz-grab'` and `cursor: 'pointer'` all typecheck, and the TypeScript language service still offers 43 completions for `cursor`.

## Pre-flight checklist

### DIFF
--- a/packages/@stylexjs/stylex/__tests__/css-types-test.js
+++ b/packages/@stylexjs/stylex/__tests__/css-types-test.js
@@ -540,17 +540,6 @@ const INCOMPLETE_UNIONS = {
   backgroundClip: ['border-area'],
   breakAfter: ['all', 'always'],
   breakBefore: ['all', 'always'],
-  // StyleX has `-webkit-grab` and `-webkit-grabbing` but not the `-moz-`
-  // equivalents or the zoom variants. `hand` is the pre-standard IE spelling.
-  cursor: [
-    '-moz-grab',
-    '-moz-grabbing',
-    '-moz-zoom-in',
-    '-moz-zoom-out',
-    '-webkit-zoom-in',
-    '-webkit-zoom-out',
-    'hand',
-  ],
   display: [
     '-moz-box',
     '-moz-inline-box',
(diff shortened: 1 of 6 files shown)


henryqdineen:
Superseded by #1885. Reopened from branches in this repository rather than my fork, so the five PRs can be a proper [stack](https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/creating-stacked-pull-requests) — cross-fork stacks aren't supported, and each PR here was showing the earlier ones' commits in its diff. Same commits, unchanged.

PR #1881, opened 2026-09-11
[types] Five worked examples of shrinking the CSS type allowlists

henryqdineen:

> **Stacked on #1880.** GitHub will not let a pull request from a fork target a branch on that fork, so this targets `main` and its diff therefore includes #1877 through #1880. Review the last five commits only.

Optional — #1877 through #1880 stand on their own. This is one worked example per allowlist from #1878, each its own commit, demonstrating the loop the tests exist for: **cross an entry off the allowlist, the test says what to change, change it, the test goes green.**

| Commit | Allowlist | Change |
| --- | --- | --- |
| Type `overflowInline` | `NOT_TYPED` | adds a property — Baseline high since 2021, and the counterpart of `overflowBlock`, which was already typed |
| Drop `overflowBlockX` | `NO_GRAMMAR` | deletes a property that is in no spec; the pair is `overflow-block`/`overflow-inline` |
| Drop `outlineColor: invert` | `KEPT_KEYWORDS` | removes a CSS 2.1 keyword only IE implemented |
| `mixBlendMode: plus-darker`/`plus-lighter` | `INCOMPLETE_UNIONS` | adds two keywords in the spec since Compositing Level 2 |
| Eleven `display` keywords | `INCOMPLETE_UNIONS` | adds `flow`, `table-caption`, and the nine legacy flexbox/grid spellings |

Each commit message quotes the failure that prompted it. (#1880 is a sixth example of the same loop, split out because it fixes a reported issue.)

## Linked PR/Issues

Stacked on #1880.

## Additional Context

**`mixBlendMode` shows the trap to watch for.** It aliased the shared `blendMode`, which `backgroundBlendMode` also uses — and the grammar says those two keywords belong to `mix-blend-mode` alone (`mix-blend-mode: <blend-mode> | plus-darker | plus-lighter` versus `background-blend-mode: <blend-mode>#`). Widening the shared alias would have wrongly allowed `backgroundBlendMode: 'plus-darker'`. Roughly half the remaining `INCOMPLETE_UNIONS` entries alias a shared type, so the grammar is what tells you whether a keyword belongs to the shared type or just the one property.

**Typing `overflowInline` made a different test fail**, because the alias it reuses is missing the same five keywords its siblings are. That commit therefore adds the property and its `INCOMPLETE_UNIONS` entry together — the tests catching a second-order gap introduced by closing a first-order one.

Remaining after these: `NO_GRAMMAR` 16, `KEPT_KEYWORDS` 19, `INCOMPLETE_UNIONS` 32, `CLOSED_DESPITE_GRAMMAR` 13, `NOT_TYPED` 24. `positionArea` alone accounts for 38 of the missing keywords.

## Pre-flight checklist


henryqdineen:
Superseded by #1886. Reopened from branches in this repository rather than my fork, so the five PRs can be a proper [stack](https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/creating-stacked-pull-requests) — cross-fork stacks aren't supported, and each PR here was showing the earlier ones' commits in its diff. Same commits, unchanged.

PR #1892, opened 2026-09-16
fix(unplugin): preserve SSR link attributes during CSS handoff

agcty:

When the development runtime is imported before an SSR document hydrates, its CSS fetch can finish first. The runtime then sets `link.disabled = true` on the server-rendered `/virtual:stylex.css` link, adding a `disabled` attribute that the hydrating application did not render. Import timing contributes to this race; deferring the import until after hydration is a possible application workaround.

Disable the loaded `CSSStyleSheet` instead, preserving the link's attributes while retaining the existing handoff to `style#__stylex_virtual__`. If the link is still loading (`link.sheet === null`), disable its sheet when the load event fires. Apply the same fix to both generated runtime scripts. No API change.

### Minimal reproduction

1. Render `<link rel="stylesheet" href="/virtual:stylex.css">` in an SSR document head.
2. Start `import('virtual:stylex:runtime')` in the client entry before hydrating the document with the same link markup.
3. Delay hydration until the runtime's CSS fetch completes.
4. Before this change, the link has acquired `disabled=""`; after this change its markup is unchanged and its original stylesheet is disabled through CSSOM.

This is a timing-sensitive integration case, not a claim that every SSR import produces a warning.

## Linked PR/Issues

No existing matching issue found in the upstream hydration search.

## Additional Context

- `yarn workspace @stylexjs/unplugin test --runInBand`: 21 tests pass.
- Four new regression tests cover both runtime paths, an already-loaded sheet, delayed loading, unrelated links, unchanged attributes, and subsequent HMR CSS replacement. All four fail against the unchanged upstream implementation.
- Focused ESLint, Prettier and `git diff --check` pass.
- Separate Chromium verification executed both generated scripts with loaded and deliberately delayed stylesheets. In all four cases the original rule stopped applying, replacement CSS applied, `link.outerHTML` stayed identical, and `link.disabled` remained false. The Jest tests mock sheet objects; they do not by themselves establish browser CSS application or full React hydration.
- Full `yarn test` passes (build, Flow, workspace tests, TypeScript tests, repository Prettier and ESLint). The first sandboxed attempt hit a Flow `sysctl` permission error; the unsandboxed rerun completed successfully.

## Pre-flight checklist


nmn (approved):
Thanks for the nice fix!

PR #1894, opened 2026-09-17
docs: document compiling StyleX that ships inside dependencies

refirst11:
Follow-up to #1891, where a Next.js app consuming a package with uncompiled StyleX hit `Unexpected 'stylex.keyframes' call at runtime`, then a second failure that prints nothing.

Two behaviors are not currently written down:

- **Neither mechanism cascades.** `transpilePackages` has to name every package in the chain, and so does the PostCSS `include` option. Listing only the direct dependency leaves a transitive one uncompiled.
- **An explicit `include` disables auto-discovery** (`resolveIncludeWithMetadata` returns early) and the default excludes (`resolveExclude`). Since the Next.js guide sets `include` to `app/`, `components/`, etc., a dependency is never scanned. The JS compiles and the class names land in the HTML, but no rules are generated for them and nothing is printed.

Auto-discovery does not help in the transitive case either: `discoverStylexPackageDirectories` reads the root manifest's dependency names and does not recurse.

Changes:

- `learn/installation/nextjs.mdx` — new "Styles from dependencies" section covering both steps, the `node_modules` glob in an explicit `include`, resolving a non-hoisted path, and `STYLEX_POSTCSS_DEBUG=1`.
- `api/configuration/postcss-plugin.mdx` — spell out the two consequences under `include`.

Docs only; no code changes.


nmn:
Thanks for the PR. The documentation needs to be reworded to make it easier to understand.

refirst11:
Thanks for the review. I've reworded it into shorter, step-by-step instructions.

